PR #129, opened 2023-11-09
support typescript

### LINKED ISSUE #79: TypeScript
We could support TypeScript for

- inline expressions
- `ts`-tagged fenced code blocks
- imported TypeScript modules (`.ts`)

We want to use [esbuild’s JavaScript API](https://esbuild.github.io/api/#js-details) for this.

We should consider type checking of TypeScript a separate (bonus, dependent) feature. To start, let’s just worry about interoperability with TypeScript, and developers can use VS Code or their own tooling to perform type checking.

Fil:
TODO:
- [x] code blocks
- [x] inline expressions
- [x] imports
- [x] tests
- [ ] documentation
- [ ] do we need the `ts` tag? if all the `js` cells are "cleaned up"…

closes #79


mbostock (changes requested):
This is now up-to-date with main, and I’d say 98% ready in some sense. But I think there is a major problem with this approach, and also a small (but potentially challenging) issue with the implementation.

The major problem is that this “supports TypeScript” in a sense, but all it really does is strip TypeScript annotations. Which means you don’t get any of the type safety of TypeScript. Which raises the question of why you’re using TypeScript in the first place? Is it just so the code runs without modification? Or do you want to have the added safety of TypeScript to validate the types? As an example, you can have a TypeScript code block like this:

```ts
const file: whatever = FileAttachment("file.txt");
```

and it happily runs since the `: whatever` is stripped. My sense is that this isn’t what people want when they ask for TypeScript in Framework; I think they want type checking. And that means we can’t simply use esbuild to strip the type annotations — we’ll need to use tsc (or whatever) to check the types.

The minor problem is preserving the semantics of expressions. The current approach looks for a trailing semicolon, and doesn’t consider comments. I think to do this correctly, we probably need to parse the TypeScript and see if it’s an expression. Which means we need a TypeScript parser (not just Acorn). Which is likely more expensive than running esbuild.

I’m not really sure how to proceed here. But I think we should investigate type checking, probably using tsc, if we want to support TypeScript in Framework.

Fil:
This is a first step that allows you to write ts (or copy/paste ts) and run the code. But the hope is that we'll be able to get type checking with a VSCode extension #882? 🤔

codingedgar:
As a user I don't want tsc to run with the framework, or at least not in dev, I can do that separately before deploy, once.

Storybook also ran tsc for the users and it made it very slow, they added a config flag to skip checks, which is a blessing, on by default now iirc.

I think the value of TypeScript will be provided by the editor itself running the checks and autogenerated definitions to glue parts of the framework together (that can come after)

Just being able to load ts files is plenty of gain.

codingedgar:
Now that I read it again, I want to clarify.

tsc should be able to run and detect errors, just not constantly in dev, but as a stand alone command; would expect `npx tsc` to work.

Not sure if thats what you mean it's not working?

mbostock:
Yes we should consider landing this without enforcing type checking and just stripping type annotations. We’ll need to merge main but that should be doable.

Issue #60, opened 2023-10-24
Named themes


mbostock:
As an alternative to including a custom stylesheet such as `dashboard.css`, we should allow named themes to be applied via YAML frontmatter, and provide built-in stylesheets for these themes. The current theme could be moved to `default.css`.

PR #414, opened 2023-12-20
bundle css; custom styles; named themes

mbostock:
This adopts esbuild to bundle our CSS. This should eventually help with #60, but the idea for now is merely to separate the CSS into separate files so it’s a bit easier to manage.

### DIFF
--- a/src/rollup.ts
+++ b/src/rollup.ts
@@ -15,17 +15,19 @@ import {getObservableUiHost} from "./observableApiClient.js";
 import {Sourcemap} from "./sourcemap.js";
 import {relativeUrl} from "./url.js";
 
+const STYLE_MODULES = {
+  "observablehq:default.css": getClientPath("./src/style/default.css"),
+  "observablehq:theme-auto.css": getClientPath("./src/style/theme-auto.css"),
+  "observablehq:theme-dark.css": getClientPath("./src/style/theme-dark.css"),
+  "observablehq:theme-light.css": getClientPath("./src/style/theme-light.css")
+};
+
 export async function bundleStyles(clientPath: string): Promise<string> {
   const result = await build({
     bundle: true,
     entryPoints: [clientPath],
     write: false,
-    alias: {
-      "observablehq:default.css": getClientPath("./src/style/default.css"),
-      "observablehq:theme-auto.css": getClientPath("./src/style/theme-auto.css"),
-      "observablehq:theme-dark.css": getClientPath("./src/style/theme-dark.css"),
-      "observablehq:theme-light.css": getClientPath("./src/style/theme-light.css")
-    }
+    alias: STYLE_MODULES
   });
   return result.outputFiles[0].text;
 }


dleeftink:
Am I right to understand that with the current bundler, a new .css file will be bundled per .md file that has a custom `style` override defined in the YAML frontmatter? 

In many cases, I need some pages to only load an additional snippet on top of the site’s global `customs.css` (which is loaded on first visit), instead of having an additional .css file bundled/served just for the snippet.

For instance, using `style: "/article.css"` as a per-page override, I would expect the page to re-use the (cached) `custom.css` as the base style, and only load the additional `article.css` when visiting a page that has the style override set.

In `article.css` I can then simply `@import` various styled components to keep the style overrides modular.

```
@import url("./article/header.css")
@import url("./article/section.css")
@import url("./article/typography.css")
```

The `article.css` will then only bundle the imports instead of bundling the full theme + page overrides. This file is then appended to the page head after the global `custom.css`.

Another way of achieving this is providing a `class` field in the YAML metadata, which allows users to bundle all their styles in a single `custom.css` and apply per-page overrides in the global stylesheet itself (a less modular approach I think).

dleeftink:
Am I right to understand that with the current bundler, a new .css file will be bundled per .md file that have a custom `style` override defined in the YAML frontmatter? 

In many cases, I need some pages to only load an additional snippet on top of the site’s global `customs.css` (which is loaded on first visit), instead of having an additional .css file bundled/served just for the snippet.

For instance, using `style: "/article.css"` as a per-page override, I would expect the page to re-use the (cached) `custom.css` as the base style, and only load the additional `article.css` when visiting a page that has the style override set.

In `article.css` I can then simply `@import` various styled components to keep the style overrides modular.

```
@import url("./article/header.css")
@import url("./article/section.css")
@import url("./article/typography.css")
```

The `article.css` will then only bundle the imports instead of bundling the full theme + page overrides. This file is then appended to the page head after the global `custom.css`.

Another way of achieving this is providing a `class` field in the YAML metadata, which allows users to bundle all their styles in a single `custom.css` and apply per-page overrides in the global stylesheet itself (a less modular approach I think).

dleeftink:
Am I right to understand that with the current bundler, a new .css file will be bundled per .md file that have a custom `style` override defined in the YAML frontmatter? 

In many cases, I need some pages to only load an additional snippet on top of the site’s global `custom.css` (which is loaded on first visit), instead of having an additional .css file bundled/served just for the snippet.

For instance, using `style: "/article.css"` as a per-page override, I would expect the page to re-use the (cached) `custom.css` as the base style, and only load the additional `article.css` when visiting a page that has the style override set.

In `article.css` I can then simply `@import` various styled components to keep the style overrides modular.

```
@import url("./article/header.css")
@import url("./article/section.css")
@import url("./article/typography.css")
```

The `article.css` will then only bundle the imports instead of bundling the full theme + page overrides. This file is then appended to the page head after the global `custom.css`.

Another way of achieving this is providing a `class` field in the YAML metadata, which allows users to bundle all their styles in a single `custom.css` and apply per-page overrides in the global stylesheet itself (a less modular approach I think).

dleeftink:
Am I right to understand that with the current bundler, a new .css file will be bundled per .md file that have a custom `style` override defined in the YAML frontmatter? 

In many cases, I need some pages to only load an additional snippet on top of the site’s global `customs.css` (which is loaded on first visit), instead of having an additional .css file bundled/served just for the snippet.

For instance, using `style: "/article.css"` as a per-page override, I would expect the page to re-use the (cached) `custom.css` as the base style, and only load the additional `article.css` when visiting a page that has the style override set.

In `article.css` I can then simply `@import` various styled components to keep the style overrides modular.

```css
@import url("./article/header.css")
@import url("./article/section.css")
@import url("./article/typography.css")
```

The `article.css` will then only bundle the imports instead of bundling the full theme + page overrides. This file is then appended to the page head after the global `custom.css`.

Another way of achieving this is providing a `class` field in the YAML metadata, which allows users to bundle all their styles in a single `custom.css` and apply per-page overrides in the global stylesheet itself (a less modular approach I think).

dleeftink:
Am I right to understand that with the current bundler, a new .css file will be bundled per .md file that have a custom `style` override defined in the YAML frontmatter? 

In many cases, I need some pages to only load an additional snippet on top of the site’s global `customs.css` (which is loaded on first visit), instead of having an additional .css file bundled/served just for the snippet.

For instance, using `style: "/article.css"` as a per-page override, I would expect the page to re-use the (cached) `custom.css` as the base style, and only load the additional `article.css` when visiting a page that has the style override set.

In `article.css` I can then simply `@import` various styled components to keep the style overrides modular.

```css
@import url("./article/header.css");
@import url("./article/section.css");
@import url("./article/typography.css");
```

The `article.css` will then only bundle the imports instead of bundling the full theme + page overrides. This file is then appended to the page head after the global `custom.css`.

Another way of achieving this is providing a `class` field in the YAML metadata, which allows users to bundle all their styles in a single `custom.css` and apply per-page overrides in the global stylesheet itself (a less modular approach I think).

Issue #595, opened 2024-01-23
Tailwind integration


mbostock:
What if we included Tailwind CSS by default, and generated an optimized bundle during build? That seems useful.

mbostock:
Bumping this. Would be really handy.

plasmak11:
Would love this. At the minimum, even getting the grid / flexbox system. (like bootstrap grid) 

The current grid system works well for desktops, but for mobile, hard to get more than linear layout.

tgy:
This is really missing. Tailwind is so handy. It would be great to be able to completely change the layout, and integrate Observable Framework as a first-class citizen within a tailwind-based application.

mbostock:
You can paste this into your Markdown and use Tailwind’s “Play CDN” approach:

```html
<script src="https://cdn.tailwindcss.com"></script>
```

But, agreed, a proper implementation requires Tailwind running at preview/build time to generate the appropriate stylesheet on the fly. And also we have to deal with Tailwind breaking all of Framework’s default styles as described in #596.

fleea:
Bump. Would love to see it happen.

vladnicula:
👍 +1

tgy:
Don't 'bump' the maintainers like that. They're not at your disposal to implement features **you** consider a priority.

vladnicula:
> Don't 'bump' the maintainers like that. They're not at your disposal to implement features **you** consider a priority.

Seem like my comment trigger you. It was not my intention. I don't expect others to be at my disposal. 

Happy to help by reading some code regarding the way styles are handled in observablehq in that case and share any findings, provided that anyone shares some pointers to what to start reading.

fleea:
I also would like to apologize for my comment. It wasn't my intention to push for the implementation.

daniel-payne:
as a workaround while i wait for full integration, i added this to **observablehq.config.js**

```
head: '<script src="https://cdn.tailwindcss.com"></script>',
```

daniel-payne:
I now use this link to get my tailwind pkg, as unpkg stopped working

<script src="https://cdn.jsdelivr.net/npm/@tailwindcss/browser@4"></script>

PR #596, opened 2024-01-23
tailwind

mbostock:
Some things to figure out:

1. Tailwind has very opinionated [“preflight” styles](https://tailwindcss.com/docs/preflight) which are essentially incompatible with Markdown. (For example, paragraphs have no margins by default, and so smush together.) We’ll need to either disable these preflight styles completely, or add our own base styles for margins on paragraphs and headings, or change our Markdown generator to apply the appropriate Tailwind styles explicitly.

2. It seems like for preview we can use Tailwind’s CDN script. But it would be better if this were self-hosted, given that we’ve already installed Tailwind locally; we want the preview server to work offline. Maybe we could use the Tailwind CLI instead, and watch files, but then we’ll also need to update the stylesheet as the content changes. Overall, I think it’ll be a fair amount of work to get live preview working.

3. For build, presumably we can use Tailwind’s CLI, so that doesn’t seem too bad. I was imagining that Tailwind would be an `observablehq:tailwind.css` module.

4. Users might need some way to configure Tailwind. Ideally, the Tailwind config is nested within the Observable config. This is all predicated on us being able to invoke Tailwind programmatically, which it doesn’t seem set up to do, but presumably we can do it using PostCSS and loading Tailwind as a plugin. 😓 

Ref. #595.

### DIFF
--- a/package.json
+++ b/package.json
@@ -69,6 +69,7 @@
     "rollup": "^4.6.0",
     "rollup-plugin-esbuild": "^6.1.0",
     "send": "^0.18.0",
+    "tailwindcss": "^3.4.1",
     "tar-stream": "^3.1.6",
     "tsx": "~4.2.1",
     "ws": "^8.14.2"
--- a/src/render.ts
+++ b/src/render.ts
@@ -81,6 +81,17 @@ if (location.pathname.endsWith("/")) {
 </script>`)
       : ""
   }
+<script src="https://cdn.tailwindcss.com"></script>
+<script>
+
+tailwind.config = {
+  corePlugins: {
+    preflight: false
+  }
+};
+
+</script>
+
 <script type="module">${html.unsafe(`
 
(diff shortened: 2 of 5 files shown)


tgy:
I think the prose plugin is the right way to go! Maybe we could also dynamically add tailwind classes to the HTML elements generated from the Markdown code?

Fil:
A way to add tailwind in `rollup.ts`:
>  yarn add --dev [esbuild-plugin-tailwindcss](https://github.com/ttempaa/esbuild-plugin-tailwindcss):

```diff
+import {tailwindPlugin} from "esbuild-plugin-tailwindcss";

     ...(path ? {entryPoints: [path]} : {stdin: {contents: renderTheme(theme!), loader: "css"}}),
     write: false,
     minify,
+    plugins: [
+      tailwindPlugin({
+        // ...options
+      })
+    ],
```

the rest is similar to this PR (`tailwind.config.js`, etc)

This works both in preview and on build (but does not solve the reset either). HTH.

Issue #396, opened 2023-12-17
Complex inline expressions aren’t parsed correctly


mbostock:
For example, this works fine:

```md
${html`<div style="display: grid; grid-auto-rows: 1fr; grid-template-columns: repeat(3, minmax(0, 1fr)); grid-gap: 1rem;">${await Promise.all(muybridge.filenames.map((f) => muybridge.file(f).image({style: "width: 100%;"})))}</div>`}
```

but this fails:

```md
${html`<div style="display: grid; grid-auto-rows: 1fr; grid-template-columns: repeat(3, minmax(0, 1fr)); grid-gap: 1rem;">
  ${await Promise.all(muybridge.filenames.map((f) => muybridge.file(f).image({style: "width: 100%;"})))}
</div>`}
```

Really, I want to do this:

```md
<div style="display: grid; grid-auto-rows: 1fr; grid-template-columns: repeat(3, minmax(0, 1fr)); grid-gap: 1rem;">
  ${Promise.all(muybridge.filenames.map((f) => muybridge.file(f).image({style: "width: 100%;"})))}
</div>
```

But I can’t do that because the wrapper spans interfere with the grid. #11

Overall, I think our current parser strategy is not very robust, and we probably want to pull inline expression parsing out of markdown-it and do it before (or after?) borrowing techniques from Hypertext Literal. Related #32.

Issue #379, opened 2023-12-13
Duplicate cells during incremental update with HEAD + META


cinxmo:
When I add another `<div>` tag, it duplicates the section:

https://github.com/observablehq/cli/assets/37343722/22beb08e-902c-449f-8b2d-8870001d3754

mbostock:
Fixed in #843.

Issue #375, opened 2023-12-12
uncaught TypeError: HTML comments with template literals and string interpolation


cinxmo:
In a markdown file, add:
```
<!--This is a comment.
Comments are not displayed in the browser ${'hello'}
-->
```
In the browser console, this shows up:
<img width="481" alt="Screenshot 2023-12-12 at 4 57 04 PM" src="https://github.com/observablehq/cli/assets/37343722/78609085-d69c-4b7a-a7a3-4ddb15e4c248">

Issue #252, opened 2023-11-28
Parse failure with text before a style element?


mbostock:
This:

```md
<link rel="stylesheet" type="text/css" href="custom.css">
text
<style type="text/css">

body { color: red; }

</style>
```

Generates a surprising (and broken) result:

```html
<span><link rel="stylesheet" type="text/css" href="./_file/custom.css">
text
<style type="text/css">
</style></span><p>body { color: red; }</p>
<span>
</span>
```

It seems that the open `<style>` is considered part of the first block, and gets separated from the style contents. Another blank line problem!

mbostock:
Fixed in #843.

PR #597, opened 2024-01-23
More robust interpolation parsing

mbostock:
This rearchitects our Markdown parser to be more robust. The intent is to fix issues such as:

- #32
- #375
- #396
- #379 (maybe)
- #252 (maybe)
- the blank line quirk when using HTML in Markdown

Rather than implementing inline expression parsing as a markdown-it plugin (which depends on markdown-it’s internal tokenization), we implement a preprocessing step that converts inline expressions `${…}` to HTML and extracts the JavaScript source expressions. This HTML is then ignored by markdown-it since HTML is allowed within Markdown. We will continue to parse fenced code blocks using markdown-it, since these are already supported by Markdown, and since these are always generate nodes.

Furthermore, by employing the same [HTML parsing state machine](https://github.com/observablehq/htl/blob/main/README.md#how-it-works) as Hypertext Literal, we can be exact about where the expressions start and end, and in what context they need to be evaluated (for example, as an attribute or as an node).

For example, this inline expression:

```md
Hello, ${'world'}!
```

is compiled to:

```md
Hello, <!-- o:1 -->!
```

In addition, the `'red'` JavaScript expression is extracted so that it can be used to define a runtime variable, which is then displayed on the client, replacing the generated comment. (Note that we could generate `<span id="cell-1"></span>` instead of the comment here, and we may end up doing that, but I’ll need to change the client to support evaluating dynamic attributes anyway, so I’m opportunistically seeing if this helps remove the wrapper span #11.)

Similarly, this:

```md
<div class=${'red'}>color</div>
```

is compiled to:

```md
<div o:class="1">color</div>
```

where `o:` is a special prefix that denotes a dynamic attribute that will be computed on the client. The exact compiled HTML syntax is still to be determined — the current approach will require using a [TreeWalker](https://developer.mozilla.org/en-US/docs/Web/API/TreeWalker) to find comments and attributes, and then bind them with the runtime variable with the associated identifier.

The last thing to fix here is probably to diff the parsed HTML rather than diffing the parsed Markdown pieces. This would alleviate the requirement that each Markdown “piece” corresponds to exactly one HTML element, which introduces the blank line quirk. Of course, not all of these depend on each other, so I might try to decouple them and approach this more incrementally.

### DIFF
--- a/src/interpolate.ts
+++ b/src/interpolate.ts
@@ -1,0 +1,571 @@
+const CODE_TAB = 9,
+  CODE_LF = 10,
+  CODE_FF = 12,
+  CODE_CR = 13,
+  CODE_SPACE = 32,
+  CODE_UPPER_A = 65,
+  CODE_UPPER_Z = 90,
+  CODE_LOWER_A = 97,
+  CODE_LOWER_Z = 122,
+  CODE_LT = 60,
+  CODE_GT = 62,
+  CODE_SLASH = 47,
+  CODE_DASH = 45,
+  CODE_BANG = 33,
+  CODE_EQ = 61,
+  CODE_DQUOTE = 34,
+  CODE_SQUOTE = 39,
+  CODE_QUESTION = 63,
+  CODE_DOLLAR = 36,
+  CODE_LBRACE = 123,
+  CODE_RBRACE = 125,
+  CODE_BACKSLASH = 92,
+  // CODE_QUOTE = 34,
+  // CODE_SINGLE_QUOTE = 39,
+  CODE_BACKTICK = 96,
+  STATE_DATA = 1,
+  STATE_TAG_OPEN = 2,
+  STATE_END_TAG_OPEN = 3,
+  STATE_TAG_NAME = 4,
+  STATE_BOGUS_COMMENT = 5,
+  STATE_BEFORE_ATTRIBUTE_NAME = 6,
+  STATE_AFTER_ATTRIBUTE_NAME = 7,
+  STATE_ATTRIBUTE_NAME = 8,
+  STATE_BEFORE_ATTRIBUTE_VALUE = 9,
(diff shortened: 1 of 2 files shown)


mkotelnikov:
Hello, 

[@statewalker/tknz](https://github.com/statewalker/statewalker-tknz) - is a tokenizer (parser) for HTML / MD / syntaxes. 
- It is very small (30k non minimised, non compressed) with no dependencies
- Produces well formed ASTs from documents with exact start/end positions for each token
- Opening/closing HTML tags are balanced (by default)
- Allows embedding of inline codes
- Produces well-formed hierarchy of  document sections based on headers

It seems that it covers most issues mentioned here. 
You can check it here: https://observablehq.com/@kotelnikov/statewalker-tknz

mbostock:
We now diff HTML, and #1416 removes the wrapper span, so two of the pieces have fallen into place.

Yet one limitation I see with this approach is the assumption that the HTML tokenizer state machine can be applied to Markdown (as-is). I think it can in about 98% of cases because nearly all of Markdown is text context, but here’s at least one example where it fails:

```md
[link](https://example.com/${"path"})
```

In the above case, `${"path"}` would appear to the HTML tokenizer as the text context, but in fact it’s interpolating into the `href` attribute. There’s also the more pathological case where the same expression in interpolated into both the text _and_ an attribute simultaneously:

```md
<https://example.com/${"path"}>
```

And I have no idea how auto-linkify should work in this case…

```md
https://example.com/${"path"}
```

These cases aren’t currently handled in main, either: the `${"path"}` isn’t recognized as an inline expression. And it’s pretty easy to rewrite this as HTML if for some reason you wanted this dynamic behavior for some reason. But it suggests that the tokenizer would at least need to recognize Markdown’s link syntax. I probably need to read the CommonMark spec to see if there are other cases.

PR #670, opened 2024-02-03
css

### LINKED ISSUE #666: CSS code blocks
We could support live CSS code blocks?

````html
```css
body {
  background: red;
}
```
````

You could use interpolation expressions within just like html code blocks, too.

### PR DESCRIPTION
**Pull request by mbostock**
Fixes #666. Not sure there’s a lot of value here (especially relative to #597), but it’s easy to do. My silly test case:

````html

```css
body {
  color: hsl(${x}, 50%, 50%);
}
```

```js
const x = Generators.observe((notify) => {
  notify(0);
  const pointermove = (event) => notify(event.clientX);
  addEventListener("pointermove", pointermove);
  return () => removeEventListener("pointermove", pointermove);
});
```
````

### DIFF
--- a/src/client/stdlib/css.ts
+++ b/src/client/stdlib/css.ts
@@ -1,0 +1,11 @@
+export default function css({raw: strings}) {
+  let string = strings[0];
+  for (let i = 1, n = arguments.length; i < n; ++i) {
+    const value = arguments[i]; // eslint-disable-line prefer-rest-params
+    if (value != null) string += `${value}`;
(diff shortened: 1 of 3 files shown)


mbostock:
We could server-side render the CSS (like we do with PRQL in #1078). And if there are interpolations, we could additionally have JavaScript to update the CSS dynamically (perhaps more selectively than we currently do? like we could only allow interpolations into CSS values rather than arbitrary chunks of CSS).

PR #701, opened 2024-02-07
sidebar: "hidden"

Fil:
For #696. Note: as I was writing the documentation I thought that "closed" would be a better word. Rather than "hide" and "show" the sidebar, we'd open and close it. wdyt?


mbostock (approved):
We don’t need the explicit *auto* option (nor likewise *auto-hidden* if you want to combine *auto* and *hidden*); we can just have the default be true or false as before, and have *hidden* if you want an initially-hidden sidebar. Simpler! 👍

mbostock (changes requested):
Hmm, in further testing, this doesn’t seem to persist the sidebar state in the session: if you visit a page and toggle the sidebar open, and then reload, the sidebar disappears; likewise if you navigate between pages, the sidebar keeps disappearing. I think that’s because we clear the session state if the sidebar would be initially visible (ignoring the *hidden* setting). In other words, the client needs to check the data-hidden attribute when deciding whether to persist the sidebar state in session storage.

mbostock:
Converting to draft given the needed changes.

PR #702, opened 2024-02-07
memoize parseMarkdown

Fil:
This will make operations such as (re)indexation (#543) and lazy config (#695) much more usable in preview, and might save a few seconds during build (instead of parsing each file 2 or 3 times, we'd do it only once), for example to assist with last-modified (#699) and hashing (#758).

To make sure that we don't inflate the memory requirements I'm capping this arbitrarily to a store that has a maximum of 16^4 shortCode keys — as long as the percentage of conflicts is small the impact on speed is not going to be a problem. (Note: we take care of checking the full hash before deciding to use the cache, so as not to mix two files which would happen to share the same shortCode slot.)

### DIFF
--- a/src/markdown.ts
+++ b/src/markdown.ts
@@ -417,30 +417,41 @@ export interface ParseOptions {
   path: string;
 }
 
+const mdmap = new Map();
+
 export async function parseMarkdown(sourcePath: string, {root, path}: ParseOptions): Promise<ParseResult> {
   const source = await readFile(sourcePath, "utf-8");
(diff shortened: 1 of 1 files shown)


Fil:
needs to be revisited after #843

Fil:
mhhh… this might still be useful for #695

mbostock:
I’m guessing we don’t need this anymore after recent optimizations?

Fil:
yes!

PR #758, opened 2024-02-12
hash files

### LINKED ISSUE #260: Use hashed names to ensure cache-breaking on deploy
This should apply to:

- file attachments
- imported modules
- framework code (e.g., style.css, client.js)

Fil:
first take at cache-busting urls hashing contents

* computes hashes for all urls if the `hashFiles` option is set for build
* saves files under filename.{hash}.ext
* renames them where they are called

For example when building dist/javascript/files.html this injects:

```diff
define({
  id: "47076500",
  inputs: ["FileAttachment","display"],
  files: [{
-    "path":"../_file/javascript/volcano.json",
+    "path":"../_file/javascript/volcano.f5223ec980.json",
    "name":"./volcano.json",
    "mimeType":"application/json"
  }],
  body: async (FileAttachment,display) => {
…
```

TODO:
- [-] references other than _file/
  - [ ] imported modules (are already hashed, double check?)
  - [-] framework code
    - [x] style.css
    - [x] client.js
    - [ ] /_observablehq/runtime.js, /_observablehq/stdlib.js (called from client.js)
    - [ ] /_observablehq/stdlib/inputs.js  (e.g., called from the inputs/button.html)
- [x] put the hash in the file name (?? maybe, I personally find this makes debugging much more difficult though)
  - [x] note that the fileHash function stores the file hash and returns the file name, so it's not too hard to change this to `{pathwithoutextension}.{hash}.{ext}` (or maybe a substring of hash)
- [x] the `hash` option is mandatory
- [x] fix tests
- [ ] see how deploy may want to use this
- [ ] add a specific build test (once it stabilizes)

closes #260

### DIFF
--- a/bin/observable.ts
+++ b/bin/observable.ts
@@ -89,9 +89,7 @@ try {
       const {
         values: {config, root}
       } = helpArgs(command, {
-        options: {
-          ...CONFIG_OPTION
-        }
+        options: {...CONFIG_OPTION}
       });
       await import("../src/build.js").then(async (build) => build.build({config: await readConfig(config, root)}));
       break;
--- a/docs/theme/generate-themes.ts
+++ b/docs/theme/generate-themes.ts
@@ -214,7 +214,7 @@ There is nothing surprising in this. If they but knew it, almost all men in thei
 }
 
 async function generateFile(path: string, contents: string): Promise<void> {
-  console.log(`${faint("generate")} ${path}`);
+  console.log(`${faint("generating")} ${path}`);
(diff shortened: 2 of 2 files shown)


Fil:
almost superseded by #843

maybe we can salvage 1 line which defines the way the hash is incorporated in the filename

mbostock:
Yeah, I want to look into incorporating this approach into #843. It looks like it be readily doable now that I broke out `rewriteHtml` from `parseMarkdown` so we can do it after running data loaders.

PR #764, opened 2024-02-13
observable convert

mbostock:
A minimal implementation of converting an Observable notebook to Markdown. For example:

```
observable convert https://observablehq.com/@observablehq/plot-scatterplot/2
```

This outputs a **plot-scatterplot,2.md** containing:

````md
<div style="color: grey; font: 13px/25.5px var(--sans-serif); text-transform: uppercase;"><h1 style="display: none;">Plot: Scatterplot</h1><a href="/plot">Observable Plot</a> › <a href="/@observablehq/plot-gallery">Gallery</a></div>

# Scatterplot

Given a dataset with two quantitative dimensions, the [dot](https://observablehq.com/plot/marks/dot) mark creates a scatterplot.

```js echo
Plot.dot(penguins, {x: "culmen_length_mm", y: "culmen_depth_mm"}).plot()
```
````

You can also use shorthand:

```
observable convert @observablehq/plot-scatterplot/2
```

This is mostly designed under the [worse is better](https://dreamsongs.com/RiseOfWorseIsBetter.html) philosophy. I didn’t try to implement anything fancy, like parsing Observable JavaScript and rewriting it to vanilla JavaScript. The expectation is that it’ll probably break, but the code will be right there so it should be “easy” to rewrite the code manually. And in many cases, I don’t think we’ll be able to rewrite the code automatically. Maybe we could have some opt-in (or opt-out) smart rewrites in the future, but this feels like a useful thing to include to get things going!

I’d also like to allow it to use your credentials to convert private notebooks if you `observable login`.

### DIFF
--- a/src/convert.ts
+++ b/src/convert.ts
@@ -7,7 +7,7 @@ import type {ClackEffects} from "./clack.js";
 import {CliError} from "./error.js";
 import {prepareOutput} from "./files.js";
 import {getObservableUiOrigin} from "./observableApiClient.js";
-import {type TtyEffects, bold, faint, inverse, defaultEffects as ttyEffects, yellow, link, cyan, reset} from "./tty.js";
+import {type TtyEffects, bold, cyan, faint, inverse, link, reset, defaultEffects as ttyEffects} from "./tty.js";
 
 export interface ConvertEffects extends TtyEffects {
   clack: ClackEffects;

### CODE REVIEW

**Comment by Fil** [comment]
> the comma in the name feels rough

this should only happen when the original notebook's slug contains a `/`:

```
> yarn observable convert @d3/bar-chart/2
◇  Downloaded bar-chart,2.md in 296ms
◇  Downloaded alphabet.csv in 596ms
```

because we can't use a slash in a file name (the choice of a comma in that case is a little arbitrary, but it lowers the risk of conflicting with another notebook (e.g. `bar-chart-2` if we used a dash).

PR #765, opened 2024-02-13
two pass render

Fil:
this builds on top of #702 (**memoization of the parsing function**) and the first few commits of #758 (the **two-pass rendering** of the **hash files** PR).

the first part (memoization) helps speed up (re)indexation (#543), lazy config (#695) reload; the second part (two-pass rendering) is necessary to make last-modified (#699) work reliably and to do content-hashes (#758).


Fil:
this will probably need to be revamped after #843

Fil:
for instance [5fbcad8](https://github.com/observablehq/framework/pull/765/commits/5fbcad8d22c84be9304c97615a53e0c7343c1352) is fixed in #843

PR #787, opened 2024-02-14
minify css

Fil:
it's only a 10% gain when we compare the gzip'd payloads, and we'd need a sourcemap?

not urgent in any case


Fil:
(I've started to look at extracting sourcemaps in #805)

mbostock (approved):

PR #788, opened 2024-02-14
support images and fonts url() in css

### LINKED ISSUE #786: esbuild crashes when style.css contains @font-face definitions
following https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face I create a style.css file with:

````
@font-face {
  font-family: "Trickster";
  src:
    local("Trickster"),
    url("trickster-COLRv1.otf") format("opentype") tech(color-COLRv1),
    url("trickster-outline.otf") format("opentype"),
    url("trickster-outline.woff") format("woff");
}
````

the css bundling crashes with recommendations to "mark paths as external" (unhelpful, since this is commandeered by framework, not by the user), [reference](https://esbuild.github.io/api/#external):

```
style docs/style.css → ✘ [ERROR] Could not resolve "trickster-COLRv1.otf"

    docs/style.css:5:8:
      5 │     url("trickster-COLRv1.otf") format("opentype") tech(color-COLRv1),
        ╵         ~~~~~~~~~~~~~~~~~~~~~~

  You can mark the path "trickster-COLRv1.otf" as external to exclude it from
  the bundle, which will remove this error and leave the unresolved path in the
  bundle.

✘ [ERROR] Could not resolve "trickster-outline.otf"

    docs/style.css:6:8:
      6 │     url("trickster-outline.otf") format("opentype"),
        ╵         ~~~~~~~~~~~~~~~~~~~~~~~

  You can mark the path "trickster-outline.otf" as external to exclude it from
  the bundle, which will remove this error and leave the unresolved path in the
  bundle.

✘ [ERROR] Could not resolve "trickster-outline.woff"

    docs/style.css:7:8:
      7 │     url("trickster-outline.woff") format("woff");
        ╵         ~~~~~~~~~~~~~~~~~~~~~~~~

  You can mark the path "trickster-outline.woff" as external to exclude it from
  the bundle, which will remove this error and leave the unresolved path in the
  bundle.


Unexpected error: Build failed with 3 errors:
docs/style.css:5:8: ERROR: Could not resolve "trickster-COLRv1.otf"
docs/style.css:6:8: ERROR: Could not resolve "trickster-outline.otf"
docs/style.css:7:8: ERROR: Could not resolve "trickster-outline.woff"

```

### PR DESCRIPTION
**Pull request by Fil**
closes #786

here's an example application (which I wouldn't necessarily recommend!) with a background image from docs/assets/ and fonts saved in docs/fonts/:

```css
:root {
  --serif: Cardo, serif, "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol";
  background-image: url("assets/paper-texture.jpeg");
}

@font-face {
  font-family: "Cardo";
  src: url("fonts/cardo.woff") format("woff");
  font-display: swap;
}
```

inline-css-images

You can also use static assets to modify the toggle or caret with the docs/toggle.svg file :

```css
:root {
  --theme-toggle: url("toggle.svg");
}
```

~~The resources are inlined in text (for svg) base64 (for binary images and fonts), so use with some caution. Note that inlining fonts avoids [FOUC](https://en.wikipedia.org/wiki/Flash_of_unstyled_content), and the style.css is cached, so it's not necessarily bad. Just put these definitions at the bottom of stylesheet? The alternative is to use absolute URLs to an assets server.~~ I'm working on saving these as assets.


mbostock (changes requested):
We should change this so that importing in CSS produces the same result as linking to a static asset in HTML (i.e., it goes in `_file`).

Fil:
superseded by #1372 😓

PR #804, opened 2024-02-15
cache data

mbostock:
~~TODO Apply the same technique to the examples. Currently this only caches the top-level project data loaders.~~ Done.


Fil (approved):
~~do we need to build examples conditionally (based on the target being absent)? If so how and when would we upgrade the examples?~~

PR #805, opened 2024-02-15
extract sourcemaps

Fil:


mbostock:
We should only do this for preview, not for build. (We could offer a way to opt-in during build, but I wouldn’t prioritize it or do it if it’s difficult.)

PR #807, opened 2024-02-15
Filtering out invalid tiers instead of filtering in.

Sylvestre67:
Following up on https://github.com/observablehq/framework/pull/790#discussion_r1490195836.  

_"...instead of filtering in we should filter out the workspaces that we know don’t allow projects"._


Fil (approved):

mbostock:
Thanks!

PR #811, opened 2024-02-15
echarts

mbostock:
Not sure if we’ll recommend it, but we can still make it easy if people want it.


mbostock:
Yeah, that didn’t work for me either. We could pin the `npm:echarts` version for now, and then remove the pinning if they fix it in the future (if we anticipate that it’s going to break immediately)?

mbostock:
Okay, pinned at 5.4.3 (for now).

Issue #715, opened 2024-02-07
Nits on getting-started.md


mcglincy:
Nits on getting-started:

- The code snippet for `temperaturePlot(forecast)` is missing a trailing semi-colon, which we show elsewhere

- does the frontmatter theme have to appear at the top of the file? And if so should we mention that in the text? (the side by side screenshot does show it on top, above the h1)

- "Deploy messages can be set using yarn deploy --message." Should we prefer example commands use npm?

mbostock:
I think these are mostly fixed?

PR #813, opened 2024-02-15
warn about implicit display gotcha in the getting-started example

ikesau:
hey friends : )

I was going through the getting started document, but got stuck for a minute when my beautiful `3` refused to appear.

https://github.com/observablehq/framework/issues/715 clued me into what was going on, and I then discovered it was written about in the docs, but I think jumping straight into `getting-started` will be a common path for many new users, so it might be worth adding this extra warning.

In general, given that Framework will situationally require semicolons, prettier's all-or-none approach to them seems incompatible. 

Possibly worth mentioning somewhere, too, given its ubiquity, but I'll hold off on doing anything else here in case you've got some specific editor setup recommendations at this point 👍


mbostock:
Good idea. Do you think it would be clearer if we just did the explicit form of `display` here?

````md
```js
display(1 + 2);
```
````

That way it won’t require a semicolon, and folks can discover the implicit display shorthand later. But then we’ll need it for the `Plot.plot` call too, which is a little more verbose but perhaps okay.

ikesau:
Yeah that makes sense, start without any magic and allow for a connected aha moment once you introduce implicit displays.

image

image

net:
This is a bit of an issue if you're using Prettier, which can only be configured to always add or always remove semicolons, but not to ignore them. I also spent some time trying to figure out why my code blocks weren't displaying 😕

mbostock (approved):

PR #825, opened 2024-02-16
index international text, ignore html entities

### LINKED ISSUE #816: search: normalize indexed text
we probably want to index  with [text.normalize](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/normalize) and remove accents. We could also index cyrillic and other typical text ranges.

    .normalize("NFD").replaceAll(/[\u0300-\u036f]/g, "") // removes accents from latin+
    .replace(/[^\p{L}\p{N}]/gu, " "); // only keeps letters (in any alphabet) and digits

etc. On the server, we must normalize (lowercase, streamline accentuated letters into ascii…) AND ignore  characters that are not "words" (emoji, punctuation, cruft…); on the client we only need to normalize.

Fil:
This normalizes international text in the search index:
- decodes HTML entities (`&copy; &egrave;` → © è) 
- accentuated letters in latin alphabet are normalized to their non-accentuated variant (é → e ; ñ → n), using [String.normalize("NFD")](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/normalize) + `.replace(/[\u0300-\u036f]/g, "")`
- all alphabets are recognized as letters with the unicode test `\p{L}` [Unicode_character_class_escape](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Regular_expressions/Unicode_character_class_escape)
- similarly digits are recognized with `\p{N}`

On the client we only need to do the NFD normalization step.

closes #816


src/search.ts
```diff
@@ -55,11 +56,17 @@ export async function searchIndex(config: Config, effects = defaultEffects): Pro
     // implicitly a leading slash here.
     const id = file.slice(0, basename(file) === "index.md" ? -"index.md".length : -3);
 
-    const text = html
-      .replaceAll(/[\n\r]/g, " ")
-      .replaceAll(/<style\b.*<\/style\b[^>]*>/gi, " ")
-      .replaceAll(/<[^>]+>/g, " ")
-      .replaceAll(/\W+/g, " ");
+    // eslint-disable-next-line import/no-named-as-default-member
```
Fil, line 59: Didn't understand why I could not `import {decode} from "he"` and had to do this instead.

mbostock: Yes, I haven’t been able to figure that out yet either. 😞

mbostock (approved):
💯

PR #826, opened 2024-02-16
bug fixes for search

### LINKED ISSUE #815: the home page should be indexed by default
if you specify pages, chances are you don't add index.md to them, but it should probably be indexed (unless it has search: false in its front matter).

Fil:
- add the home page to the default list of indexable pages
  - even though it's not in pages, it's not ‘unlisted’
  - one can still opt out with index: false, as we do for this documentation

closes #815

- support empty titles (replaced with a emdash), and titles such as numbers

you can test this by adding this to a listed and indexed page:

```
---
title: 3.141592
---

# Happi as π


```

then if you searched for "happi" you would see the client error in the console (and no result was displayed) as it was trying to apply the escaping code to a Number.


mbostock (approved):
💯

PR #835, opened 2024-02-16
windows tip

### LINKED ISSUE #822: Put doesn't work on Windows on the Framework getting started page.

Fil:
closes #822


Fil:
Since a lot of people don't read the docs, I would also like the create script to do an env check and exit if node is < 20.6 or the user is running windows.

mbostock (changes requested):
Can you put it as a separate tip with a yellow caution (instead of combining)? Thanks.

mbostock:
Can we add an os requirement to package.json, like we do with engines? Separate thread though.

mbostock (changes requested):

mbostock (approved):

PR #836, opened 2024-02-16
support firefox esr

Fil:
Allow Firefox ESR users to visit the documentation website.

The current version of Firefox ESR (v.115) is supported [until 1 October 2024](https://whattrainisitnow.com/calendar/)—at which point we'll be allowed to remove this patch since the next ESR is going to be based on v.128. I'll be setting myself a reminder.


mbostock (changes requested):
I would rather fix this by dropping the `:has` in the `tocSelector`, and doing it in JavaScript. I also don’t think we should squelch DOMExceptions — we want to notice when something is broken and fix it.

mythmon (approved):
I confirmed that this fixes the issue on lower Firefox version numbers.

Fil:
okay it wasn't hard; please take another look

mbostock (approved):

Fil:
The time has come to revise this… but I believe it's OK to do nothing after https://github.com/observablehq/framework/pull/1283

PR #840, opened 2024-02-16
keywords

mbostock:


docs/inputs/range.md
```diff
@@ -1,3 +1,7 @@
+---
+keywords: sliders
```
eagereyes, line 2: Maybe slider instead?

docs/javascript/files.md
```diff
@@ -1,3 +1,7 @@
+---
+keywords: file, fileattachment, attachment
```
tophtucker, line 2: just earlier today i was trying to get to this page typing blindly fast and landed on file input!! thank you 🙏

tophtucker (approved):
feels much better

docs/inputs/range.md
mbostock: It does prefix matching so this will match both.

PR #843, opened 2024-02-17
download npm imports; parser & resolver improvements

### LINKED ISSUE #16: During live preview, SyntaxError: Cannot use 'import.meta' outside a module
I’m not sure how we would fix this, but our current approach of eval’ing JavaScript source during live preview makes it incompatible with import.meta; however, you _can_ use import.meta when the page loads (and statically when building the site). So, it’d be nice if we could make import.meta work consistently.

### LINKED ISSUE #20: During build, download imported ES modules from jsDelivr

### LINKED ISSUE #252: Parse failure with text before a style element?

### LINKED ISSUE #260: Use hashed names to ensure cache-breaking on deploy

### LINKED ISSUE #379: Duplicate cells during incremental update with HEAD + META

### LINKED ISSUE #415: Linked stylesheets should auto-refresh on edit
If you have 

```html
<link rel="stylesheet" type="text/css" href="test.css">
```

then if you edit `test.css`, we want the page to update. This means:

1. We need to detect `test.css` as a file and watch it (which should already be happening in `normalizePieceHtml`, but doesn’t appear to be working as intended for some reason), and
2. We need to rewrite the `href` attribute to include the content hash, so that when the linked file changes, the browser will pick up the new file as intended.

### LINKED ISSUE #584: Render syntax errors statically
We currently render them in JavaScript, but we could just as easily render them statically.

### LINKED ISSUE #688: Skip unused client bundles during build

### LINKED ISSUE #763: Loading an Arrow or Parquet file should preload apache-arrow/parquet-wasm
You can workaround this manually with:

```js
import "npm:apache-arrow";
import "npm:parquet-wasm/esm/arrow1.js";
```

We’ll only be smart enough to do this in simple cases of `FileAttachment(name).parquet()` though.

### LINKED ISSUE #908: Linked assets should auto-refresh on edit
A generalization of #415. For example if you have an `` tag, we detect this a linked file; we could generate a URL with a hash, and push down the changes over the socket if the file changes.

mbostock:
- [x] Resolve static `npm:` imports to a local cache (instead of hot-linking from jsDelivr)
- [x] Resolve dynamic `npm:` imports to a local cache (e.g., Mermaid plugins)
- [x] Resolve `npm:` specifier versions against the local npm cache (for consistency & performance)
- [x] Populate the local npm cache during preview
- [x] Proxy `/_npm/` to jsDelivr during preview
- [x] Add a content hash to file attachments during preview (for live updating)
- [x] Add a content hash to file attachments during build (for cache-breaking)
- [x] Add a content hash to stylesheets during preview (for live updating)
- [x] Add a content hash to stylesheets during build (for cache-breaking)
- [x] Add a content hash to imported modules during preview (for live updating)
- [x] Add a content hash to imported modules during build (for cache-breaking)
- [x] Add a content hash to the search index during build (for cache-breaking)
- [x] Detect implicit `npm:` imports based on `FileAttachment` method or file extension
- [x] Detect implicit `npm:` imports based on unbound references (_e.g._, `d3` or `Plot`)
- [x] Detect implicit `npm:` stylesheets (_e.g._, for Leaflet and Observable Inputs)
- [x] Fix file references in static HTML
- [x] Fix file watching during preview
- [x] Fix incremental update during preview
- [x] Fix file watching of data loaders
- [x] Fix resolution of dynamic `npm:` imports in `transpileModule`
- [x] Fix issues downloading Mermaid’s plugins (jsDelivr timeouts)
- [x] Download (and rewrite) imports from npm during build
- [x] Download stylesheets from npm during build (_e.g._, Leaflet)
- [x] Download global stylesheets (KaTeX, Leaflet, Mapbox) during build
- [x] Download KaTeX fonts during build
- [x] Download extra files from npm during build (_e.g._, wasm bundles)
- [x] Fix build content hash for files generated by data loaders 
- [x] Fix incremental update when code output is moved

This also includes some fairly major re-architecting of parsing, transpiling, and rendering. Notes to come.

- Fixes #16.
- Fixes #20.
- Fixes #260.
- Fixes #252.
- Fixes #379.
- Fixes #415.
- Fixes #584.
- Fixes #688.
- Supersedes #758.
- Fixes #763.
- Incorporates #798.
- Fixes #908.


Fil:
When I open 127.0.0.1:3000/loaders the server crashes with

```
GET /_npm/_npm/d3@7.8.5/+esm.js
npm:_npm/d3@7.8.5/+esm.js → GET /_npm/_npm/isoformat@0.2.1/+esm.js
npm:_npm/isoformat@0.2.1/+esm.js → GET /_npm/_npm/interval-tree-1d@1.0.4/+esm.js
npm:_npm/interval-tree-1d@1.0.4/+esm.js →
GET /_npm/jszip@3.10.1/+esm.js
Error: unable to fetch: https://cdn.jsdelivr.net/npm/_npm/d3@7.8.5/+esm
    at file:///Users/fil/Source/framework/src/javascript/imports.ts:371:29
    at processTicksAndRejections (node:internal/process/task_queues:95:5)
Error: unable to fetch: https://cdn.jsdelivr.net/npm/_npm/d3@7.8.5/+esm
    at file:///Users/fil/Source/framework/src/javascript/imports.ts:371:29
    at processTicksAndRejections (node:internal/process/task_queues:95:5)
```

it looks like the root cause is that the path is repeating `_npm/_npm/` twice.

The issue seems to go away by modifying src/javascript/imports.ts:

```diff
-        value = relativeUrl(path, value);
+        value = relativeUrl("/_npm/" + path, value);
```

There's probably a cleaner way to do this. We should also return a 404 or 500 error when the path invoked under _npm does not exist on npm (or if the connection fails for random lack of luck).

mbostock:
Looks like I broke something when I switched the cache directory location (but that I hadn’t noticed yet because I have everything cached because I’ve been mostly offline). I’ll fix it — thanks for letting me know!

src/build.ts
```diff
@@ -157,9 +160,46 @@ export async function build(
     await effects.copyFile(sourcePath, outputPath);
   }
 
-  // Copy over the imported modules.
+  // Resolve some special additional downloads.
```
mbostock, line 163: I think this means we’ll need a config option that allows you to enumerate extra files to download from npm as part of build. (We don’t need them in preview because we can lazily fetch them when the code is evaluated.) But we can add that as a future enhancement.

docs/lib/vega-lite.md
```diff
@@ -25,33 +25,17 @@ Or you can use a [Vega-Lite JSON view specification](https://vega.github.io/vega
...
-        "type": "quantitative",
-        "field": "Anomaly",
-        "scale": {
-          "range": "diverging",
-          "reverse": true
-        }
-      }
+    width: 640,
+    height: 400,
+    data: {url: await FileAttachment("gistemp.csv").url(), format: {type: "csv"}},
```
mbostock, line 30: This is needed because Vega-Lite infers the format type from the file extension (`.csv`), but doesn’t check for the query string and so is confused by the now-present `?sha=…`. It’s better to be explicit about the format anyway.

mbostock:
Okay, so the latest commit uses Mermaid’s provided bundle, rather than relying on jsDelivr to compute it. This means that Mermaid’s dependencies (such as D3) are baked-in to the Mermaid bundle, but that’s relatively efficient and it’s what they recommend — and perhaps most usefully it bypasses jsDelivr’s timeouts. This again means that we’ll have to download all of Mermaid’s plugins during build, but I think that’s the best option since we can’t statically determine what you’ll need.

I also re-enabled self-hosting of dynamic imports.

src/search.ts
```diff
@@ -48,7 +48,7 @@ export async function searchIndex(config: Config, effects = defaultEffects): Pro
     const listed = pagePaths.has(`/${file.slice(0, -3)}`);
     const indexed = data?.index === undefined ? listed : Boolean(data.index);
     if (!indexed) {
-      if (listed) effects.logger.log(`${faint("skip")} ${file}`);
+      if (listed) effects.logger.log(`${faint("index")} ${strikethrough(path)} ${faint("(skipped)")}`);
```
Fil, line 51: neat :)

mbostock:
I see an outstanding issue with the content hashes of files generated by data loaders: the hashes are of the data loader rather than the generated file. That’s useful during preview because it ensures that we re-request the file when the data loader changes. But in build it’s undesirable because the hashes don’t change unless the data loader changes (and data loaders by nature are not expected to return the same data every time you run them!). This feels related to the #699 — if we can generate the files first, then we could use the content hash of the files in the data loader cache.

mbostock:
I guess the data loader hashes being wrong isn’t an issue yet until we want to guarantee cache-breaking, in which case we’ll want to put the hashes into the file names rather than the query string.

mbostock:
Okay! I think this is all ready to go and I’m excited about it! 🚀

There’s still some room for future improvement, such as content-hashed names for built-in modules (_e.g._, `/_observablehq/client.js`) but these at least should only change when you upgrade Framework. I also thought it might be easier to put the Framework version number in the path such as `/_observablehq@1.0.0/client.js` a bit like we do with npm imports. But I’m leaving that for future work! Let’s do this.

src/client/search.js
```diff
@@ -8,7 +8,7 @@ const resultsContainer = document.querySelector("#observablehq-search-results");
 const activeClass = "observablehq-link-active";
 let currentValue;
 
-const index = await fetch(import.meta.resolve("./minisearch.json"))
+const index = await fetch(import.meta.resolve(global.__minisearch))
```
mbostock, line 11: We could use the `importMetaResolve` Rollup plugin to say `import.meta.resolve("observablehq:minisearch.json")` here, but that’s just polish.

Fil: Unlike `/_observablehq/client.js ` which changes only when we update framework, the contents of `/_observablehq/search.js` change about each time we build the site.

test/output/build/search-public/page1.html
```diff
@@ -22,7 +22,7 @@
     <label id="observablehq-sidebar-close" for="observablehq-sidebar-toggle"></label>
     <li class="observablehq-link"><a href="./">Home</a></li>
   </ol>
-  <div id="observablehq-search" data-root="./"><input type="search" placeholder="Search"></div>
+  <div id="observablehq-search"><input type="search" placeholder="Search"></div>
```
Fil, line 25: nice!

Fil (approved):
this is great!

src/client/search.js
mbostock: The latest commit implements `import.meta.resolve` resolution so you can now say `fetch(import.meta.resolve("observablehq:minisearch.json"))`. 🥳

mbostock:
Seems like GitHub is broken and the latest commit 51dfd9acdb29efeb003a22e389d2f2a7c95d438b implementing `import.meta.resolve` was not showing up, so I manually squash-merged as de4fe09b57661b326647695a5b3d0654ed5ced00.

Fil:
import.meta.resolve doesn't seem to work for minisearch.json:

````md
```js
import.meta.resolve("observablehq:minisearch.json");
```
````

results in:

```
build node_modules/@observablehq/framework/src/client/minisearch.json →
Unexpected error: Could not resolve entry module "node_modules/@observablehq/framework/src/client/minisearch.json".

```

mbostock:
I tested it and it worked fine, so I dunno @Fil. Can you isolate a reproduction and file a bug if it doesn’t work? Thank you.

Fil:
filed under https://github.com/observablehq/framework/issues/948

PR #844, opened 2024-02-17
sql cells, backed by DuckDB

### LINKED ISSUE #48: SQL fenced code blocks
Depends on DatabaseClient #46. A client-side alternative to a data loader #38, presumably. Depends on #8 to implement tagged template literal code blocks, plus some additional directives to indicate which database you’re querying, and what name the result set should be exposed as.

mbostock:
Fixes #48. TODO

- [x] `sql` tagged-template literal implemented by `npm:@observablehq/duckdb`
- [x] `sql` needs implicit dependency on duckdb
- [x] `sql` literal needs a shared `DuckDBClient` under the hood (not a new one for each query!)
- [x] `sql` frontmatter option to declare which tables should be available to `sql`
- [x] generate `registerTable` calls, or some such, to register these tables for `sql`
- [x] a way to name (`id`) the output of a `sql` cell for access in JavaScript?
- [x] throw a syntax error if the specified `id` is invalid
- [x] destructuring assignment (_e.g._, `id=[{foo}]`)
- [x] register table sources as files in `getResolvers`
- [x] live update of `sql` when front matter is edited during preview
- [x] `display` directive
- [x] display as table rather than the default inspector
- [x] ~~support DuckDB files as source?~~ [future]
- [x] re-register tables when file attachments change
- [x] documentation

I’m imagining the frontmatter would look something like this:

```md
---
sql: chinook.db
---
```

But possibly you can declare the tables individually:

```md
---
sql:
  customers: customers.csv
  orders: orders.parquet
---
```

And in any case we’ll need to make sure you can use a data loader to generate the backing files (_e.g._, `customers.csv.js`). And ideally we’d correctly handle incremental edits to the frontmatter during preview, un-registering and re-registering tables, and hot module replacement to the tables 😅 with file watching. But if we have to reload the page in the interim I think this would still be a great start.

A challenge with `sql.init` will be avoiding the race condition — if we start to process a `sql` query before we’ve registered the tables, the results will be wrong. But if we’re going to do reactive updates to the `sql` definition anyway, maybe the answer is that `sql` should be defined as a generator, and any call to register or unregister a table will cause this generator to yield a new value, and then reactively reevaluate all the downstream queries. That sounds like the right path since it leverages reactivity (though it might mean some brief flashes of “wrong” results?).

One way to implement a better display for `sql` cells would be to use a `sql.table` tagged-template literal instead of `sql` (like we do for `tex.block`). We could also change the generated code under the hood so instead of this:

```js
sql`SELECT 1 + 2`
```

it’s more like this:

```js
const foo = sql`SELECT 1 + 2`;
display(Inputs.table(foo));
```

As for naming the output, that could be something like:

````md
```sql id=foo
SELECT 1 + 2
```
````

### CODE REVIEW

**Comment by Fil** [comment]
Neat! I wonder if sql is the right name for the list of data sources; it could be interesting to make it more generic, since these data sources could be consumed by something else than sql in the future (maybe… a smart data table)?

**Comment by mbostock** [comment]
I want to use the name `sql` because it matches the `sql` fenced code block and the `sql` template literal. But I could see wanting to change the engine to _e.g._ SQLite, so maybe that means a more explicit form like

```md
---
sql:
  engine: sqlite
  tables: chinook.db
---
```

As for the display, I think we would change <code>```sql</code> to have a better default display in the future? I’m not sure that needs to be configurable, but I guess it could be an attribute of the fenced code block or page-level option too.

Fil:
The interactive view tends to demultiply the number of tables

https://github.com/observablehq/framework/assets/7001/db1993a8-e110-42b3-878d-5b43261f9dd3

Fil:
Feature requests (probably for future iterations):

- what would be the easiest way to apply this on a JavaScript array instead of a file? For example to work with live data fetched (and data-wrangled) from an API.

We'll probably need to add some helper code similar to this snippet?

          const table = arrow.tableFromJSON(data);
          const buffer = arrow.tableToIPC(table);
          const conn = await db.connect();
          await conn.insertArrowFromIPCStream(buffer, {
            name,
            schema: "main",
            ...options
          });


Also, I would love to use an absolute url as a source (to query data from files hosted somewhere else).

For example:

````md
```yaml
sql:
  adresses: https://static.data.gouv.fr/resources/bureaux-de-vote-et-adresses-de-leurs-electeurs/20230626-135723/table-adresses-reu.parquet
```
````

(this currently fails with  **Error**: Invalid Error: Opening file 'table-adresses-reu.parquet' failed with error: Failed to open file: table-adresses-reu.parquet)

**Comment by mbostock** [comment]
Yes, I mentioned the race condition in Slack. Is because there promises resolve out of order, and you can get a display from an older query after an earlier query. We need to handle this race condition in the display function (by ignoring the call). I can work on this later today but feel free to take a crack at it if you’re interested.

**Comment by mbostock** [comment]
On your requested enhancements, don’t forget that you can just override the definition of `sql` on a page, so you can use the full capabilities of `DuckDBClient` to point to in-memory or remote data.

```js
const db = await DuckDBClient.of({gaia: FileAttachment("./lib/gaia-sample.parquet")});
const sql = db.sql.bind(db);
```

mbostock:
On the race condition, I have filed a bug and a simple reproduction in #995 (which confirms that it’s a pre-existing problem, so we could ignore it in this PR for now — say by switching to a radio button instead of a range slider). It’ll also be a bit tricky to fix, I expect; see the linked issue for discussion.

Fil:
> guess the format from the file extension

I could imagine passing it as an option in a long-form description of the source, for when it's opaque:

```yaml
sql:
  - tableName:
    - source: https://<something opaque>
    - format: parquet
```

Fil:
> const sql = db.sql.bind(db);

I would not have guessed this invocation, but we can add syntactic sugar later.

docs/sql.md
```diff
@@ -0,0 +1,148 @@
...
+```js echo
+top
+```
+
+To pull out the minimum value of the `phot_g_mean_mag` column:
+
+````md
+```sql id=[{min}]
+SELECT MIN(phot_g_mean_mag) AS min FROM gaia
+```
```
Fil, line 86: I would have liked this to be extensible to define multiple values, but it doesn't seem to work:

````md
```sql id=[{min, max}]
SELECT MIN(phot_g_mean_mag) AS min, MAX(phot_g_mean_mag) AS max FROM gaia
```
````

defines only `min`.

(Not a blocker.)

**Review by Fil — APPROVED** [review]
this is truly magnificent

File: docs/sql.md (hunk 1 above)
Comment (mbostock) [reply docs/sql.md:86]: Hmm, that’s surprising. I would expect that to work and I’ll investigate.

Comment (Fil) [reply docs/sql.md:86]: In fact it works, I was editing the wrong code block! 😭

Comment (Fil) [reply docs/sql.md:86]: We could update that example in the documentation to include min and max?

Comment (mbostock) [reply docs/sql.md:86]: Ah, so the attribute gotcha applies, which means you either need quotes

````md
```sql id="[{min, max}]"
SELECT MIN(phot_g_mean_mag) AS min, MAX(phot_g_mean_mag) AS max FROM gaia
```
````

or to remove spaces

````md
```sql id=[{min,max}]
SELECT MIN(phot_g_mean_mag) AS min, MAX(phot_g_mean_mag) AS max FROM gaia
```
````

**Comment by mbostock** [comment]
We could add this shorthand in the future:

```js
const sql = DuckDBClient.sql({gaia: FileAttachment("./lib/gaia-sample.parquet")});
```

Issue #831, opened 2024-02-16
add java support to dataloaders


cyrilou242:
Hey, 
It'd be nice to have support for java for dataloaders.

As of java 11 it's possible to run a java class directly with: 

```
java Test.java
```

so from what I understand it's only a matter of doing a PR similar to this one: 
https://github.com/observablehq/framework/pull/673/files

If you are ok I can do this for a first version. 

A second iteration would be to think of how dependencies are passed, but I haven't seen how this is done for other languages so maybe you don't want to allow this for the moment? 
A no dev way in java would be to pass it as an env var: 
```
export CLASSPATH=/some/dependencies.jar
```
other option is to pass it as an argument to the java command but this would be more involved. 

Thanks for the nice project!

Fil:
Thanks, we welcome contributions.

Re: environments, we established a clear separation of concerns: setting up the env is _not_ the responsibility of Observable Framework. The `preview`, `build` (and `deploy`) commands only call the corresponding interpreter (or whatever is on the $PATH with that name), passing the loader script as argument. This works well in general for vanilla scripts. Anything beyond that, like requiring modules, extensions, etc, must be prepared in advance. (Of course there is an obvious escape hatch, in that a data loader can be a bash script that does all this work of setting up a venv.)

cyrilou242:
Makes sense thanks. 
I'll try integrating and using java. Will create a PR when it's ready.

PR #846, opened 2024-02-17
add support for java data loaders

cyrilou242:
Add support for java data loaders as discussed here:
https://github.com/observablehq/framework/issues/831

I put a [ForecastJava.json.java](https://github.com/observablehq/framework/compare/main...cyrilou242:framework:main#diff-e5a7882f72679fb0b6628ef66848b4597d6eeaa55e6793b2ab79897491c16df7) to have an example of how to load json in java. It's not used anywhere but it's easy to try be replacing in the getting started 
```
const forecast = FileAttachment("./data/forecast.json").json();
```

by 
```
const forecast = FileAttachment("./data/ForecastJava.json").json();
```
Can be removed if it's not helpful.

### DIFF
--- a/docs/HelloJava.txt.java
+++ b/docs/HelloJava.txt.java
@@ -1,0 +1,5 @@
+public class HelloJava {
+  public static void main(String[] args) {
+    System.out.println("Hello, world!");
+  }
+}
--- a/docs/data/ForecastJava.json.java
+++ b/docs/data/ForecastJava.json.java
@@ -1,0 +1,43 @@
+import com.google.gson.JsonObject;
+import com.google.gson.JsonParser;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.net.HttpURLConnection;
+import java.net.URL;
+
+// requires gson in the classpath. For instance, have the following env var set:
+// export CLASSPATH=/Users/<YOUR_USER_NAME>/.m2/repository/com/google/code/gson/gson/2.8.6/gson-2.8.6.jar
+public class ForecastJava {
+
+  public static void main(String[] args) {
+    double latitude = 37.80;
+    double longitude = -122.47;
+
+    try {
+      JsonObject station = getJson("https://api.weather.gov/points/" + latitude + "," + longitude);
+      JsonObject properties = station.getAsJsonObject("properties");
+      String forecastUrl = properties.getAsJsonPrimitive("forecastHourly").getAsString();
+      JsonObject forecast = getJson(forecastUrl);
+
+      System.out.println(forecast);
+    } catch (IOException e) {
+      e.printStackTrace();
+    }
+  }
+
+  private static JsonObject getJson(String url) throws IOException {
+    HttpURLConnection connection = (HttpURLConnection) new URL(url).openConnection();
+    connection.setRequestMethod("GET");
+    connection.setRequestProperty("Accept", "application/json");
+
+    try (InputStream inputStream = connection.getInputStream();
+        InputStreamReader reader = new InputStreamReader(inputStream)) {
+
+      return JsonParser.parseReader(reader).getAsJsonObject();
+    } finally {
+      connection.disconnect();
+    }
+  }
+}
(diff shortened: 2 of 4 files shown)


docs/loaders.md
```diff
@@ -145,12 +146,13 @@ For example, for the file `quakes.csv`, the following data loaders are considere
 - `quakes.csv.R`
 - `quakes.csv.rs`
 - `quakes.csv.go`
+- `quakes.csv.java`
 - `quakes.csv.sh`
 - `quakes.csv.exe`
 
-If you use `.py`, `.R`, `.rs`, or `.go`, the corresponding interpreter (`python3`, `Rscript`, `rust-script`, or `go run`, respectively) must be installed and available on your `$PATH`. Any additional modules, packages, libraries, _etc._, must also be installed before you can use them.
+If you use `.py`, `.R`, `.rs`, `.go` or `java`, the corresponding interpreter (`python3`, `Rscript`, `rust-script`, `go run` or `java`, respectively) must be installed and available on your `$PATH`. Any additional modules, packages, libraries, _etc._, must also be installed before you can use them.
```
mbostock, line 153:
```suggestion
If you use `.py`, `.R`, `.rs`, `.go`, or `java`, the corresponding interpreter (`python3`, `Rscript`, `rust-script`, `go run`, or `java`, respectively) must be installed and available on your `$PATH`. Any additional modules, packages, libraries, _etc._, must also be installed before you can use them.
```

docs/loaders.md
```diff
@@ -145,12 +146,13 @@ For example, for the file `quakes.csv`, the following data loaders are considere
...
 
-Whereas `.js`, `.ts`, `.py`, `.R`, `.rs`, `.go`, and `.sh` data loaders are run via interpreters, `.exe` data loaders are run directly and must have the executable bit set. This is typically done via [`chmod`](https://en.wikipedia.org/wiki/Chmod). For example:
+Whereas `.js`, `.ts`, `.py`, `.R`, `.rs`, `.go`, `.java` and `.sh` data loaders are run via interpreters, `.exe` data loaders are run directly and must have the executable bit set. This is typically done via [`chmod`](https://en.wikipedia.org/wiki/Chmod). For example:
```
mbostock, line 155:
```suggestion
Whereas all other data loaders are run via interpreters (such as `node` for `.js`), `.exe` data loaders are run directly and must have the executable bit set. This is typically done via [`chmod`](https://en.wikipedia.org/wiki/Chmod). For example:
```

I think we have enough data loaders now that it makes sense to refer to them in aggregate and only call out `.exe` is being special.

docs/loaders.md
```diff
@@ -134,6 +134,7 @@ Data loaders live in the source root (typically `docs`) alongside your other sou
 - `.R` - R (`Rscript`)
 - `.rs` - Rust (`rust-script`)
 - `.go` - Go (`go run`)
+- `.java` - Java >=11 (`java`)
```
mbostock, line 137:
```suggestion
- `.java` - Java (`java`; requires Java 11+ and [single-file programs](https://openjdk.org/jeps/330))
```

docs/HelloJava.txt.java
mbostock, line 1: Can we rename this file to `hello-java.txt.java` to match the Go and Rust “hello world” loaders? I tested it and it appeared to work, but I don’t know if that’s a recent relaxation of naming constraints. I also tested making the class non-public (just `class HelloJava`) and that worked too, and I think that’s always be allowed.

docs/data/ForecastJava.json.java
mbostock, line 1: Per the other comment, ideally this would be named `forecast.json.java` so that a Java data loader can serve the same `forecast.json` rather than revealing it’s Java-ness to the client.

mbostock (changes requested):
Thanks for this! 🙏

Fil:
wrapped under #935

Fil:
solved in https://github.com/observablehq/framework/pull/935

cyrilou242:
Thanks for this!

Issue #856, opened 2024-02-18
Getting started - Display Forecast gets Promise not Object


bertt:
Hi, I'm following the getting started on WSL / Node v20.11.1 (https://observablehq.com/framework/getting-started) but get an Promise when doing the 'display(forecast)':

```
const forecast = FileAttachment("./data/forecast.json").json();
display(forecast);
```

![image](https://github.com/observablehq/framework/assets/538812/98c1c25a-7837-4f1a-85bb-2c136e3d768d)

Function 'node docs/data/forecast.json.js' works ok.

Any idea how to get the weather object showing up?

Fil:
Maybe we should push https://github.com/observablehq/framework/pull/356 a bit further and make display await Promises.

````
```js
const forecast = FileAttachment("./data/forecast.json").json();
display(await forecast);
```
````

bertt:
ah yes that works thanks

bertt:
https://github.com/observablehq/framework/pull/857

Fil:
Thanks for reporting this; since the copy says “another code block”, it's not technically wrong (not a bug). But there's room for improvement—either in the docs or in the way we display promises.

PR #857, opened 2024-02-18
Update getting-started.md - await weather forecast

bertt:
Fixes https://github.com/observablehq/framework/issues/856


Fil:
The copy says “another code block” so not technically wrong. But since it's easy to gloss over this, we might want to call this out by adding a line to say “if called in the same block, you will have to explicitly await the Promise like so…".

mbostock:
I don’t think we should do this. It’s typically an antipattern (since the await forces serialization when it’s better to load files in parallel).

mbostock:
Even though we’re not accepting it, thank you for the suggestion! 🙏

Issue #850, opened 2024-02-17
Compile to JavaScript and drop runtime dependency on tsx


mbostock:
I love tsx during development, but I think we should compile to JavaScript on publish and drop the runtime dependency. This would allow us to support Node 18+ instead of requiring Node 20.6+, and it would fix confusing errors such as #847. It would just be “fewer moving parts” and more robust, and it shouldn’t be too technically challenging to run esbuild or tsc or whatever on publish.

mbostock:
Some things I learned so far in #963:

- Use `find` instead of glob (ref. https://github.com/evanw/esbuild/issues/263#issuecomment-660405381)
- Drop `--target=esnext` because [it’s the default](https://esbuild.github.io/api/#target)
- Add `--platform=node` because we’re not [building for the browser](https://esbuild.github.io/api/#platform)

In general it seems to work well, but the fact that the generated sources live in a different directory I think are going to cause problems for methods such as `getClientPath` which assume they are in `src/client/`. I also would like our testing and local development environment to be as close as possible to what we publish, so that we are less likely to have issues that only affect the published distribution. esbuild has some lovely built-in watch and serving features for local development, which makes me wonder if we couldn’t just drop tsx entirely? But I also want to minimize churn, so my primary goal for this week is to get a no-tsx distribution working.

Issue #783, opened 2024-02-14
Things break when a file name contains emoji


dunstan:
I put an emoji in the filename and the preview of the emoji page rendered but the command line wasn't happy. I was also able to deploy it, but when I tried to view the emoji page on staging I got a "not found" error.

What do we think should happen here? Is this a bug?

<img width="1032" alt="Screenshot 2024-02-13 at 15 26 40" src="https://github.com/observablehq/framework/assets/1316196/dade147e-7bbe-4884-8a13-da815fc5d338">
<img width="1112" alt="Screenshot 2024-02-13 at 15 27 29" src="https://github.com/observablehq/framework/assets/1316196/8392eae0-7785-4362-9188-68aeed8ef1e1">

![Screenshot 2024-02-13 at 15 26 35](https://github.com/observablehq/framework/assets/1316196/a5df91e9-dbb7-42eb-a2bc-f023187d5118)

![image (20)](https://github.com/observablehq/framework/assets/1316196/44c60e6e-c5b3-4a60-a8a3-88adcf05296f)

@Fil [also notes that he can put a < character in a file name](https://observablehq.slack.com/archives/C05UDG0BRDY/p1707866315692459).

Fil:
related #778

PR #881, opened 2024-02-20
Distinguish between URL paths and file paths at the type level

mythmon:
This is an attempt to improved Windows support. The idea here is that we can use TypeScript to tag some strings as being either `FilePath`s or `UrlPath`s. From there we can make a series of helpers around the Node filesystem and path functions that are stricter about which one they take.

I'm putting this up as a draft now to get some early feedback about this approach. I think there is some inherent boilerplate here to tag strings as the two types. I think there may also be some boilerplate in this PR that we could get rid of. Any comments or suggestions about how we could improve this are welcome!

As of now, I have all the tests passing on Windows locally.

Besides the path annotations, I've introduced some libraries to ease in cross-platform compatibility:
- `cross-env` - Handles setting environment variables inline in `package.json` scripts
- `rimraf` - A cross platform `rm`-like tool
- `cross-spawn` - For executing files more easily in a cross platform way.

I'm not sure how we can get `.sh` data loaders working, so for now I've simply disabled those tests on Windows. I think we can save that for another PR.

In order for our snapshot tests to work, I added a `.gitattributes` file that insists on posix-style line endings everywhere. I based that off of [these GitHub docs](https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings) and [this blogpost](https://adaptivepatchwork.com/2012/03/01/mind-the-end-of-your-line/).

---

# To Test

- Check out this branch
- Run `yarn install` to update packages
- Run `yarn observable create` to create a new project
- `cd` into the project directory
- Edit the project to use the branch instead of the npm published version of Framework:
  - `yarn remove @observablehq/framework`
  - `yarn add link:..`
- Use Framework with commands like `yarn dev`, `yarn build`, `yarn observable login` and `yarn deploy`.


Fil (commented):
two small questions for now

src/config.ts
```diff
@@ -49,16 +50,16 @@ export interface Config {
   search: boolean; // default to false
 }
 
-export async function readConfig(configPath?: string, root?: string): Promise<Config> {
+export async function readConfig(configPath?: FilePath, root?: FilePath): Promise<Config> {
   if (configPath === undefined) return readDefaultConfig(root);
-  const importPath = join(process.cwd(), root ?? ".", configPath);
+  const importPath = "file:///" + filePathToUrlPath(fileJoin(process.cwd(), root ?? ".", configPath));
```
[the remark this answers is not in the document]
mythmon: I'm not quite sure what you mean. The original version already called `join(process.cwd(), ...)`, so the `importPath` would be an absolute path. Windows _does_ insist on `file://` protocol here, but I think there is a relative form we could use.

src/brandedPath.ts
```diff
@@ -0,0 +1,101 @@
...
+}
+
+export function unFilePath(path: FilePath | string): string {
+  return path as unknown as string;
+}
+
+export function unUrlPath(path: UrlPath | string): string {
+  return path as unknown as string;
+}
+
```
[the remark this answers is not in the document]
mythmon: One thing that I was disappointed about was that I couldn't figure out how to safely tell TypeScript that `UrlPath` and `FilePath` were actually just type safe strings. If we remove the overrides of `replace` and similar in the definitions of the path types, it can understand it, but then we erase types when working on the path.

I think if we wanted to eliminate these functions, we'd have to inline the `as unknown as string` text everywhere we call this function. I chose not to do that because I treat `as`-casts as dangerous code. I'd prefer to isolate it to as few places as possible. Then those isolated places can be thoroughly vetted, either through tests or through reading the code. In this case the cast is trivial to verify, so just having this function is enough.

I agree though that I would prefer some better way to do this, like maybe a "type level function". I don't know of one though.

Fil (commented):
If I understand correctly, this PR includes 3 changes that are almost independent:
1. generalizes some system call (`rimraf` and `cross-env`)
2. types all paths as either `filePath` or `urlPath`
3. carries `path.sep` into the internal representation of file paths

(It also adds platform-specific tests—in CI, and in unit tests.)

Generalizations/abstractions seem like the right thing to do. Adding platform-specific tests is obviously needed.

For step 2, I think it's great that you're typing these paths as "filesystem path" or "url path", as it helps bring clarity to the code, where it's easy otherwise to get a bit confused between what the paths are for. It certainly helps with step 3. However in my mind it's almost an orthogonal concern.

Step 3 feels overwhelming. Our app code now sees different strings depending on the OS it runs on, and has to reason about them differently. On windows, file paths will be a string `docs\sub\file.md`, and on linux an macos it will be `docs/sub/file.md`. This representation is carried throughout, meaning that at every step we need to convert between filepath and urlpath (which always use `/`). In a sense, now every line of code needs to know about windows.

Could we consider an approach where we would _normalize the internal representation_ of all paths to forward slashes, and adapt/convert only when talking to the filesystem? Basically an abstraction layer on top of fs that would use `/`-delimited paths as inputs and outputs, restricting the use of `\` to the underlying OS calls.

An analogy that comes to mind is to use an ORM if we want to talk to different database flavors. If that is possible, it means we could do a much smaller step 3.

More importantly, when some logic needs to reason about a path, say "if this file is foo/bar, do something", it can be written directly rather than having to call conversion functions. For a concrete example: when we create the search index (in `search.ts`), we attribute an `id` to each page, which corresponds to its relative path to the docs root. That id is also used as a `link[href]` when we present the results in the browser. I’d rather keep this logic insulated from having to care about windows-style paths.

If that approach proves possible, we could even skip step 2 altogether for now (and do it separately later, independently of windows support).

We could also land step 1 separately.

Finally, abstracting `fs` might also help with unicode file names (for international text or even emoji #783).

src/config.ts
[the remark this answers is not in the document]
Fil: Ah yes it was already absolute. As a nit if we must start with`file:///` we should add only two `//` since the third one is the first char of `process.cwd()`. Currently I see 4 of them :)

`file:////Users/fil/Source/framework/observablehq.config.js`

mythmon:
Thanks for the comments, Fil.

Your first two summary points I agree with. I'm not sure if it makes sense to split 1 out, since the project is still quite broken even with those changes in place.

I don't quite agree about the third though. `path.sep` was already in the internal representation of these paths when running on Windows. However, it was there in an uncontrolled, chaotic way. When working through the test cases, I occasionally found paths like `foo/bar\baz` where different slashes had been mixed within one path. What this PR does is make the existing code explicitly handle that.

I like the idea of using internally normalized paths.

For me the biggest hurdle is that it would require writing wrappers around all of `node:path` and `node:fs`. However, as a part of this implementation I ended up writing writing these kind of wrappers anyways in the form of the branded path and fs modules.

The other concern I have is about the appropriateness of abstraction. Framework has to deal with both files and HTTP very directly. There are a few places where we need to understand the differences between win32 and posix paths. I'm not sure if that is fundamental complexity, or if we would prefer to work in an abstracted space where we don't have to worry about it.

Given that, I plan to continue with this "slash-aware" version, and get it passing in CI. After I'm done with that, I can explore the idea of internally normalized paths, if someone else doesn't beat me to it.

Regarding the `FilePath`/`UrlPath` annotations: Given that the work is fundamentally done in that area, I think it would make sense to keep it. Even when we aren't mixing slashes, it made the intent of some of the code clearer for me.

Fil:
> path.sep was already in the internal representation

Claro. (I did not mean to imply that it was working before.) My question is whether we can control it at the file system interface, and continue ignoring it in the app code (but, confident in that it is safe and controlled).

test/dataloaders-test.ts
```diff
@@ -1,95 +1,98 @@
...
-        "[stale] "
-      ]
-    );
-  });
+  };
+  const loader = Loader.find(FilePath("test"), FilePath("dataloaders/data1.txt"))!;
+  // save the loader times.
+  const {atime, mtime} = await stat(loader.path);
+  // set the loader mtime to Dec. 1st, 2023.
+  const time = new Date("2023-11-01");
```
mythmon, line 63: I changed this from "seconds since epoch" to a Date because non-date values throw an error on Windows: https://github.com/nodejs/node/issues/5561#issuecomment-192306799

examples/eia/docs/data/eia-bia-reference.csv
mythmon, line 1: I converted this file to use LF endings instead of CRLF. If modifying this file isn't ok, I could add an exception to `.gitattributes` for it.

mythmon:
With this branch, I was able to create, preview, build, login, and deploy in Windows. It's a big change, but I think most of the changes only affect their immediate area.  It works for me, and it works on CI, though I would certainly appreciate more testing on this, both on Windows machines as well as Non-Windows.

I'm going to start another PR based on this to explore Fil's idea of using forward slashes everywhere internally, and only converting to OS-correct paths at the last moment.

src/brandedPath.ts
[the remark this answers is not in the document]
visnup: All of the wrapping worries me, but I can't express why. It feels excessive or invasive maybe? I'm definitely wishing we didn't have to do it.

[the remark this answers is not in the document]
mythmon: This pattern, of having a new type that is just an existing type with a new name is something that I've used in other languages. In those languages there are often nice features built into the language that help make this easier, like implicit conversion, operator overloading, and monad-style-map. TypeScript doesn't really give us any of those things, so it is all a bit clunky.

I could imagine taking out most of these wrappers now (except for the places where we need slash conversion). I don't know how we would prevent new changes to the code base from getting it wrong in the future. Maybe just test coverage and Windows in CI? That feels much less robust than TypeScript to me.

GordonSmith:
Just a lurker here - and don't want to get into the whole functional v OOP debate - but wouldn't this change be a lot cleaner with a FilePath Class and more use of the Node URL Class?

mythmon:
@GordonSmith thanks for the comments. I'm not sure if a class based approach would help. The original intent was to do this entirely at the TypeScript, except where conversion was needed. It would be nice if this had as little runtime overhead as possible. I'm not sure how well that worked out though. I think that a class would only be a change of "spelling" without really changing the overall cleanliness.

I'm working on an alternative approach that fill Fil recommended that doesn't use any wrapping at all. I think it is promising, but we'll see once I get it working.

As for the URL class, as far as I know that requires full, absolute URLs, including hostname. I don't think that would work well here.

.gitattributes
mbostock, line 1: I think we should disable autocrlf.

```
* text=false
```

bin/observable-init.js
mbostock, line 1: I want to remove our runtime dependency on tsx. #850 Also, we’ll need the same fix in the `@observablehq/create` repo.

examples/eia/docs/data/eia-bia-reference.csv
mbostock, line 1: Let’s leave these as-is to minimize this PR.

package.json
```diff
@@ -92,6 +94,7 @@
     "c8": "^8.0.1",
     "chai": "^4.3.10",
     "chai-http": "^4.4.0",
+    "cross-env": "^7.0.3",
```
mbostock, line 97: If this is in `dependencies` already, we shouldn’t also need it in `devDependencies`.

```suggestion
```

mythmon:
Closing in favor of #944.

PR #886, opened 2024-02-21
prompt for project access level on creation

mythmon:
CC @Fil who always has better word smithing than I do.

image

image


mbostock:
We should not default to public. Also I think this isn’t the right place to prompt for visibility — better to create the project as private, then visit the deployed project and confirm that it looks good, and then make it public or invite others.

In short I don’t think we should make this change. I don’t think it’s right from a product perspective. I’m also highly sensitive to making it too easy to make projects public (since our business emphasis is on work and the private use case, and these projects can highly sensitive!).

mythmon:
Sounds good. I'd like to keep a couple small things out of this, like the changes to the deploy order and outro. I'll put them into other PRs.

mbostock:
Maybe a good easy thing to do would be to log that the project is initially private and show the URL to visit (or instructions) to change permissions or invite people to the project.

PR #891, opened 2024-02-21
draft: true

### LINKED ISSUE #867: draft mode
A front matter option such as ~~`status: draft` or~~ `draft: true` makes a page visible in preview, but skipped during build.

The `--drafts` flag controls whether we build them or not (defaults to false for build and deploy, and true for preview).

drafts should not be indexed nor listed in the sidebar (unless set in pages).

Fil:
Introduces a **draft** booleanish option in front matter, which specifies if a page is a draft.

Drafts are rendered in preview, but skipped in build (and deploy). When previewing a draft, a tag indicates the draft mode:
draft mode

I played a bit with the design of this tag, went back and forth between a pill with a focus background, and this simpler design.

I considered `status: draft` but looking at what other SSG do `draft: true` seems more common.

I also considered having a `drafts` global option that preview would activate; but it introduces a lot of complexity (we have to carry it to the config and to build the pages), for nothing really useful.

The only place I could document it was the "front matter" section of [/markdown](https://observablehq.com/framework/markdown#front-matter), which had nothing, so I documented the other options as well.

closes #867


mbostock:
How would you feel about `build: false` instead of `draft: true`, and to have no visible indication on the page that it’s not built (_i.e._, `build: false` simply disables build without altering the appearance of the page)? That feels easier for me to approve since then I don’t have to think about what the draft badging should look like. But it’s also less opinionated about when `build: false` should be used, unlike `draft: true` which suggests a more opinionated workflow.

Fil:
The badge is just a class that we add to the main container, styled with css, so it's easy for folks to remove it or do something different. It also only appears only in preview so it has no impact on the built site (save for a few css styles, that we could load separately).

Since `draft: true` is more common across SSGs, I think it's worth adopting it over `build: false`, even if it means taking more time to hash out what it does exactly.

We have to be opinionated in a sense, because we wouldn't want a page that is not built but gets indexed or listed as part of the default pages in the sidebar.

docs/markdown.md
```diff
@@ -13,6 +13,13 @@ toc: false
 ---
 ```
 
+The front matter supports the following options:
+
+- **title** — the title, defaults to the (first) first-level heading of the page, if any
+- **toc** — if false, disables the [table of contents](./config#toc)
+- **index** — if [search](./search) is enabled, should this page be indexed (defaults to true for pages listed in the sidebar, and false for any other page)
+- **draft** — whether this page is a draft; drafts are rendered in preview but skipped during build and deploy; drafts are not included in the default sidebar.
```
mbostock, line 21:
```suggestion
- **draft** — whether to skip this page during build; drafts are also not listed in the default sidebar
```

docs/markdown.md
```diff
@@ -13,6 +13,13 @@ toc: false
 ---
 ```
 
+The front matter supports the following options:
+
+- **title** — the title, defaults to the (first) first-level heading of the page, if any
+- **toc** — if false, disables the [table of contents](./config#toc)
+- **index** — if [search](./search) is enabled, should this page be indexed (defaults to true for pages listed in the sidebar, and false for any other page)
```
mbostock, line 20:
```suggestion
- **index** — whether to index this page if [search](./search) is enabled; defaults to true for listed pages
```

docs/markdown.md
```diff
@@ -13,6 +13,13 @@ toc: false
 ---
 ```
 
+The front matter supports the following options:
+
+- **title** — the title, defaults to the (first) first-level heading of the page, if any
```
mbostock, line 18:
```suggestion
- **title** — the page title; defaults to the (first) first-level heading of the page, if any
```

src/build.ts
```diff
@@ -73,6 +73,10 @@ export async function build(
     effects.output.write(`${faint("parse")} ${sourcePath} `);
     const start = performance.now();
     const page = await parseMarkdown(sourcePath, options);
+    if (page?.data?.draft) {
+      effects.logger.log(yellow("skipping [draft]"));
```
mbostock, line 77:
```suggestion
      effects.logger.log(faint("(skipped)"));
```

This would match what we do for search indexing.

src/markdown.ts
```diff
@@ -27,6 +27,7 @@ export interface MarkdownPage {
   title: string | null;
   html: string;
   data: {[key: string]: any} | null;
+  draft?: true;
```
mbostock, line 30:
```suggestion
```

src/markdown.ts
```diff
@@ -297,6 +298,7 @@ export async function parseMarkdown(
     html,
     data: isEmpty(parts.data) ? null : parts.data,
     title: parts.data?.title ?? findTitle(tokens) ?? null,
+    draft: parts.data?.draft,
```
mbostock, line 301: It shouldn’t be necessary to expose this here, since it’s already exposed via `page.data`.

```suggestion
```

src/render.ts
```diff
@@ -66,7 +67,7 @@ ${preview ? `\nopen({hash: ${JSON.stringify(resolvers.hash)}, eval: (body) => ev
     toc.show ? html`\n${renderToc(findHeaders(page), toc.label)}` : ""
   }
 <div id="observablehq-center">${renderHeader(options, page.data)}
-<main id="observablehq-main" class="observablehq">
+<main id="observablehq-main" class="observablehq${draft ? " observablehq-draft" : ""}">
```
mbostock, line 70: Since the “draft” state is a modifier rather than representing a unique element, it should have two hyphens.

```suggestion
<main id="observablehq-main" class="observablehq${draft ? " observablehq--draft" : ""}">
```

src/style/global.css
```diff
@@ -276,3 +276,13 @@ a[href].observablehq-header-anchor {
 .muted {
   color: var(--theme-foreground-muted);
 }
+
+.observablehq-draft::before {
```
mbostock, line 280:
```suggestion
.observablehq--draft::before {
```

mbostock (changes requested):
I think the “DRAFT” badge is a little ugly and want to make it prettier (or remove it) before landing. For example we could put “[DRAFT]” in the page title instead.

Fil:
I've added "[DRAFT] " before the first-child (which is usually the title, but there may be no title, or there may be something else. I left the focus color on, but we could remove it, too.

example:
Capture d’écran 2024-03-01 à 17 38 45

src/style/global.css
```diff
@@ -276,3 +276,8 @@ a[href].observablehq-header-anchor {
 .muted {
   color: var(--theme-foreground-muted);
 }
+
+.observablehq--draft > :first-child::before {
+  content: "[DRAFT] ";
```
mbostock, line 281: Slight preference to put it _after_ the title rather than before, and use muted instead of focus.

mbostock (approved):

src/style/global.css
Fil: Putting it after is not great if the first-child is (say) an image or a paragraph.

image-after
paragraph-after


and using first-child is probably not a great idea because… 🙃 (see screen cap below)

https://github.com/observablehq/framework/assets/7001/9a90e695-f14e-41f9-b755-9cdfc6984271

mbostock: Sorry, I assumed this was for `h1:first-of-type`, not `:first-child`. I think the [draft] should only apply to the h1.

Fil: `.observablehq--draft > h1:first-of-type::after` it is!

mbostock (approved):
💯

Issue #877, opened 2024-02-20
'copied' information badge


imhalid:
What do you think about adding a small notification to the copy button to improve the user experience?

I prepared a small demo for this. You can see it in the videos below.

#### Without Badge 

https://github.com/observablehq/framework/assets/40598819/679e37a0-3120-4ec9-a047-04212bab57d3

#### With Badge

https://github.com/observablehq/framework/assets/40598819/961e474c-b72b-4e56-96d6-0ffb5603a23e

mbostock:
Fixed in #893. Thank you!

Issue #865, opened 2024-02-19
i18n


Fil:
To support internationalization, we might need:
- a **lang** option (for a project and in front matter), to inform the [lang](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/lang) attribute of the document
- a **dir** option (or lang_dir) for the [dir](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/dir) attribute — necessary for right-to-left writing systems (Arabic, Farsi…).
- tracking the lang and dir of each page for the sidebar and search results
- translation files for the few keywords we have (“Search”, “Previous”, “Next“…), maybe using  .po files (not sure about what is the current best practice?)

khashashin:
Thanks for opening this @Fil, i18n support would be a significant enabler for teams running Framework in multilingual environments.

I'd like to share some requirements thinking that might help guide the implementation. Drawing from experience with i18n in similar data-focused platforms, here are the areas I think are worth considering:

**Zero cost when unused.** Single-language projects (the common case) should see no performance hit, no extra build output, and no behavior change. Strictly opt-in.

**User-provided content translations.** Beyond built-in UI strings, teams need to translate their own content: page titles, sidebar labels, chart titles, descriptions, popup content, and other metadata. This is often the harder problem. Requirements to consider:
- A convention for organizing translated content (e.g. locale-keyed files, parallel directory structures, or front matter overrides).
- The ability to export/import translation bundles so translators can work outside the Framework dev environment.
- Fallback behavior: if a translation is missing for a given locale, show the default language text rather than nothing.

PR #893, opened 2024-02-22
feat: add copied notice for copy button #877

imhalid:
I made the notification for the copy button this way, if there is any problem I can fix it.

| Dark Notice | Light Notice |
|--------|--------|
| dark-notice| light-notice |


src/client/pre.js
```diff
@@ -1,7 +1,15 @@
 const copyButton = document.createElement("template");
-copyButton.innerHTML = '<button title="Copy code" class="observablehq-pre-copy"><svg width="16" height="16" viewBox="0 0 16 16" fill="none" stroke="currentColor" stroke-width="2"><path d="M2 6C2 5.44772 2.44772 5 3 5H10C10.5523 5 11 5.44772 11 6V13C11 13.5523 10.5523 14 10 14H3C2.44772 14 2 13.5523 2 13V6Z M4 2.00004L12 2.00001C13.1046 2 14 2.89544 14 4.00001V12"></path></svg></button>'; // prettier-ignore
+
+copyButton.innerHTML = `<button title="Copy code" class="observablehq-pre-copy">
+  <span class='data-copied'>Copied</span>
```
Fil, line 4:
```suggestion
  <span class="observablehq-data-copied">Copied</span>
```

use observablehq namespace to avoid conflicting with user styles.

src/client/pre.js
```diff
@@ -1,7 +1,15 @@
 const copyButton = document.createElement("template");
-copyButton.innerHTML = '<button title="Copy code" class="observablehq-pre-copy"><svg width="16" height="16" viewBox="0 0 16 16" fill="none" stroke="currentColor" stroke-width="2"><path d="M2 6C2 5.44772 2.44772 5 3 5H10C10.5523 5 11 5.44772 11 6V13C11 13.5523 10.5523 14 10 14H3C2.44772 14 2 13.5523 2 13V6Z M4 2.00004L12 2.00001C13.1046 2 14 2.89544 14 4.00001V12"></path></svg></button>'; // prettier-ignore
+
+copyButton.innerHTML = `<button title="Copy code" class="observablehq-pre-copy">
```
Fil, line 3: can we make this 1 line as it was before?

src/client/pre.js
```diff
@@ -1,7 +1,15 @@
...
+  <span class='data-copied'>Copied</span>
+  <svg width="16" height="16" viewBox="0 0 16 16" fill="none" stroke="currentColor" stroke-width="2">
+    <path d="M2 6C2 5.44772 2.44772 5 3 5H10C10.5523 5 11 5.44772 11 6V13C11 13.5523 10.5523 14 10 14H3C2.44772 14 2 13.5523 2 13V6Z M4 2.00004L12 2.00001C13.1046 2 14 2.89544 14 4.00001V12"></path>
+  </svg>
+</button>`; // prettier-ignore
+
 
 enableCopyButtons();
+const copiedNotice = document.querySelector(".data-copied");
```
Fil, line 12:
```suggestion
const copiedNotice = document.querySelector(".observablehq-data-copied");
```

src/client/pre.js
```diff
@@ -16,5 +24,25 @@ export function enableCopyButtons() {
...
+  },
+];
+
+const fadeInOutOptions = {
+  duration: 400,
+  easing: "cubic-bezier(0.25, 1, 0.5, 1)",
+  fill: "forwards",
+  direction: "alternate",
+  iterations: 2,
+};
```
Fil, line 48: nit: I'd put this inside the .animate call if they're not used anywhere else

src/style/layout.css
```diff
@@ -422,6 +422,19 @@
   align-items: center;
 }
 
+.data-copied {
```
Fil, line 425:
```suggestion
.observablehq-data-copied {
```

Fil (changes requested):
thanks for the suggestion! I have noticed a few details

imhalid:
> thanks for the suggestion! I have noticed a few details; also you can run tests locally (as you can see, [prettier is complaining](https://github.com/observablehq/framework/actions/runs/7999593408/job/21850967489)).

i fixed issues and run locally tests.

src/client/pre.js
```diff
@@ -1,13 +1,15 @@
...
 
 enableCopyButtons();
+const copiedNotice = document.querySelector(".observablehq-data-copied");
 
 export function enableCopyButtons() {
   for (const pre of document.querySelectorAll("pre:not([data-copy=none])")) {
     const parent = pre.parentNode;
-    if (parent.classList.contains("observablehq-pre-container")) continue;
     const div = parent.insertBefore(document.createElement("div"), pre);
+    if (parent.classList.contains("observablehq-pre-container")) continue;
```
Fil, line 12: why do you have to move this?

imhalid: to be honest I'm not sure if this works. but I put it back because I thought I deleted it by mistake.

should it be removed?

Fil: If you have doubts, we should definitely double check that it works, but my question is simply, why are you suggesting to switch these two lines?

imhalid: Actually I don't suggest changing it. that line was already there before. 

I thought it solved a problem that I can't see now, so I chose to add it again. I hope I understood correctly.

Fil: If you check the full diff you'll see that it is now below the line that reads
`const div = parent.insertBefore(document.createElement("div"), pre);`
when it was previously above it.

imhalid: Oh sorry, I completely misunderstood. I don't know where my mind was, now I looked at the original code and saw the mistake. :(

imhalid: i fix now

src/client/pre.js
```diff
@@ -1,7 +1,9 @@
 const copyButton = document.createElement("template");
-copyButton.innerHTML = '<button title="Copy code" class="observablehq-pre-copy"><svg width="16" height="16" viewBox="0 0 16 16" fill="none" stroke="currentColor" stroke-width="2"><path d="M2 6C2 5.44772 2.44772 5 3 5H10C10.5523 5 11 5.44772 11 6V13C11 13.5523 10.5523 14 10 14H3C2.44772 14 2 13.5523 2 13V6Z M4 2.00004L12 2.00001C13.1046 2 14 2.89544 14 4.00001V12"></path></svg></button>'; // prettier-ignore
+
+copyButton.innerHTML = '<button title="Copy code" class="observablehq-pre-copy"><span class="observablehq-data-copied">Copied</span><svg width="16" height="16" viewBox="0 0 16 16" fill="none" stroke="currentColor" stroke-width="2"><path d="M2 6C2 5.44772 2.44772 5 3 5H10C10.5523 5 11 5.44772 11 6V13C11 13.5523 10.5523 14 10 14H3C2.44772 14 2 13.5523 2 13V6Z M4 2.00004L12 2.00001C13.1046 2 14 2.89544 14 4.00001V12"></path></svg></button>'; // prettier-ignore
 
 enableCopyButtons();
+const copiedNotice = document.querySelector(".observablehq-data-copied");
```
mbostock, line 6: I expect `copiedNotice` is `null` here because `copyButton` hasn’t been inserted into the DOM yet. But I also expect you could fix this by saying:

```suggestion
const copiedNotice = copyButton.querySelector(".observablehq-data-copied");
```

imhalid: For some reason I don't understand, it gives error when I use `copyButton` instead of `document`. 

I think using `currentTarget.firstElementChild.animate` instead of `querySelector` might be a better solution.

mbostock: Oh, right, it’s a template element that we haven’t instantiated yet. You’ll need a slightly different fix in the event listener since you can’t capture a reference to the copied notice statically.

src/client/pre.js
```diff
@@ -1,5 +1,6 @@
 const copyButton = document.createElement("template");
-copyButton.innerHTML = '<button title="Copy code" class="observablehq-pre-copy"><svg width="16" height="16" viewBox="0 0 16 16" fill="none" stroke="currentColor" stroke-width="2"><path d="M2 6C2 5.44772 2.44772 5 3 5H10C10.5523 5 11 5.44772 11 6V13C11 13.5523 10.5523 14 10 14H3C2.44772 14 2 13.5523 2 13V6Z M4 2.00004L12 2.00001C13.1046 2 14 2.89544 14 4.00001V12"></path></svg></button>'; // prettier-ignore
+
+copyButton.innerHTML = '<button title="Copy code" class="observablehq-pre-copy"><span class="observablehq-data-copied">Copied</span><svg width="16" height="16" viewBox="0 0 16 16" fill="none" stroke="currentColor" stroke-width="2"><path d="M2 6C2 5.44772 2.44772 5 3 5H10C10.5523 5 11 5.44772 11 6V13C11 13.5523 10.5523 14 10 14H3C2.44772 14 2 13.5523 2 13V6Z M4 2.00004L12 2.00001C13.1046 2 14 2.89544 14 4.00001V12"></path></svg></button>'; // prettier-ignore
```
Fil, line 3:
```suggestion
copyButton.innerHTML = '<button title="Copy code" class="observablehq-pre-copy"><span class="observablehq-data-copied"></span><svg width="16" height="16" viewBox="0 0 16 16" fill="none" stroke="currentColor" stroke-width="2"><path d="M2 6C2 5.44772 2.44772 5 3 5H10C10.5523 5 11 5.44772 11 6V13C11 13.5523 10.5523 14 10 14H3C2.44772 14 2 13.5523 2 13V6Z M4 2.00004L12 2.00001C13.1046 2 14 2.89544 14 4.00001V12"></path></svg></button>'; // prettier-ignore
```

src/style/layout.css
```diff
@@ -422,6 +422,19 @@
   align-items: center;
 }
 
+.observablehq-data-copied {
```
Fil, line 425:
```suggestion
.observablehq-data-copied:before {
  content: "Copied"
}

.observablehq-data-copied {
```

Fil (changes requested):
This set the word "Copied" with css, so it does not become part of the page's textContent. So you can ctrl-A ctrl-C the whole page and not see an old "Copied" text in your clipboard. It will also make it easier  when time comes to translate in other languages (#865).

mbostock (approved):

PR #897, opened 2024-02-22
save search query, and restore on cmd-K

### LINKED ISSUE #889: remember previous search query
It would be neat if the search input remembered the previous query you've typed, and put it back (selected) when you hit cmd-K. (using sessionStorage)

Fil:
as I'm working on a project with ~300 pages, I find that not having this feature is _very_ annoying, since when I'm working on maps (say…), I have to re-type the same search query again and again (I'll test a bit with the feature on, I hope that it's not annoying in a different way :-) )

closes #889

https://github.com/observablehq/framework/assets/7001/c1e75d2e-746f-4ce8-8a78-dbd1639d802c

### DIFF
--- a/src/client/search-init.ts
+++ b/src/client/search-init.ts
@@ -23,6 +23,7 @@ addEventListener("keydown", (event) => {
     // persistently after you blur the search input.)
     toggle.classList.add("observablehq-sidebar-open");
     input.focus();
+    input.value = sessionStorage.getItem("search-query") ?? "";
     input.select();
     event.preventDefault();
   }
--- a/src/client/search.js
+++ b/src/client/search.js
@@ -28,6 +28,7 @@ const index = await fetch(import.meta.resolve("./minisearch.json"))
 input.addEventListener("input", () => {
   if (currentValue === input.value) return;
   currentValue = input.value;
+  sessionStorage.setItem("search-query", currentValue);
   if (!currentValue.length) {
     container.setAttribute("data-shortcut", shortcut);
     sidebar.classList.remove("observablehq-search-results");


Fil:
It can be live-tested on the "pangea" project.

mbostock:
What if…

- We keep a recent history of searches (captured either on clearing the search field by clicking the **x** or hitting Esc, or by navigating to a result)
- When the search box is empty, the up arrow would restore the most-recent search result and select all the text of the search query
- When the search query is selected in its entirety (and non-empty), the up and down arrow would navigate the search history, changing the query (instead of navigating within the search results)
- When the search query is selected in its entirety, hitting Enter would unselect the text and put the caret to the end of the search query, allowing you to either edit the query or hit Enter again to select the top result

I think it’s a good idea to have a way to restore previous queries. I just want it to be accessible through clicking rather than only Command-K, and I want it to be a bit more of a deliberate action than changing the normal behavior.

Fil:
The current PR is not just restoring the previous query (only one in this PR), it's also positioning the cursor to the current page (if it's one of the search results), so you can iterate through the list with one arrowDown. I'm not sure if what you suggest would be as convenient for this use case, but the only way to know is to try. (And I'm not sure if this use case is that important… it's just what I've been optimizing for as I was working on porting dozens of notebooks.)

In any case I fully agree that it should work on focus, rather than only on cmd-K.

Fil:
This PR works as before, but on focus; it only remembers/restores the previous query, but requires a small number of actions from the user. I'm exploring the alternative UI in #1044.

Fil:
rebased

src/client/search-init.ts
```diff
@@ -8,6 +8,10 @@ const input = container.querySelector<HTMLInputElement>("input")!;
 const load = () => import("observablehq:search");
 input.addEventListener("focus", load, {once: true});
 input.addEventListener("keydown", load, {once: true});
+input.addEventListener("focus", () => {
+  input.value = sessionStorage.getItem("search-query") ?? "";
+  input.select();
```
mbostock, line 13:
```suggestion
  if (input.value) return; // only restore the query if empty
  input.value = sessionStorage.getItem("search-query") ?? "";
  input.select();
```

Calling `input.select` on focus interferes with the normal behavior, _e.g._, you can’t click to place the caret or select a subset of the text because all the text is automatically selected. I think we would only want to assign the value and select it if the input is empty.

Issue #898, opened 2024-02-22
Support for Julia data loader


vnegi10:
Although the [Data loaders](https://observablehq.com/framework/loaders) page shows a small Julia example, support for Julia does not seem to be implemented. I keep getting `RuntimeError: Unable to load file` error on the dashboard.

I think updating `src/dataloader.ts` should be enough. Please let me know if the [linked PR](https://github.com/observablehq/framework/pull/899) needs any further changes.

Thank you for all the hard work and looking forward to using Julia as the back-end!

Fil:
The julia example in the documentation is supposed to show how to use _any_ language (even those we haven't included), thanks to a shebang (`#!/usr/bin/env julia`) and a `.exe` extension. But apparently it misses the point?

PR #899, opened 2024-02-22
Add support for Julia data loader

vnegi10:
Closes https://github.com/observablehq/framework/issues/898

Summary of changes:
- Added Julia extension and call
- Updated docs to reflect support for Julia

### DIFF
--- a/docs/data/forecast.json.jl
+++ b/docs/data/forecast.json.jl
@@ -1,0 +1,35 @@
+using Pkg
+Pkg.add("HTTP")
+Pkg.add("JSON")
+
+# It will be useful to create and commit a Project.toml to manage dependencies.
+# Activate environment as shown below and add packages. Once done, remove
+# any package add statements. Committing the Manifest.toml file is generally
+# not recommended.
+# Pkg.activate(@__DIR__)
+# Pkg.instantiate()
+
+using HTTP, JSON
+
+longitude = -122.47
+latitude = 37.80
+
+URL = "https://api.weather.gov/points/$(latitude),$(longitude)"
+
+function get_api_response(url::String)
+
+    request = HTTP.request("GET",
+                            url;
+                            verbose = 0,
+                            retries = 2)
+    
+    response_text = String(request.body)
+    response_dict = JSON.parse(response_text)
+    
+    return response_dict
+end
+
+response_dict = get_api_response(URL)
+hourly_dict = get_api_response(response_dict["properties"]["forecastHourly"])
+
+JSON.print(stdout, hourly_dict)
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -649,6 +649,39 @@ forecast = requests.get(station["properties"]["forecastHourly"]).json()
 json.dump(forecast, sys.stdout)
 ```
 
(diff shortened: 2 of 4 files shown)


Fil:
Thanks! I tested with `print("Hello World")`, but maybe you could share an interesting example?

(Also I think the sooner we sort the extensions by alphabetical order, the better — otherwise the precedence when two languages "compete" for the same file will just be … bizarre.)

vnegi10:
> Thanks! I tested with `print("Hello World")`, but maybe you could share an interesting example?
> 
> (Also I think the sooner we sort the extensions by alphabetical order, the better — otherwise the precedence when two languages "compete" for the same file will just be … bizarre.)

I added a forecast example analogous to what has been shown with Python and JavaScript. Do you also want me to sort the extensions alphabetically?

docs/data/forecast.json.jl
```diff
@@ -0,0 +1,35 @@
+using Pkg
+Pkg.add("HTTP")
+Pkg.add("JSON")
+
+# It will be useful to create and commit a Project.toml to manage dependencies.
+# Activate environment as shown below and add packages. Once done, remove
+# any package add statements. Committing the Manifest.toml file is generally
+# not recommended.
+# Pkg.activate(@__DIR__)
+# Pkg.instantiate()
```
Fil, line 10: I'd like to understand this comment. Is this something that needs to run once to set up the HTTP and JSON packages? If so, how would the same code run both locally and in continuous deployment?

I have the same question with R's [install_packages](https://search.r-project.org/CRAN/refmans/easypackages/html/install_packages.html) which don't seem to have an option "install only when it's not already here" ;-)

Fil:
Thank you for the example. We don't need it committed to this repo, but it's nice to have something more elaborate than `print("Hello World")`.

We'll sort the extensions in a different PR.

docs/data/forecast.json.jl
vnegi10: When Julia packages are added directly (as shown in the example using `Pkg.add`) without creating a specific environment, they become available in the global scope. This is generally not a good practice. Hence, adding a `Project.toml` is recommended (see [example](https://github.com/vnegi10/WeatherReport.jl/blob/main/Project.toml)), which contains the dependencies and optional sections to impose compatibility constraints.

The code can be run any number of times, both locally and during CI. If the packages have been downloaded before on a system, they will simply be added to the scope and not downloaded again. For instance, you will see the downloads happening only the first time locally. However, since CI runs in a new environment every time, packages will always be downloaded there.

I added this comment to make the user aware of a practice that I like to follow, which is having a `Project.toml` also along side the data loader. In that case, packages should be loaded as follows: 
```
Pkg.activate(@__DIR__)
Pkg.instantiate()
```

I didn't commit the `Project.toml` since I wasn't sure if you wanted it to be part of the repo.

vnegi10:
> Thank you for the example. We don't need it committed to this repo, but it's nice to have something more elaborate than `print("Hello World")`.
> 
> We'll sort the extensions in a different PR.

Shall I remove the example file `docs/data/forecast.json.jl`? I think it will be useful to at least leave the code in `docs/getting-started.md`?

Fil:
solved in #935

PR #901, opened 2024-02-23
Added PHP support

stefanak-michal:


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@observablehq/framework",
   "license": "ISC",
-  "version": "1.0.0",
+  "version": "1.0.1",
```
mbostock, line 4: Please revert this change. We’ll bump the version when we prepare a release. Thank you.

Fil:
wrapped under https://github.com/observablehq/framework/pull/935 ; let us know what you think of the proposal and questions it raises.

PR #905, opened 2024-02-23
autorefresh linked images and style sheets

### LINKED ISSUE #415: Linked stylesheets should auto-refresh on edit

trebor:
add logic to send a `update-stylesheet` message to the client when the stylesheet has changed.  the path & full url with new hash of the style sheet is passed to the client, which is used to update the `<link>` element.

currently, the url with the hash is sent on any file change action.  it would be nice to only send the `update-stylesheet` message only if the hash changed, and i can try to add logic to diff the hashes, but it adds complexity, so not sure if worth it.

fixes #415 and i think #908 now.

### DIFF
--- a/src/client/preview.js
+++ b/src/client/preview.js
@@ -97,8 +97,8 @@ export function open({hash, eval: compile} = {}) {
         break;
       }
       case "update-image": {
-        const image = document.querySelector(`img[src^="${message.path}"]`);
-        image.src = message.src;
+        const images = document.querySelectorAll(`img[src^="${message.path}"]`);
+        images.forEach(image => image.src = message.src);
         break;
       }
       case "add-stylesheet": {
--- a/src/client/preview.js
+++ b/src/client/preview.js
@@ -105,6 +105,11 @@ export function open({hash, eval: compile} = {}) {
         document.head.appendChild(link);
         break;
       }
+      case "update-stylesheet": {
+        const link = document.querySelector(`link[rel="stylesheet"][href^="${message.path}"]`);
+        link.href = message.href;
+        break;
+      }
       case "remove-stylesheet": {
         document.head.querySelector(`link[rel="stylesheet"][href="${message.href}"]`)?.remove();
         break;
--- a/src/client/preview.js
+++ b/src/client/preview.js
@@ -98,7 +98,7 @@ export function open({hash, eval: compile} = {}) {
       }
       case "update-image": {
         const images = document.querySelectorAll(`img[src^="${message.path}"]`);
-        images.forEach(image => image.src = message.src);
+        images.forEach((image) => (image.src = message.src));
         break;
       }
       case "add-stylesheet": {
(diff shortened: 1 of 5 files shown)


cinxmo (approved):
Nice! I originally had the idea to modify `getStylesheets` but linked stylesheets should be treated the same as any other attachment/referenced file. Thanks!

src/markdown.ts
```diff
@@ -316,7 +318,7 @@ function renderIntoPieces(renderer: Renderer, root: string, sourcePath: string):
     }
     let result = "";
     for (const piece of context.pieces) {
-      result += piece.html = normalizePieceHtml(piece.html, sourcePath, context);
+      result += piece.html = normalizePieceHtml(piece.html, root, sourcePath, context);
```
Fil, line 321: If we must introduce it, could `root` be part of the `context`?

mbostock (commented):
Can we generalize this so that it works for e.g. images too?

src/markdown.ts
trebor: done.  this cleaned up the code a bunch thank you!

trebor:
> Can we generalize this so that it works for e.g. images too? (Not necessarily blocking this PR since I know we already have some specific stuff for stylesheets. But see #908, which I thought I filed previously but I forgot!)

done.  :)

src/client/preview.js
```diff
@@ -96,6 +96,11 @@ export function open({hash, eval: compile} = {}) {
         enableCopyButtons();
         break;
       }
+      case "update-image": {
+        const image = document.querySelector(`img[src^="${message.path}"]`);
```
mbostock, line 100: This won’t work if the same image is referenced in multiple places. And also this could theory interfere with dynamically generated content by the user. But I don’t think an update-image event should be necessary in any case, since image tags are server-side rendered, and so should already be handled by diffing the generated HTML. That was my point about generalizing the updates of static assets — to some degree it should happen automatically. Stylesheets are special because they get rendered into the head and so aren’t covered by normal diffing.

mbostock (changes requested):
The latest changes don’t seem right to me, since we already diff the generated HTML and hence we shouldn’t need to create new update-image events over the socket.

src/client/preview.js
trebor: i've swapped in `querySelectorAll`, but it sounds like a different general approach is in order.

mbostock:
I’m kind of changing everything in #843 so I suggest we hold off on this for a bit.

mbostock:
This has now been subsumed by #843. I suggest we close this PR (and sorry for the conflict Trebor!).

trebor:
oh, no worries. :)

Issue #90, opened 2023-11-02
Support Windows


cinxmo:
Framework does not currently support Windows (though some users have had success with the [Windows Subsystem for Linux (WSL)](https://learn.microsoft.com/en-us/windows/wsl/install)). It should work, and we are working on it.

mbostock:
We don’t currently support Windows, but you can use the Windows Subsystem for Linux (WSL):

https://learn.microsoft.com/en-us/windows/wsl/about

GordonSmith:
> Most of Observable uses Linux or Macs, but not Windows. We should test the CLI on a Windows machine before launch

I suspect that for most of your corporate / enterprise customers (the ones who are likely to subscribe) this would not be the case (not a big deal for me personally, but with my day job hat on...)

mbostock:
@GordonSmith Agree. The “Most of Observable” was referring to Observable, Inc. employees, not users of Observable.

mbostock:
It looks like you have the `HOSTNAME` environment variable set erroneously which is overriding the default of 127.0.0.1. Perhaps we shouldn’t check this environment variable and instead require you to set it explicitly via the `--hostname` argument.

https://github.com/observablehq/framework/blob/3f09fab3ea186531eddeda670b27233af1bc09ce/bin/observable.ts#L131

onnobaudouin:
Thank you

I did try to look for where the server started to see what it did to help further but was looking in the src files .

the way i managed to get around this was in 2 ways:

### **The idiot way** 

 ```hosts
  $ sudo vi /etc/hosts
 
127.0.0.1 localhost
 # 172.0.1.1 DESKTOP-R45NOT8.        DESKTOP-R45NOT8  <-- this was generated by WSL2 
0.0.0.0   DESKTOP-R45NOT8.        DESKTOP-R45NOT8    
 ```

note that putting 127.0.0.1 didn't work either  - although it would start it would not allow the windows host to connect to.

Also as AlmaLinux on WSL2 does use systemd i could not use the normal ways to set the hostname.

### **The Build In Way** 

I then parsed you comments properly and simply changed my `package.json`

```json
...
"scripts": {
  ...
  "dev": "observable preview --host 127.0.0.1",    
  ...
}
...  
```

This one is also passed through directly whereas --host 0.0.0.0 also worked but requires knowing the full actual IP instead.

Happy to be able to start experimenting 

As for being human: the elegance of D3 tells me we are somewhat outside that limiting factor

[...]

PR #922, opened 2024-02-27
Use posix-style paths internally, only convert for IO

mythmon:
This is an alternate approach to Windows support (#90), in contrast to #881. The two PRs share a lot of the common cross platform fixes, but differe in how paths are handled.

In #881, every path was "branded" as either a `FilePath` or a `UrlPath`, and there were helpers that helped convert to, from, and between these types. In this PR, every path is kept as a "posix" style path (with forward slashes), until we have to interact with an external system like writing to a file or calling rollup. As late as possible the path is converted to use the OS path format, which is a no-op on macOS and Linux.

Comparing the two, #881 changed 87 files, with a line change of +2020/-1264. This PR touched 68 files, with a line change of +1186/-646. All other things being equal, I like smaller changes. The line count of #881 is inflated, due to it often making lines longer and causing extra wrapping.

From a subjective point of view, the two approaches feel quite different to work with. When working with the branding path types, the issue of OS vs URL paths is felt everywhere, and needs to be constantly thought about. However, it is almost always an easy thing to figure out. With this PR, the internal consistency meant that (as long as you use the right imports) the thought about different paths is almost completely invisible. However, when it *is* felt, it is a lot harder to figure out, since there aren't any guard rails from the type system.

In this PR I also had to figure out to deal with Windows' driver letters, which #881 got to pretty much ignore. The approach I went with is to turn a Windows path like `C:\Users\Amaya\` to `/C:/Users/Amaya/`. The file system handlers then have to know how to recognize that prefix and handle it appropriately. Getting that working correctly was about 1/4 of the time I spent on this project. I'm only confident I got it working due to good test coverage.

Assuming that CI passes here, my vote would be for this PR. I like the pattern of concentrating "the hard parts" into well defined interfaces instead of spreading it lighly over the whole code base. This is a leaky abstraction due to drive letters, but in the spirit of [worse is better](https://dreamsongs.com/RiseOfWorseIsBetter.html) I think that it is the right trade off.


.gitattributes
```diff
@@ -0,0 +1,16 @@
...
+*.db binary
+*.jpg binary
+*.mp4 binary
+*.parquet binary
+*.png binary
+*.tar binary
+*.tgz binary
+*.webp binary
+*.xlsx binary
+*.zip binary
```
Fil, line 16: We don't plan to run the repo under Windows, except for unit tests—and in that case I'd rather see a failure if something changes. Can we just remove the gitattributes file and live with whatever this results in?

bin/observable-init.js
```diff
@@ -1,4 +1,10 @@
-#!/usr/bin/env -S node --no-warnings=ExperimentalWarning
+#!/usr/bin/env node
+import {fileURLToPath} from "node:url";
+import crossSpawn from "cross-spawn";
 
-await import("tsx/esm");
-await import("./observable.ts");
+const observablePath = fileURLToPath(import.meta.resolve("./observable.ts"));
+crossSpawn.sync(
```
Fil, line 6: this would probably be easier after #850  (just noting for reference)

src/convert.ts
```diff
@@ -1,11 +1,11 @@
-import {existsSync} from "node:fs";
-import {utimes, writeFile} from "node:fs/promises";
-import {join} from "node:path";
 import * as clack from "@clack/prompts";
 import wrapAnsi from "wrap-ansi";
 import type {ClackEffects} from "./clack.js";
 import {CliError} from "./error.js";
 import {prepareOutput} from "./files.js";
+import {} from "./normalizedFs.js";
```
Fil, line 6:
```suggestion
```

src/pager.ts
```diff
@@ -8,6 +8,7 @@ export type PageLink =
 // Pager links in the footer are computed once for a given navigation.
 const linkCache = new WeakMap<Config["pages"], Map<string, PageLink>>();
 
+/** Strip URL query string and hash fragment. */
```
Fil, line 11:
```suggestion
// Strip URL query string and hash fragment.
```

test/dataloaders/data4.txt.exe
```diff
@@ -1,3 +1,3 @@
-#!/usr/bin/env python3
+#!/usr/bin/env python
```
Fil, line 1: this fails on my machine: `env: python: No such file or directory`

(maybe I should fix my machine, though)

Fil (commented):
Thank you so much for exploring this pathway. To me this feels much better than #881.

It leaves open the question of tracking which path is a url or a file, but this could be done as a separate PR, and maybe is just an issue of how we name variables rather than typescripting it (??).

The part that feels redundant is the special handling of \r\n. In my opinion (and it's just a weak opinion at this point), we should not change anything here in the code and the data assets; and instead, make tests smarter (for instance, we don't care if a python data loader returns "python\r\n" or "python\n" — instead let's test that the trimmed output is "python").

.gitattributes
mythmon: I didn't add this to support local development; it is required for Windows CI. I could remove the extra binary attributes and just trust Git to handle binary detection.

src/pager.ts
mythmon: These two types of comments are not equivalent. The `/**` style that I wrote causes the documentation to be available the call site, which is where I needed. Using a `//` style comment only makes this documentation visible at the implementation site, and if I'm already looking at the implementation, it is very clear what this function is doing.

test/dataloaders/data4.txt.exe
mythmon: I did this to get the test to pass in CI. I don't really want to get into the business of customizing the Window's CI images if we can avoid it.

Fil: I'm not sure this is the right call, and in any case if we change from python3 to python, we must also change the documentation that currently says:

> If you use `.py`, `.R`, `.rs`, or `.go`, the corresponding interpreter (`python3`, `Rscript`, `rust-script`, or `go run`, respectively) must be installed and available on your `$PATH`.

Fil: related: #935

mythmon: I don't think we need to adjust the docs. This isn't a python data loader, and those will continue to work as is. This is just an arbitrary exe data loader that we use in our tests, I didn't modify the Python support.

mythmon:
Note that the "Waiting for status to be reported" state of CI is because I changed the names of the expected CI jobs. We'll need to have a sort of sea-change to merge this, where we update the CI to expect the new jobs.

One thing we might consider is to introduce a "Capstone" job which itself depends on all the needed jobs. That gives us an option to change CI without having to change Github settings in the future.

test/dataloaders/data4.txt.exe
Fil: fixed by using `perl` which is available on both CI images and on macos for local devs.

mythmon:
Closing in favor of #944.

PR #924, opened 2024-02-27
Filter out workspaces for guest members

cinxmo:
Filters out workspaces for the login/deploy command where the user is a guest member and does not have owner/editor access on at least one project on that workspace

### DIFF
--- a/src/deploy.ts
+++ b/src/deploy.ts
@@ -8,7 +8,7 @@ import {commandRequiresAuthenticationMessage} from "./commandInstruction.js";
 import type {Config} from "./config.js";
 import {CliError, isApiError, isHttpError} from "./error.js";
 import type {Logger, Writer} from "./logger.js";
-import {type AuthEffects, defaultEffects as defaultAuthEffects, formatUser, loginInner} from "./observableApiAuth.js";
+import {type AuthEffects, defaultEffects as defaultAuthEffects, formatUser, loginInner, validWorkspaces} from "./observableApiAuth.js";
 import {ObservableApiClient} from "./observableApiClient.js";
 import {
   type GetCurrentUserResponse,
@@ -104,12 +104,9 @@ export async function deploy(
   let authError: null | "unauthenticated" | "forbidden" = null;
   try {
     if (apiKey) {
-      const invalidTiers = new Set(["basic", "enterprise", "public", "pro", "pro_enterprise"]);
       currentUser = await apiClient.getCurrentUser();
       // List of valid workspaces that can be used to create projects.
-      currentUser.workspaces = currentUser.workspaces.filter(
-        (w) => (w.role === "owner" || w.role === "member") && !invalidTiers.has(w.tier)
-      );
+      currentUser.workspaces = validWorkspaces(currentUser.workspaces);
     }
   } catch (error) {
     if (isHttpError(error)) {
--- a/src/observableApiAuth.ts
+++ b/src/observableApiAuth.ts
@@ -18,6 +18,8 @@ import {bold, defaultEffects as defaultTtyEffects, inverse, link, yellow} from "
 
 const OBSERVABLE_UI_ORIGIN = getObservableUiOrigin();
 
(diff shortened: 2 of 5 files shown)


src/deploy.ts
```diff
@@ -104,12 +110,9 @@ export async function deploy(
   let authError: null | "unauthenticated" | "forbidden" = null;
   try {
     if (apiKey) {
-      const invalidTiers = new Set(["basic", "enterprise", "public", "pro", "pro_enterprise"]);
       currentUser = await apiClient.getCurrentUser();
       // List of valid workspaces that can be used to create projects.
-      currentUser.workspaces = currentUser.workspaces.filter(
-        (w) => (w.role === "owner" || w.role === "member") && !invalidTiers.has(w.tier)
-      );
+      currentUser.workspaces = validWorkspaces(currentUser.workspaces);
```
Fil, line 115: can we write this without mutation?

maybe:

```
({workspaces, ...currentUser} = await apiClient.getCurrentUser());
currentUser.workspaces = validWorkspaces(workspaces);
```

or

```suggestion
      currentUser = {...currentUser, workspaces: validWorkspaces(currentUser.workspaces)};
```

or make `validWorkspaces` consume and return a (different) user.

src/observableApiAuth.ts
```diff
@@ -108,6 +110,7 @@ export async function loginInner(effects: AuthEffects): Promise<{currentUser: Ge
   apiClient.setApiKey({source: "login", key: apiKey.key});
   const currentUser = await apiClient.getCurrentUser();
   spinner.stop(`You are logged into ${OBSERVABLE_UI_ORIGIN.hostname} as ${formatUser(currentUser)}.`);
+  currentUser.workspaces = validWorkspaces(currentUser.workspaces);
```
Fil, line 113: same remark here

Fil (changes requested):

src/observableApiAuth.ts
```diff
@@ -149,3 +152,19 @@ export async function whoami(effects = defaultEffects) {
...
+  return workspaces.filter((w) => {
+    if (
+      VALID_TIERS.has(w.tier) &&
+      (w.role === "owner" ||
+        w.role === "member" ||
+        (w.role === "guest_member" &&
+          w.projects_info.some((info) => info.project_role === "owner" || info.project_role === "editor")))
+    )
+      return true;
+    return false;
```
mythmon, line 168:
```suggestion
    return (
      VALID_TIERS.has(w.tier) &&
      (w.role === "owner" ||
        w.role === "member" ||
        (w.role === "guest_member" &&
          w.projects_info.some((info) => info.project_role === "owner" || info.project_role === "editor")))
    )
```

Since this just returns true/false, we can simplify it. (Prettier might not like this as formatted here, sorry)

test/mocks/observableApi.ts
```diff
@@ -342,25 +346,46 @@ const workspaces = [
...
+    id: "0000000000000005",
+    login: "mock-user-ws-5",
+    name: "Mock User's Fifth Workspace Guest Member as Viewer",
+    tier: "pro_2024",
+    type: "team",
+    role: "guest_member",
+    projects_info: [{project_slug: "test-project-2", project_role: "viewer"}]
   }
-];
+] as GetCurrentUserResponse["workspaces"];
```
mythmon, line 388: `as` will mask any errors here, which can cause problems down the line. If you are trying to mask an error, please make this actually match the type so the error goes away. If you are trying to make sure that TS sees this as the right kind of value, some approaches that are helpful are:

1. type the value: `const workspaces: GetcurrentUserResponse["workspaces"] = [...]`
2. replace the `as` with `satisfies` which does a safe type check without masking errors.
3. use `as const` which often helps TS realize the right type.
3.

mythmon (changes requested):

mythmon (approved):

Fil (approved):

PR #926, opened 2024-02-27
absolute FileAttachment.url

### LINKED ISSUE #925: FileAttachment.url should return an absolute URL, not a relative path
Ref. https://uwdata.github.io/mosaic-framework-example/mosaic-duckdb-wasm

mbostock:
Fixes #925.


mbostock:
Ooh, good catch, it in fact says “resolved **relative path**” which is now wrong. I’ll fix.

PR #927, opened 2024-02-27
make login test faster by not waiting 1s to poll

mythmon:
We might also be able to do this by having timeouts be an effect instead of using a short timeout, but this seems like an easier way to deal with it.


mbostock:
Thank you for fixing this. 🙏

mcglincy (approved):
LGTM

PR #935, opened 2024-02-28
interpreters

### LINKED ISSUE #831: add java support to dataloaders

### LINKED ISSUE #898: Support for Julia data loader

Fil:
This PR introduces a new **interpreters** configuration option, which is a key-value object where keys are extensions (denoted with a dot, e.g. ".pl"), and values are arrays of commands and arguments (e.g. ["perl"]).

This option is _merged_ with the default interpreters, which allows us to tweak the supported interpreters' options. One can use null(ish) to disable an interpreter. As such, it does not allow reordering the interpreter priority in case of conflict for the same file (but I don't think it would be a great idea for a user to have two competing data loaders for a file anyway).

~~A less incremental approach would be to replace the default interpreters. Maybe it's better? The cost is that if you just need to add support for (say) perl, but don't want to lose support for the others, you need to copy the defaults from the documentation. But the upside is simplicity, ease of understanding.~~

~~Since it's much simpler to add an interpreter now, we could withdraw support for py, R, julia, ts and only support js data loaders "out of the box". It would be a breaking change, but it would allow users to experiment more freely?~~

I considered calling this "languages", but it risks conflicting with internationalization #865.

I made a fun data loader in osascript, straightforward port of https://stackoverflow.com/questions/71104893/macos-how-to-access-the-live-text-ocr-functionality-from-applescript-jxa, which reads text from a screenshot png. (We wouldn't want to support osascript by default since it's (heavily) platform-dependent.)

closes #831
supersedes #846
closes #898
supersedes #899
supersedes #901
supersedes #969 (maybe)

TODO:
- [x] pass interpreters to getFileHash()


mbostock:
> Since it's much simpler to add an interpreter now, we could withdraw support for py, R, julia, ts and only support js data loaders "out of the box". It would be a breaking change, but it would allow users to experiment more freely?

We should include more “in the box”, not less (and you can always change what’s in the box with the config).

Fil:
If we land this we'll want to open another issue requesting examples of data loaders in java, julia, php. The original PRs have good starting points.

docs/config.md
```diff
@@ -195,3 +195,35 @@ toc: false
 ## search
 
 Whether to enable [search](./search) on the project; defaults to false.
+
+## interpreters
+
+The interpreters option specifies additional languages for data loaders, indicating the extension and associated interpreter (see [routing](./loaders#routing) for details). The list of supported interpreters defaults to:
```
mbostock, line 201:
```suggestion
The **interpreters** option specifies additional interpreted languages for data loaders, indicating the file extension and associated interpreter (see [loader routing](./loaders#routing) for more). The default list of interpreters is:
```

docs/config.md
```diff
@@ -195,3 +195,35 @@ toc: false
...
+
+```js run=false
+{
+  ".js": ["node", "--no-warnings=ExperimentalWarning"],
+  ".ts": ["tsx"],
+  ".sh": ["sh"],
+  ".exe": [""],
```
mbostock, line 208: Shouldn’t this be the empty array?

```suggestion
  ".exe": [],
```

Fil: good catch! that's what is in the code, in fact.

docs/config.md
```diff
@@ -195,6 +195,38 @@ toc: false
 ## search
 
 Whether to enable [search](./search) on the project; defaults to false.
+code
```
Fil, line 198:
```suggestion

```

Fil:
I think this plumbing has a benefit in that is shows the dependency of certain files on data loaders, while others do not depend on them.

In particular you can see that we were previously looking for data loader extensions for missing files in getFileHash, even if these files were styles. cf. https://github.com/observablehq/framework/pull/935/files#diff-7287acfe1f0dd19ab919708ad1be7cc05d3e9878e5c6e9420ed7b57110f233f4 line 139.

docs/loaders.md
```diff
@@ -126,42 +126,38 @@ Like with any other file, these files from generated archives are live in previe
...
 - `.ts` - TypeScript (`tsx`)
+- `.sh` - shell script (`sh`)
+- `.exe` - arbitrary executable
+- `.go` - Go (`go run`)
+- `.java` — Java (`java`; requires Java 11+ and [single-file programs](https://openjdk.org/jeps/330))
+- `.jl` - Julia (`julia`)
+- `.php` - PHP (`php`)
 - `.py` - Python (`python3`)
 - `.R` - R (`Rscript`)
 - `.rs` - Rust (`rust-script`)
```
mbostock, line 141: Changing the order of precedence would be non-backwards-compatible, and I don’t want to call this Framework 2.0. Can we restore the original order?

src/dataloader.ts
```diff
@@ -79,8 +68,13 @@ export abstract class Loader {
    * abort if we find a matching folder or reach the source root; for example,
    * if docs/data exists, we won’t look for a docs/data.zip.
    */
-  static find(sourceRoot: string, targetPath: string, {useStale = false} = {}): Loader | undefined {
-    const exact = this.findExact(sourceRoot, targetPath, {useStale});
+  static find(
+    sourceRoot: string,
+    targetPath: string,
+    interpreters: Config["interpreters"],
+    {useStale = false} = {}
```
mbostock, line 75: Why is `useStale` passed as a named option, but `interpreters` is a positional argument? Because it’s required, whereas `useStale` has a default? I feel like we need a better convention here, as we have a number of internal functions that have lots of arguments. I might experiment with some alternatives. Maybe it shouldn’t be a blocker, but I’m still hung up on the plumbing.

test/config-test.ts
```diff
@@ -2,6 +2,20 @@ import assert from "node:assert";
 import {normalizeConfig as config, mergeToc, readConfig, setCurrentDate} from "../src/config.js";
 
 const root = "test/input/build/config";
+const interpreters = new Map([
```
mbostock, line 5:
```suggestion

const interpreters = new Map([
```

docs/loaders.md
Fil: I wanted to reorder alphabetically as this list is going to build up and then look messy. The order of precedence is not something we should design for: it would not be a great idea for a user to have two competing data loaders for a file, and to rely on this order to pick up the chosen data loaders. (It's also not something that the user can change with the **interpreters** option.)

You're right however that this is what the code currently does, and that changing the order will break someone's code. I'm thinking now that this should be addressed by erroring — instead of picking the first match from an arbitrary list. But if we don't error, then yes we should keep a backwards compatible order.

src/dataloader.ts
Fil: I don't any argument in favor of this style. Passing it as {useStale, interpreters} would be fine too.

docs/loaders.md
Fil: Also consider the following scenario:

A user wants to write a script in a language that is not supported (say, fortran). They write their data loader as `data.csv.f77`, then realize it doesn't work, and write `data.csv.sh` which sets up a few variables and execs `data.csv.f77`.

A week later a new version of Framework is out and supports the `f77` extension.

test/fileWatchers-test.ts
```diff
@@ -3,10 +3,11 @@ import {mkdirSync, renameSync, unlinkSync, utimesSync, writeFileSync} from "node
 import {basename, dirname, extname, join} from "node:path/posix";
 import {InternSet, difference} from "d3-array";
 import {temporaryDirectoryTask} from "tempy";
+import {normalizeConfig} from "../src/config.js";
 import {FileWatchers} from "../src/fileWatchers.js";
 import {resolvePath} from "../src/path.js";
 
-describe("FileWatchers.of(root, path, names, callback)", () => {
+describe("FileWatchers.of(root, path, names, callback)", async () => {
```
mbostock, line 10: You can’t have an `async` callback to `describe`; any tests after an `await` would be ignored.

src/config.ts
```diff
@@ -136,6 +153,7 @@ export async function normalizeConfig(spec: any = {}, defaultRoot = "docs"): Pro
   toc = normalizeToc(toc);
   deploy = deploy ? {workspace: String(deploy.workspace).replace(/^@+/, ""), project: String(deploy.project)} : null;
   search = Boolean(search);
+  interpreters = new Map(Object.entries({...defaultInterpreters, ...interpreters}).filter(([, i]) => i != null));
```
mbostock, line 156: I don’t like that the user-facing interface is an object, but it’s converted to a Map here. Let’s just leave it as an object.

test/dataloaders-test.ts
```diff
@@ -1,46 +1,52 @@
...
+import {normalizeConfig} from "../src/config.js";
 import {type LoadEffects, Loader} from "../src/dataloader.js";
 
 const noopEffects: LoadEffects = {
   logger: {log() {}, warn() {}, error() {}},
   output: {write() {}}
 };
 
-describe("Loader.find(root, path)", () => {
+describe("Loader.find(root, path)", async () => {
```
mbostock: This is really evil! If you have an `async` callback to `describe`, it silently skips the tests! As a result we are currently running only 615 of 650 tests in main. 😬

mbostock (approved):

PR #938, opened 2024-02-28
document scripts; use head

mbostock:
Followup to #921. Using **head** is simpler (and we already do that for Google Analytics). Adds documentation for the **scripts** option, too.


visnup (approved):
Do we care about all the examples too?

mbostock:
Ah yes, and we should have Google Analytics there too for consistency.

visnup (approved):

PR #941, opened 2024-03-01
adopt turbo

Fil:
With turbo, the unit tests are running faster (mostly because they get to be parallelized, I guess?). It should be possible to split the test suite into 5 or 10 buckets and get more speed.

ref. https://turbo.build/repo/docs/reference/configuration

It also seems to fix the pesky terminal bug!

todo:
- [ ] split tests into 4 or 8 buckets?
- [x] fix coverage (reported as 0%)


Fil:
using concurrently instead

superseded by #945

PR #943, opened 2024-03-01
add leaflet images

mbostock:
Some of these would be statically discoverable by parsing the CSS, but some of them are only referenced in the code (and without using `import.meta.resolve`), so we’ll need to enumerate these anyway if we want it to work out of the box.

### DIFF
--- a/src/npm.ts
+++ b/src/npm.ts
@@ -194,6 +194,7 @@ export async function resolveNpmImports(root: string, path: string): Promise<Imp
   promise = (async function () {
     try {
       const filePath = await populateNpmCache(root, path);
+      if (!/\.(m|c)?js$/i.test(path)) return []; // not JavaScript; TODO traverse CSS, too
       const source = await readFile(filePath, "utf-8");
       const body = Parser.parse(source, parseOptions);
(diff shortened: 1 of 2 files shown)


src/npm.ts
```diff
@@ -194,6 +194,7 @@ export async function resolveNpmImports(root: string, path: string): Promise<Imp
   promise = (async function () {
     try {
       const filePath = await populateNpmCache(root, path);
+      if (!/\.(m|c)?js$/i.test(path)) return []; // not JavaScript; TODO traverse CSS, too
```
mbostock, line 197: This avoids a spurious “unable to fetch or parse” when `import.meta.resolve` is passed a non-JavaScript file. I’m a little loathe to enumerate the allowed file extensions for JavaScript, but I think this is basically fine and >99% of the time it should be `.js` anyway, and plus we need a supported file extension to serve the MIME type besides. And it’s not like we have a bundler that will support TypeScript or JSX (yet).

Fil:
Ouch… Is this something that users might face when requesting other libraries? (Or using additional leaflet plugins?).

Fil (approved):

PR #944, opened 2024-03-02
windows support

### LINKED ISSUE #90: Support Windows

mythmon:
Fixes #90.

- **disable autocrlf**
- **test on windows**
- **cross-spawn**
- **rimraf**
- **use cross-env to run mocha**
- **fix tests with posix paths**
- **one less op**
- **mostly posix**
- **fix data loader and config tests**
- **more tests**
- **all tests passing**


mbostock (approved):

mythmon:
I'm enabling auto merge on this; but it won't merge until our CI configuration change to look for the newly added CI jobs instead of the old one.

.github/workflows/test.yml
```diff
@@ -11,7 +11,9 @@ jobs:
     strategy:
       matrix:
         version: [20, 21]
-    runs-on: ubuntu-latest
+        os: [ubuntu-latest, windows-latest]
+      fail-fast: false
```
mbostock, line 15: What does this do?

src/files.ts
```diff
@@ -1,17 +1,27 @@
...
+}
+
+export function fromOsPath(path: string): string {
+  return path.split(op.sep).join(sep);
+}
+
 /**
  * Returns the relative path from the current working directory to the given
- * Framework source file, such as "./src/client/search.js". This is typically
+ * Framework source file, such as "./src/client/search.js". This is typigit pucally
```
mbostock, line 20:
```suggestion
 * Framework source file, such as "./src/client/search.js". This is typically
```

.github/workflows/test.yml
mythmon: `fail-fast: true` (the default) means that if any CI job fails, the rest are immediately cancelled. `fail-fast: false` lets them all run to completion. `true` saves on CI time, `false` leads to easier debugging.

mythmon:
We need to add rimraf to the template package.json's

PR #945, opened 2024-03-02
run tests concurrently

mbostock:
Seems simpler than turbo #941 and just as fast. Runs the tests in ~4.5s for me, vs. 10s on main.


Fil:
On my machine I'm down to 12s instead of 16s.

A major bummer though: I see in [the docs](https://developer.ibm.com/articles/parallel-tests-mocha-v8/) that “[Exclusive tests](https://mochajs.org/#exclusive-tests) (.only()) do not work.” I've tried and it's still true with the current version.

```
[test:mocha]   1) Uncaught error outside test suite:
[test:mocha]      Uncaught Error: `.only` is not supported in parallel mode
```

mbostock:
It’s true, but using `it.only` is relatively rare, and we could have a separate target for that so that the default of running everything is still fast.

Fil (approved):

PR #952, opened 2024-03-02
tempy for FileWatchers tests

mbostock:
These tests are flakey when run in parallel because they interact with each other and with other tests. To isolate them, we use `tempy` to create temporary directories that are independent and automatically cleaned up when the tests complete.


package.json
```diff
@@ -29,6 +29,7 @@
     "test": "concurrently npm:test:mocha npm:test:tsc npm:test:lint npm:test:prettier",
     "test:coverage": "c8 yarn test:mocha",
     "test:mocha": "rimraf --glob test/.observablehq/cache test/input/build/*/.observablehq/cache && cross-env OBSERVABLE_TELEMETRY_DISABLE=1 TZ=America/Los_Angeles tsx --no-warnings=ExperimentalWarning ./node_modules/mocha/bin/mocha.js -p 'test/**/*-test.*'",
+    "test:mocha:serial": "rimraf --glob test/.observablehq/cache test/input/build/*/.observablehq/cache && cross-env OBSERVABLE_TELEMETRY_DISABLE=1 TZ=America/Los_Angeles tsx --no-warnings=ExperimentalWarning ./node_modules/mocha/bin/mocha.js 'test/**/*-test.*'",
```
mbostock, line 32: This is for `.only` as suggested in https://github.com/observablehq/framework/pull/945#issuecomment-1974738677.

Fil (approved):

PR #957, opened 2024-03-02
jsr: imports

### LINKED ISSUE #956: JSR imports
E.g.,

```js
import {printProgress} from "jsr:@luca/flag";
```

But it’s not clear to me that the JSR API supports conversion to vanilla ES modules? Like, I get that I can fetch https://jsr.io/@luca/flag/1.0.0/main.ts, but how would I get the default entry into a module? And how would I bundle and convert TypeScript to vanilla JavaScript? 🤔 

Ref. https://jsr.io/docs/native-imports
Ref. https://jsr.io/docs/api
Ref. https://jsr.io/docs/native-imports#implementing-native-jsr-imports-in-tools (not yet written)

mbostock:
Ref. https://jsr.io/. Fixes #956. TODO

- [x] Resolve `jsr:` imports
- [x] Populate the `_jsr` cache during resolution
- [x] Resolve and rewrite transitive `jsr:` imports (_e.g._, `@jsr/quentinadam__assert` for `jsr:@quentinadam/hex`)
- [x] Resolve and rewrite transitive `npm:` imports (e.g., `jsr:@oak/oak` depends on `npm:path-to-regexp`)
- [x] Preload transitive imports
- [x] Respect the dependency version range specified in package.json
- [x] Adopt [resolve.exports](https://github.com/lukeed/resolve.exports) for more correct conditional export resolution
- [x] Documentation
- [x] Tests


mbostock:
It’s alive!

Screenshot 2024-03-02 at 10 50 56 AM

Fil:
I tried a few of the modules listed, to find some good stuff and make tests.

[@libs/markdown](https://jsr.io/@libs/markdown): `import * as mod from "jsr:@libs/markdown";` crashed with `Error: Missing "./slugify" specifier in "@jsr/std__text" package`. Did not find how to fix this one.

[@ch/tomd](https://jsr.io/@ch/tomd): `import { tomd } from "jsr:@ch/tomd"` displays `SyntaxError: The requested module '/_npm/happy-dom@15.7.4/_esm.js' does not provide an export named 'Browser'` 

I managed to crash the preview server when I tried to import [k-means-pp](https://jsr.io/@ppz/k-means-pp) from `jsr:k-means-pp` (the correct specifier was `jsr:@ppz/k-means-pp`), my mistake.

(Maybe these errors can be controlled?)

Most modules I've tried worked out the the box. There is good stuff: [@ppz/k-means-pp](https://jsr.io/@ppz/k-means-pp) — [@storynode/jql](https://jsr.io/@storynode/jql) — [@dbushell/hyperless](https://jsr.io/@dbushell/hyperless) — [@notwoods/spherical-geometry-js](https://jsr.io/@notwoods/spherical-geometry-js)…


I'm watching https://www.youtube.com/watch?v=MFCn4ce5dVc for the rationale.

mbostock:
Not sure if we will be able to fix any of those errors, but I can look. Do you think they are blocking?

Fil:
No, not blocking. I'm now trying to publish a package to see what happens "end-to-end".

Fil (approved):

mbostock:
The `slugify` error seems like a bug in @libs/markdown. In @std/text 1.0.5, `slugify` [was available](https://jsr.io/@std/text/1.0.5/slugify.ts) but documented as unstable; in @std/text 1.0.6 it [was moved](https://jsr.io/@std/text/1.0.6/unstable_slugify.ts) to `unstable_slugify`. But @libs/markdown says it just depends on @std/text 1, but it needs exactly 1.0.3, 1.04, or 1.0.5 to work. It has 1.0.4 in [its deno.lock file](https://github.com/lowlighter/libs/blob/53807ffba381d5e2010effa929335a308790adea/markdown/deno.lock#L17) but I don’t think we have to (or should) respect lock files when resolving transitive dependencies.

mbostock:
The @ch/tomd error is because we’re trying to load the happy-dom package from jsDelivr, but jsDelivr refuses to serve it and replaces it with an error saying:

```
Failed to bundle using Rollup v2.79.1: the file imports a not supported node.js built-in module "node:perf_hooks". If you believe this to be an issue with jsDelivr, and not with the package itself, please open an issue at https://github.com/jsdelivr/jsdelivr'
```

I’m not entirely sure why jsDelivr’s error isn’t shown; but at any rate the error saying that it doesn’t export a symbol called `Browser` is accurate, because jsDelivr wasn’t able to bundle the code. I don’t think we need to do anything for this one either; this is just a package that doesn’t work in a browser environment. (Though perhaps it would be nice to detect when jsDelivr fails to bundle the code, rather than downloading a broken library… I can file a bug for that.)

mbostock:
I fixed the `jsr:k-means-pp` error, and also am now populating the in-memory JSR version resolution cache correctly. There are probably some more bugs to fix here. We need more tests.

PR #959, opened 2024-03-03
remove cautionary note about windows

Fil:
reverts #835


mbostock:
Looks good, but I think we should put this into a release branch for 1.1 so that it goes live the same time 1.1 goes live.

PR #963, opened 2024-03-03
build with esbuild

### LINKED ISSUE #850: Compile to JavaScript and drop runtime dependency on tsx

mbostock:
Fixes #850. Lots to do…

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 .env
 .DS_Store
 **/.observablehq/cache/
+build/
 coverage/
 docs/themes.md
 docs/theme/*.md
--- a/package.json
+++ b/package.json
@@ -11,11 +11,8 @@
     "url": "https://github.com/observablehq/framework.git"
   },
   "files": [
-    "bin/**/*.js",
-    "bin/**/*.ts",
-    "src/**/*.js",
-    "src/**/*.ts",
-    "src/**/*.css",
+    "build/**/*.js",
+    "build/**/*.css",
     "templates"
   ],
   "bin": {
@@ -23,6 +20,7 @@
   },
   "scripts": {
     "dev": "rimraf --glob docs/themes.md docs/theme/*.md && (tsx watch docs/theme/generate-themes.ts & tsx watch --no-warnings=ExperimentalWarning ./bin/observable.ts preview --no-open)",
+    "build:esbuild": "rimraf build && yarn esbuild --outdir=build --platform=node $(find bin src \\( -name '*.ts' -o -name '*.js' -o -name '*.css' \\) ! -name '*.d.ts' )",
     "build": "yarn rebuild-themes && rimraf dist && tsx --no-warnings=ExperimentalWarning ./bin/observable.ts build",
     "deploy": "yarn rebuild-themes && tsx --no-warnings=ExperimentalWarning ./bin/observable.ts deploy",
     "rebuild-themes": "rimraf --glob docs/themes.md docs/theme/*.md && tsx docs/theme/generate-themes.ts",
@@ -75,7 +73,6 @@
     "semver": "^7.5.4",
     "send": "^0.18.0",
     "tar-stream": "^3.1.6",
-    "tsx": "~4.2.1",
     "untildify": "^5.0.0",
     "wrap-ansi": "^9.0.0",
     "ws": "^8.14.2"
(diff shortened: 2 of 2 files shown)


mbostock:
Some additional learnings…

- We had a number of imports of .json files without import attributes (or assertions). These only worked because tsx is loose, so we have to rewrite them to be explicit to make them standard JavaScript (since esbuild will not rewrite these imports).
- Import attributes are only supported on more recent versions of tsx, so we have to upgrade tsx if we want to continue using it for tests.
- However, more recent versions of tsx break code coverage with c8 (https://github.com/privatenumber/tsx/issues/433), so this means we don’t have any code coverage… unless we can figure out how to run the tests using esbuild instead of tsx
- It also means we need to upgrade Prettier. But I hate the [new ternary formatting](https://prettier.io/blog/2023/11/13/3.1.0) introduced in Prettier 3.1 (and more generally I hate the idea that our formatting styles are subject to the whims of the Prettier maintainers, and apparently they can evolve over time, causing churn in our codebase). So either we need to avoid import attributions so we can stick with Prettier <3.1, or plug our noises and upgrade to Prettier 3.1+, or drop Prettier entirely (perhaps in favor of some alternative but I’m not aware of one). I don’t like any of these options. (I’ll also note that there are 100+ `// prettier-ignore` directives in our repo, but most of them are for long lines in tests.)
- If we drop tsx, then we won’t be able to support TypeScript project configuration files `observablehq.config.ts`, at least via direct import… so maybe we’ll need to use esbuild to load it on the fly? I guess we can do the latter and it shouldn’t be a big deal because we already have a runtime dependency on esbuild

tsconfig.json
```diff
@@ -2,7 +2,7 @@
   "compilerOptions": {
     "noImplicitAny": false,
     "lib": ["ES2023", "DOM.Iterable"],
-    "module": "Node16",
+    "module": "NodeNext",
```
mbostock, line 5: This is now needed for the import attributes.

src/bin/observable.ts
```diff
@@ -217,7 +217,7 @@ try {
         );
       }
     } else {
-      console.error(`\n${red("Unexpected error:")} ${error.message}`);
+      console.error(`\n${red("Unexpected error:")} ${(error as any).message}`);
```
mbostock, line 220: This is needed because I upgraded TypeScript (which was needed because of import attributes).

mbostock:
Oh also, if we use import attributes, then we require Node 20.10 — I think that’s probably too bleeding edge (2023-11-22)? If we use the now-deprecated import assertions (and drop tsx at runtime), we can support Node 18. Maybe we can drop that in favor of `process.env.npm_package_version`? Not sure if that works in Yarn though… and also it’ll be wrong when Framework is run as a dependency, surely.

.gitignore
```diff
@@ -1,11 +1,13 @@
...
-.DS_Store
 **/.observablehq/cache/
-coverage/
-docs/themes.md
-docs/theme/*.md
-dist/
+.DS_Store
+.env
+/build.test/
+/build/
```
mbostock, line 5: The leading slash here means only ignore `build/` at the project root, not `test/input/build/` etc.

mbostock:
I think the remaining issue is that I don’t know how to do the equivalent of `$(find …)` on Windows. We could use [glob](https://www.npmjs.com/package/glob), but I still don’t know how to pass the result of that to the esbuild command.

mbostock:
I think I’ll have to write a script in JavaScript and use esbuild’s JavaScript API so that it works on Windows. 😞

Fil:
I've tested it as much as I could (and could imagine how to).

One thing that I found a bit confusing is that `yarn build` in this repo builds the documentation site (`dist`), as an alias of `observable build`. You need `yarn prepublishOnly` to build the `build` folder. Since this is just for us developers of Framework, it's probably OK, but we could also change this either to make it clearer by changing the commands names — or more simply by building both `dist` and `build` (it's 0.7s on my machine, so fast that it's not going to be a burden).

mbostock:
Can you approve if it looks good?

I thought about changing `yarn dev` to `yarn docs:dev` etc. but I don’t see a big benefit and it’ll be more disruptive. And I don’t think it makes sense to combine the two unrelated actions into one.

Fil (approved):

mbostock:
Like really the current `dist` should be `dist-docs` or something, and the current `build` should be `dist`… but they’re all just names internal to this repo so I don’t think it matters.

PR #965, opened 2024-03-04
markdown-it plugins

### LINKED ISSUE #940: Support footnotes in markdown text
Notebooks often need to omit details in the main text, but these details still need to be there somewhere. Footnotes are a perfect way to achieve this.

[markdown-it](https://github.com/markdown-it/markdown-it) does support footnotes. What's stopping us from having them in our Observable Framework notebooks? :)

Bonus point would be to have the possibility to have the footnotes on the right side of the notebook, or have them appear on mouseOver.

mbostock:
Fixes #940.

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -195,3 +195,16 @@ toc: false
 ## search
 
 Whether to enable [search](./search) on the project; defaults to false.
+
+## markdownIt
+
+A hook for registering additional [markdown-it](https://github.com/markdown-it/markdown-it) plugins. For example, to use [markdown-it-footnote](https://github.com/markdown-it/markdown-it-footnote):
+
+```js
+import type MarkdownIt from "markdown-it";
+import MarkdownItFootnote from "markdown-it-footnote";
+
+export default {
+  markdownIt: (md: MarkdownIt) => md.use(MarkdownItFootnote);
+};
+```
--- a/src/config.ts
+++ b/src/config.ts
@@ -2,6 +2,7 @@ import op from "node:path";
 import {basename, dirname, join} from "node:path/posix";
 import {cwd} from "node:process";
 import {pathToFileURL} from "node:url";
+import type MarkdownIt from "markdown-it";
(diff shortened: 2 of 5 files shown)


Fil:
Super nice to support markdown-it plugins. But out of the box footnotes look bad.

notes

I would say that this is the DIY version of #940. True support for footnotes means we'd have nice styles too, and for now this means the user has to add their own custom style.css (and cannot use themes). In other words it's a necessary step, but I don't think it's enough to really close the OP.

This also bears the question of which plugins we want to include by default. We can start with none, of course.

Besides footnotes, I think [markdown-it-container](https://github.com/markdown-it/markdown-it-container) would make total sense since it would allow us to use markdown in conjunction with our component classes (::: card, ::: warning, etc).

For example if we configure:

```ts
import MarkdownItContainer from "markdown-it-container";
import MarkdownItFootnote from "markdown-it-footnote";
```

```ts
  markdownIt: (md: MarkdownIt) =>
    md
      .use(MarkdownItFootnote)
      .use(MarkdownItContainer, "card")
      .use(MarkdownItContainer, "tip")
      .use(MarkdownItContainer, "warning")
```

we can get this:

````md

::: card
That’s a _nice_ card, ain’t it? Here is a footnote reference,[^1] and another.[^longnote]
:::

::: tip
That’s a _nice_ **tip**, ain’t it?
:::

::: warning
That’s a _nice_ **warning**, ain’t it?
:::

[^1]: Here is the footnote.
[^longnote]: Here's one with multiple blocks.

    Subsequent paragraphs are indented to show that they
    belong to the previous footnote.
`````

to render like so:
Capture d’écran 2024-03-04 à 10 06 04


I would certainly use those.

mbostock:
I think we should add this now and see how people use it before adding anymore preemptively (or speculatively). Sound/look good?

Fil (approved):
Sure. I'm approving this, but think we should keep #940 open.

PR #970, opened 2024-03-04
separate docs:build

mbostock:
Per https://github.com/observablehq/framework/pull/963#issuecomment-1976110292, and to fix the examples build, which needs both `yarn docs:build` to build the docs, and `yarn build` so that the examples can run `build/bin/observable.js`.


mbostock:
(Maybe we should have generated to `docs/.observablehq/dist` by default like VitePress does but changing that would be disruptive now.)

Fil (approved):
thanks!

PR #980, opened 2024-03-05
fix transitive static imports

mbostock:
There was a bug where a transitive static import from a dynamic import was considered a static import, but it should be considered a dynamic import. Adds more tests for `getResolvers` to increase coverage.


Fil (approved):
(note to myself)

the tests that would fail without this patch are
* detects local transitive dynamic imports (3)
* detects local transitive dynamic imports (4)

PR #986, opened 2024-03-05
Docs: Added quotes in cli for Windows PowerShell compatibility

danmarshall:
Thanks for adding Windows support! 🎉🍻

In Windows Powershell (the default terminal for VSCode), the `@` character is reserved, and the current command will only run `npm init`.


Fil:
wouldn't `yarn create "@observablehq"` work everywhere? (I'd prefer a single generic line than listing cases.)

danmarshall:
@Fil fixed.

mbostock (approved):
I’d like to find a way to hide these quotes via CSS on non-Windows (purely for aesthetics), but this is a fine place to start.

PR #987, opened 2024-03-06
unpin echarts

mbostock:
Seems like the latest version works, as long as you use the correct entry (`dist/echarts.esm.min.js`).

### DIFF
--- a/docs/lib/echarts.md
+++ b/docs/lib/echarts.md
@@ -4,7 +4,7 @@ index: true
 
 # Apache ECharts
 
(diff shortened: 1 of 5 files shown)


Fil:
Do we need to use echarts.esm.min.js? I tried with

```diff
-export const echarts = () => import("npm:echarts/dist/echarts.esm.min.js");
+export const echarts = () => import("npm:echarts");
```

and it seems to work. (It's also what we document in http://127.0.0.1:3000/lib/echarts). Maybe I'm confused by some caching mystery, though.

mbostock:
They’re the same thing now, with the changes in this PR. I can shorten it.

Fil (approved):
I've pushed two changes:
* in recommendedLibraries, simplify
* in the documentation, add a link to the Echarts website

PR #988, opened 2024-03-06
use concrete datasets in examples rather than default datasets

### LINKED ISSUE #870: Use file attachments instead of sample datasets in default templates
The example project used when creating a new project includes two charts - penguins and a stock chart.  Both use a dataset which is not declared anywhere if you grep the whole project (`penguins` and `aapl`).  

I think these are _globals_ to observable (a passing reference was found in: #24).   It took me a while to figure out what was going on and made me uncertain that I'd understood the concept of loaders.  Perhaps the demo could load the data via the loader mechanism or be explicit about where the data is sourced from.

Thanks for (another) fantastic project.

### LINKED ISSUE #985: make data examples use explicit files
Tracking aapl and penguins “magically” is nice to make quick demos, but it makes things unclear, and also does not help users understand how to replace the data with their own.

We could keep and maybe document the references (#401, #817), but the examples ought to track concrete datasets.

Fil:
closes #870
closes #985

note: the default datasets are still used "magically" in the documentation, for instance:
* docs/css/color.md
* docs/javascript.md


templates/default/docs/index.md
```diff
@@ -1,4 +1,5 @@
 ---
+theme: alt
```
mbostock, line 2: This change isn’t related to the issue, and note the two templates have deviated. Revert this part, or add it to both?

mbostock (approved):

templates/default/docs/index.md
Fil: good catch, sorry for this; it was unfortunate combination of a copy/paste from templates/empty/docs/index.md and an incomplete proof-read

PR #996, opened 2024-03-06
document data loaders' cwd and import.meta.url

### LINKED ISSUE #853: How to debug a data loader
I have tried to implement an R data loader.
It works as a script but it errors in Observable Framework.
I was looking for ways/suggestions on how to debug data loaders in the documentation but failed.
Maybe it could be a useful addition...

Fil:
it's possible that the example code might be too much or too convoluted… we can edit it down if it's not salvageable, and do a real-life complete example instead

closes #853


mbostock (approved):
Nice. I want some code formatting for import.meta.url and docs. But I can fix that when I’m back at my desk.

mbostock:
Do you think it would be easier to understand if the cwd were the data loader’s own directory?

Fil:
I guess it depends where we expect "supporting files" to be saved. I'd say that it feels a bit more natural to have them in docs/ if they are data, but I haven't thought about all the use cases. If your supporting files are in node_modules the current system makes more sense.

PR #1004, opened 2024-03-06
Add skeleton of sql item

naltmann:
Pretty sure an entry for sql is needed, but not sure how it should read


Fil:
probably just a link to the [SQL Guide](./sql) ?

mbostock (approved):

Issue #1011, opened 2024-03-07
Better support for Mosaic?


Fil:
I'd like to reduce the boilerplate needed to run Mosaic and make it compatible with the new [SQL front matter](https://observablehq.com/framework/sql) declarations.

Ideally I would reference a dataset (or several) with:

```yaml
---
sql:
  trips: nyc-taxi.parquet
---
```

and then call

````
```js
const vg = vgplot(sql);
```
````

where vgplot would be a known symbol (or maybe it should be called mosaic… anyway, as a supported library). This would return `vg`, a mosaic instance attached to the same DuckDB instance as the `sql` template literal.

Currently, the most concise code I can come up with is the following 6 LOC:

````
```js
import {createAPIContext, wasmConnector} from "npm:@uwdata/vgplot";

// Instantiate a Mosaic context and a database
const vg = createAPIContext();
const mc = vg.coordinator();  // use default coordinator
mc.databaseConnector(wasmConnector());

// Load the dataset into the database
const trips = await FileAttachment("nyc-taxi.parquet").url();
mc.exec([vg.loadParquet("trips", trips)]);
```
````

6 lines is not awful, but I feel it could be better integrated. Another issue with my current approach is that this is siloed from `sql`, as it loads the parquet file separately and with a different instance (even a different version) of DuckDB.

cc: @jheer I'm sure you'll have suggestions.

jheer:
There are three Mosaic initialization steps to consider:

1. Creating an API context. This is useful if you want to isolate a plot "namespace" (standalone legends may refer to a plot by name), use a custom coordinator, and/or extend the API. However, in most cases we could skip this step and use exported vgplot methods directly (which will then use a default API context).

2. Setting the desired database connector. Mosaic currently defaults to a socket connector. One reason was to aid tree-shaking, so that DuckDB-WASM doesn't get bundled when not needed. However, having seen more usage, I'm more convinced that maybe we should use the wasm connector by default. If we make this change, that's one less init step needed here. *BUT* I'm not sure if that really helps here, as ideally we want to give Mosaic access to the same DuckDB instance that is used to load any `sql` front-matter files.

[...]

PR #1012, opened 2024-03-07
document Mosaic

Fil:
related #1011
mosaic

cc: @jheer

### DIFF
--- a/docs/lib/mosaic.md
+++ b/docs/lib/mosaic.md
@@ -1,3 +1,8 @@
+---
+sql:
+  trips: nyc-taxi.parquet
+---
+
 # Mosaic
 
 [Mosaic](https://uwdata.github.io/mosaic/) is a system for linking data visualizations, tables, and input widgets, all leveraging a database (DuckDB) for scalable processing. With Mosaic, you can interactively visualize and explore millions —and even billions— of data points.
@@ -15,17 +20,6 @@ The map views for pickups and dropoffs are coordinated. You can also select an i
 The code below creates an instance of Mosaic and the three coordinated views. (Please upvote [#1011](https://github.com/observablehq/framework/issues/1011) if you would like better support for Mosaic.)
 
 ```js echo
-import {createAPIContext, wasmConnector} from "npm:@uwdata/vgplot";
-
-// Instantiate a Mosaic context and a database
-const vg = createAPIContext();
-const mc = vg.coordinator();
-mc.databaseConnector(wasmConnector());
-
-// Load the dataset into the database
-const trips = await FileAttachment("nyc-taxi.parquet").url();
-mc.exec([vg.loadParquet("trips", trips)]);
-
 // Create a shared filter
 const $filter = vg.Selection.crossfilter();
 
--- a/docs/lib/mosaic.md
+++ b/docs/lib/mosaic.md
@@ -1,0 +1,88 @@
+# Mosaic
+
+[Mosaic](https://uwdata.github.io/mosaic/) is a system for linking data visualizations, tables, and input widgets, all leveraging a database (DuckDB) for scalable processing. With Mosaic, you can interactively visualize and explore millions —and even billions— of data points.
+
+The example below shows 1 million taxi rides pickup and dropoff points in NYC on Jan 1-3, 2010. The dataset is stored in a 8MB parquet file, originally generated with a data loader.
+
+${maps}
+
+${histogram}
+
+The map views for pickups and dropoffs are coordinated. You can also select an interval in the histogram to filter the maps.
+_What spatial patterns can you find?_
+
+---
+
+The code below creates an instance of Mosaic and the three coordinated views. (Please upvote [#1011](https://github.com/observablehq/framework/issues/1011) if you would like better support for Mosaic.)
+
+```js echo
+import {createAPIContext, wasmConnector} from "npm:@uwdata/vgplot";
+
+// Instantiate a Mosaic context and a database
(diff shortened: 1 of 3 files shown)


docs/lib/nyc-taxi.parquet
mbostock, line 1: Can we use a data loader to make this, or at least include the script to make it alongside this cached output?

Fil: Absolutely. I just didn't want to document a data loader as part of this page, which has enough matter already.

It's not a hugely complicated data loader, except it needs to install the binary duckdb.
```sh
export TMPDIR="docs/.observablehq/cache"
export PATH=$TMPDIR:$PATH
duckdb :memory: << EOF
-- Load spatial extension
INSTALL spatial; LOAD spatial;

-- Project, following the example at https://github.com/duckdb/duckdb_spatial
CREATE TEMP TABLE rides AS SELECT
  pickup_datetime::TIMESTAMP AS datetime,
  ST_Transform(ST_Point(pickup_latitude, pickup_longitude), 'EPSG:4326', 'EPSG:32118') AS pick,
  ST_Transform(ST_Point(dropoff_latitude, dropoff_longitude), 'EPSG:4326', 'EPSG:32118') AS drop
FROM 'https://uwdata.github.io/mosaic-datasets/data/nyc-rides-2010.parquet';

-- Write output parquet file
COPY (SELECT
  HOUR(datetime) + MINUTE(datetime) / 60 AS time,
  ST_X(pick)::INTEGER AS px, -- extract pickup x-coord
  ST_Y(pick)::INTEGER AS py, -- extract pickup y-coord
  ST_X(drop)::INTEGER AS dx, -- extract dropff x-coord
  ST_Y(drop)::INTEGER AS dy  -- extract dropff y-coord
FROM rides
ORDER BY 2,3,4,5,1 -- optimize output size by sorting
) TO '$TMPDIR/trips.parquet' (COMPRESSION 'ZSTD', row_group_size 10000000);
EOF

cat $TMPDIR/trips.parquet >&1  # Write output to stdout
rm $TMPDIR/trips.parquet       # Clean up
```

mbostock: I think have the data-loaded checked in but not functional, even if undocumented, is a big help!

jheer: This looks great. Thanks for including it! I agree that the data loader is an important component of the example, even if non-functional in this particular deployment.

Fil: OK I'll check in a simple version that does not try to work in CI — with the expectation that duckdb is on the $PATH.

docs/lib/mosaic.md
```diff
@@ -0,0 +1,88 @@
+# Mosaic
+
+[Mosaic](https://uwdata.github.io/mosaic/) is a system for linking data visualizations, tables, and input widgets, all leveraging a database (DuckDB) for scalable processing. With Mosaic, you can interactively visualize and explore millions —and even billions— of data points.
+
+The example below shows 1 million taxi rides pickup and dropoff points in NYC on Jan 1-3, 2010. The dataset is stored in a 8MB parquet file, originally generated with a data loader.
```
mbostock, line 5:
```suggestion
The example below shows 1 million taxi rides pickup and dropoff points in New York City on Jan 1–3, 2010. The dataset is stored in a 8MB [Apache Parquet](./arrow#apache-parquet) file, generated with a data loader.
```

(Dropping the “originally” for simplicity. It’s still accurate to say that it was generated by a data loader even if it’s not currently being re-generated by the data loader every time you build.)

docs/lib/mosaic.md
```diff
@@ -0,0 +1,88 @@
...
+
+${maps}
+
+${histogram}
+
+The map views for pickups and dropoffs are coordinated. You can also select an interval in the histogram to filter the maps.
+_What spatial patterns can you find?_
```
mbostock, line 12:
```suggestion
The map views for pickups and dropoffs are coordinated. You can also select an interval in the histogram to filter the maps. _What spatial patterns can you find?_
```

docs/lib/nyc-taxi.parquet.sh
```diff
@@ -0,0 +1,25 @@
+duckdb :memory: << EOF
```
mbostock, line 1: Should this have a shebang? One of these probably…

```suggestion
#!/usr/bin/env bash
```

```suggestion
#!/bin/sh
```

mbostock (approved):

docs/lib/nyc-taxi.parquet.sh
Fil: I don't think so; shebang are only used for .exe?

mbostock: Ah, so even if you have a shebang, it’s ignored because we explicitly run it with `sh`?

Fil: yes sh ignores the shebang and interprets the rest of the code directly

```sh
❯ chmod +x ./test.py
❯ cat ./test.py
#! /usr/bin/env python3
print("hello, world")
❯ bash ./test.py
./test.py: line 2: syntax error near unexpected token `"hello, world"'
./test.py: line 2: `print("hello, world")'
❯ ./test.py
hello, world
```

mbostock: Thanks for educating me. 😄

mbostock:
Waiting for #1015 to land and then we’ll update this.

Issue #845, opened 2024-02-17
Crash on init TypeError ERR_UNKNOWN_FILE_EXTENSION


ggl:
Unfortunately it crashes on init with this message.

```
$  npm init @observablehq
node:internal/errors:477
    ErrorCaptureStackTrace(err);
    ^

TypeError [ERR_UNKNOWN_FILE_EXTENSION]: Unknown file extension ".ts" for /home/ggl/.npm/_npx/f417d01c081ec748/node_modules/@observablehq/framework/bin/observable.ts
    at __node_internal_captureLargerStackTrace (node:internal/errors:477:5)
    at new NodeError (node:internal/errors:387:5)
    at Object.getFileProtocolModuleFormat [as file:] (node:internal/modules/esm/get_format:80:11)
    at defaultGetFormat (node:internal/modules/esm/get_format:122:38)
    at defaultLoad (node:internal/modules/esm/load:81:20)
    at nextLoad (node:internal/modules/esm/loader:173:28)
    at ESMLoader.load (node:internal/modules/esm/loader:616:26)
    at ESMLoader.moduleProvider (node:internal/modules/esm/loader:472:22)
    at new ModuleJob (node:internal/modules/esm/module_job:63:26)
    at #createModuleJob (node:internal/modules/esm/loader:491:17)
    at ESMLoader.getModuleJob (node:internal/modules/esm/loader:449:34)
    at async Promise.all (index 0)
    at async ESMLoader.import (node:internal/modules/esm/loader:541:24)
    at async file:///home/ggl/.npm/_npx/f417d01c081ec748/node_modules/@observablehq/create/bin/observable-create.js:6:1 {
  code: 'ERR_UNKNOWN_FILE_EXTENSION'
}
```
1 info using npm@8.15.0
2 info using node@v18.7.0

Node is from Nodesource 18 branch running on Debian 12.5.

mbostock:
Per the [Getting started guide](https://observablehq.com/framework/getting-started), you must have Node.js 20.6 or later installed before you can install Framework. I’m surprised you made it this far because the engines requirement is stated in the package.json:

https://github.com/observablehq/create/blob/04d7134f1721486e50a0faaaf09bfbf3dc46b12a/package.json#L28-L30

[...]

PR #1017, opened 2024-03-07
Update Node.js version requirement to 20 or later

georgemarrows:
Getting started doc currently says Node 18+ but that doesn't work and this [comment](https://github.com/observablehq/framework/issues/845#issuecomment-1949797130) says 20+ is needed.


Fil:
That message was before https://github.com/observablehq/framework/releases/tag/v1.1.0; Framework works with node 18 now.

An error I've sometimes seen is `Unknown file extension ".ts" for observablehq.config.ts`, but you can remove that file.

If you see something else, please share details.

PR #1021, opened 2024-03-08
re-resolve npm imports; apache-arrow 15

### LINKED ISSUE #1019: Multiple versions of libraries
As noted in https://github.com/observablehq/framework/pull/1015#issuecomment-1984893780, we can get duplicate versions of libraries. Sometimes this is because libraries themselves have non-overlapping version ranges (this is often the case with apache-arrow because they don’t use semantic versioning and frequently bump the major version number), and other times it’s because jsDelivr resolves the exact version of dependencies when a library is published and does not rebuild when dependencies are updated (this is most noticeable in the vega packages, and jsDelivr says that this behavior is by design).

So I think we’ll have to rewrite the imports we get from jsDelivr and re-resolve them against the versions we have locally, rather than respecting the baked-in versions that jsDelivr has. We could do this naively by respecting the major version, or better, we could load the package.json and respect the declared range.

And in the case of apache-arrow, I think we may chose to ignore the declared range, and force all versions to be the same (currently 15.0.1). This is particularly so because some libraries use instanceof checks on Arrow tables, so if you have multiple instances of Arrow loaded, things will break.

All in all, kind of a mess! But fortunately we can compensate for this mess as a static site generator. 🙂

mbostock:
Fixes #1019. Rather than relying on the versions baked-in to jsDelivr, we now download the `package.json` and re-resolve the versions of transitive dependencies. This ensures that you get consistent versions, when possible, of transitive dependencies. In addition, we override @uwdata/mosaic-core’s dependencies on @duckdb/duckdb-wasm and apache-arrow to match Framework’s pinned versions.

Loading the `package.json` is a bit slow, but I can’t think of a faster way to do this. Perhaps we could have an `observable install` command that you can run from the command-line to populate the npm cache for all known pages. We could then run that command as part of `observable create` so you don’t experience slow page loads.

### DIFF
--- a/src/npm.ts
+++ b/src/npm.ts
@@ -159,8 +159,8 @@ export async function getDependencyResolver(
     for (const dependency of dependencies) {
       const {name: depName, path: depPath = "+esm"} = parseNpmSpecifier(dependency.slice("/npm/".length));
       const range =
-        (name === "arquero" || name === "@uwdata/mosaic-core") && depName === "apache-arrow"
-          ? "latest" // force Arquero & Mosaic to use the latest version of Arrow
+        (name === "arquero" || name === "@uwdata/mosaic-core" || name === "@duckdb/duckdb-wasm") && depName === "apache-arrow" // prettier-ignore
+          ? "latest" // force Arquero, Mosaic & DuckDB-Wasm to use the (same) latest version of Arrow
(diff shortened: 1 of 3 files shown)


Fil:
It's slow only the first time you open a page—I don't think that's a huge problem, but yeah it would be even better if this cache could be populated in the background.

Fil (approved):

Issue #92, opened 2023-11-03
Optionally emit page links with “.html” for static file servers that don’t support implicit extensions when routing


pettiross:
I copied my `dist` folder to a basic web server. The index file loaded, but the links to pages like Observable GitHub fail to load because the URL specified was just to `/github` instead of `/github.html`.

nelreina:
> There should definitely be an option for emitting `.html`, and I think perhaps even this should be the default behavior because it works universally. Even `http-server` doesn’t handle serving `/javascript` correctly, for example. #171

This will be handy spacially when want to run in docker with minimal storage footprint. E.g. busybox httpd server is lika < 10mb . Nginx static is like 40mb / if I use node then it will be 260+ MB

PR #1022, opened 2024-03-08
normalize page path="/" to "/index"

### LINKED ISSUE #689: Pager appears when there’s only one page in pages
E.g.,

```js
export default {
  pages: [
    {name: "Home", path: "/"}
  ]
};
```

Kind of a degenerate case, but anyway.

Fil:
this feels wrong… but solves the issue and passes the tests
what am I missing

closes #689


Fil:
This should be covered as part of #92, and probably after #1034.

Fil:
PTAL

Issue #23, opened 2023-10-18
Summary data table display


mbostock:
We should have the equivalent of Observable’s data table display for a nice visual summary of tabular data.

maelp:
I'd love that!

declann:
> I've shown a quick prototype of this using [mosaic](https://uwdata.github.io/mosaic/), which has the benefit of bringing SQL and using great optimization techniques under the hood, allowing to process millions of rows without a sweat.
> 
> [snip]
> 
> It's a bit ugly, but you have to start somewhere:
> <img alt="summary-table" width="698" src="https://private-user-images.githubusercontent.com/7001/279732930-9acb40cf-118c-427c-a2da-885b0aed9757.png?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MjI1MDYxMjIsIm5iZiI6MTcyMjUwNTgyMiwicGF0aCI6Ii83MDAxLzI3OTczMjkzMC05YWNiNDBjZi0xMThjLTQyN2MtYTJkYS04ODViMGFlZDk3NTcucG5nP1gtQW16LUFsZ29yaXRobT1BV1M0LUhNQUMtU0hBMjU2JlgtQW16LUNyZWRlbnRpYWw9QUtJQVZDT0RZTFNBNTNQUUs0WkElMkYyMDI0MDgwMSUyRnVzLWVhc3QtMSUyRnMzJTJGYXdzNF9yZXF1ZXN0JlgtQW16LURhdGU9MjAyNDA4MDFUMDk1MDIyWiZYLUFtei1FeHBpcmVzPTMwMCZYLUFtei1TaWduYXR1cmU9MDBjYTJkNzg0ZThiMzY2OTFjMTBiZGNkMmYwNjFhM2M0NDgwN2NhYjJhZGMxNDg2ZjkxODA0YTk0YmI0YjY5NCZYLUFtei1TaWduZWRIZWFkZXJzPWhvc3QmYWN0b3JfaWQ9MCZrZXlfaWQ9MCZyZXBvX2lkPTAifQ.L60ppHTGxYjac3hm1oNaVyhNIL6m-u80EK_M3HDNyh0">
> 
[...]

PR #1024, opened 2024-03-08
non-streaming DuckDBClient.query

### LINKED ISSUE #1016: DuckDBClient doesn't materialize query results and this breaks PIVOT operations (maybe other operations too?)
If I attempt to PIVOT ON and do not use an IN clause, DuckDB throws the error:

Error: Invalid Error: PendingQuery can only take a single statement

This file illustrates the problem:
[events.md](https://github.com/observablehq/framework/files/14529864/events.md)

If I use stmt.query(...) instead of stmt.send(...) the PIVOT works.  This is what tipped me off: https://duckdb.org/docs/api/wasm/query - "// ... and run the query with materialized results".  

```js
    async simpleQuery(query){
        const connection = await this._db.connect();
        let result
        try{
            result = await connection.query(query);
            return result.toArray();
        } finally {
            await connection.close()
        }
    }
```

I wonder if DuckDBClient could be enhanced to not always stream query results?

mbostock:
Fixes #1016. Changes DuckDBClient.query to use `statement.query` or `connection.query` under the hood, rather than `statement.send` or `connection.send`, which noticeably improves performance. In addition, returns an Arrow Table, rather than copying the results into an array of objects. Interactive SQL queries are faster!

### DIFF
--- a/src/client/stdlib/duckdb.js
+++ b/src/client/stdlib/duckdb.js
@@ -91,7 +91,7 @@ export class DuckDBClient {
       throw error;
     }
     return {
-      schema: getArrowTableSchema(batch.value),
+      schema: batch.value.schema,
       async *readRows() {
         try {
           while (!batch.done) {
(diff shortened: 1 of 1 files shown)


src/client/stdlib/duckdb.js
```diff
@@ -91,7 +91,7 @@ export class DuckDBClient {
       throw error;
     }
     return {
-      schema: getArrowTableSchema(batch.value),
+      schema: batch.value.schema,
```
mbostock, line 94: We’re now deviating from the DatabaseClient specification used in Observable notebooks, but I think it’s better to standardize on the Apache Arrow than our own specification. (Also this schema is unused in Framework — it exists in notebooks to power the data table cell, and we haven’t written than in Framework yet #23, but when we do, it should be based on the Apache Arrow schema.)

Fil (approved):
incredibly fast!

PR #1025, opened 2024-03-08
dark

### LINKED ISSUE #1018: Reactive “dark” variable
```js
const dark = Generators.observe((change) => {
  const media = matchMedia("(prefers-color-scheme: dark)");
  const changed = (event) => change(event.matches);
  media.addEventListener("change", changed);
  change(media.matches);
  return () => media.removeEventListener("change", changed);
});
```

Fil:
closes #1018 

(loosely related: #903, just because I like to make links)

### DIFF
--- a/docs/lib/generators.md
+++ b/docs/lib/generators.md
@@ -74,3 +74,39 @@ const width = Generators.width(document.querySelector("main"));
 ```js echo
 width
 ```
+
+## dark()
+
+Returns an async generator that yields a boolean indicating whether the page is currently displayed with a dark [color-scheme](https://developer.mozilla.org/en-US/docs/Web/CSS/color-scheme).
+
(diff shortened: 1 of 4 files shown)


docs/lib/generators.md
```diff
@@ -74,3 +74,39 @@ const width = Generators.width(document.querySelector("main"));
 ```js echo
 width
 ```
+
+## dark()
+
+Returns an async generator that yields a boolean indicating whether the page is currently displayed with a dark [color-scheme](https://developer.mozilla.org/en-US/docs/Web/CSS/color-scheme).
+
+If the page supports dark mode (for example with the default Framework themes), the value reflects the user’s prefered color scheme. It will yield if that value changes, and update the charts that depend on it without needing to reload the page — this happens at dusk (if the user settings adapt from light to dark), and conversely at dawn, from dark to light; and of course, when the user is playing with their settings.
```
mbostock, line 82:
```suggestion
If the page supports dark mode (for example with the default Framework themes), the value reflects the user’s preferred color scheme. It will yield if that value changes, and update the charts that depend on it without needing to reload the page — this happens at dusk (if the user settings adapt from light to dark), and conversely at dawn, from dark to light; and of course, when the user is playing with their settings.
```

PR #1031, opened 2024-03-09
linkify carefully

Fil:
I know this can be done with markdownit plugins, but I feel that it is a basic requirement and I'm surprised each time I type a URL and it doesn't link.

However I disable so-called "fuzzy links", because 1. they're too invasive and risk matching on false positive text and 2. they're wrong (e.g. they use http:// when we'd most of the time want https:// anyway).

Relevant API documentation: https://markdown-it.github.io/linkify-it/doc/

### DIFF
--- a/src/markdown.ts
+++ b/src/markdown.ts
@@ -279,12 +279,11 @@ export interface ParseOptions {
   style?: Config["style"];
 }
 
-export function parseMarkdown(
-  input: string,
-  {root, path, markdownIt = (md) => md, style: configStyle}: ParseOptions
-): MarkdownPage {
+export function parseMarkdown(input: string, {root, path, markdownIt, style: configStyle}: ParseOptions): MarkdownPage {
   const parts = matter(input, {});
-  const md = markdownIt(MarkdownIt({html: true}));
+  let md = MarkdownIt({html: true, linkify: true});
+  md.linkify.set({fuzzyLink: false, fuzzyEmail: false});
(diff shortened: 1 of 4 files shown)


Fil:
(I'd also be in favor of adding the base typographer option: cf. https://github.com/markdown-it/markdown-it/blob/master/lib/rules_core/replacements.mjs)

src/markdown.ts
```diff
@@ -279,12 +279,11 @@ export interface ParseOptions {
...
-export function parseMarkdown(
-  input: string,
-  {root, path, markdownIt = (md) => md, style: configStyle}: ParseOptions
-): MarkdownPage {
+export function parseMarkdown(input: string, {root, path, markdownIt, style: configStyle}: ParseOptions): MarkdownPage {
   const parts = matter(input, {});
-  const md = markdownIt(MarkdownIt({html: true}));
+  let md = MarkdownIt({html: true, linkify: true});
+  md.linkify.set({fuzzyLink: false, fuzzyEmail: false});
+  if (typeof markdownIt === "function") md = markdownIt(md);
```
mbostock, line 286: We shouldn’t ignore non-functions here; we should error.

```suggestion
  if (typeof markdownIt !== undefined) md = markdownIt(md);
```

mbostock (approved):
GitHub seems to do fuzzyEmail, so I’d be fine with turning that on, too. But fine either way.

PR #1035, opened 2024-03-10
resolve script[src]

mbostock:
For JavaScript scripts (see [script[type]](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script#type)), the `src` attribute is resolved as an import; for non-JavaScript scripts, it is resolved as a file. This means you can now load vanilla JavaScript scripts (or modules), and it “just works”. For module scripts, the module is also preloaded.

### DIFF
--- a/src/html.ts
+++ b/src/html.ts
@@ -18,7 +18,9 @@ const ASSET_PROPERTIES: readonly [selector: string, src: string][] = [
 ];
 
 export function isJavaScript({type}: HTMLScriptElement): boolean {
-  return !type || type === "text/javascript" || type === "application/javascript" || type === "module";
+  if (!type) return true;
+  type = type.toLowerCase();
+  return type === "text/javascript" || type === "application/javascript" || type === "module";
 }
 
 export function parseHtml(html: string): DOMWindow {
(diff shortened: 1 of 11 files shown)


mbostock:
Hmm, so maybe scripts need to go into `_import` instead of `_file`? Because they should be allowed to do a transitive `import` of a local module. I guess I can’t see a strong reason to use `<script src>` instead of `import` in a JavaScript fenced code block but I feel like they should behave the same way at least.

Fil:
It's difficult to think of all the use cases.  Concretely when would a script[src] want to use another local script? Would this make sense only with [type=module]?

I'm also wondering if and how this is could help with the [coolprop](https://github.com/observablehq/framework/discussions/1033#discussioncomment-8736881) example, which wants to load a companion .wasm file. For now I haven't found a way to make this work locally both in preview and build.

mbostock:
Okay, this is ready! I was motivated after another support request https://github.com/observablehq/framework/discussions/1061 came up.

Fil (approved):
Nice! Do you think it would be possible to alias the script (from `foo.js?sha=1234` to `foo.1234.js`) in both cases? Or is there something that makes it impossible in principle?

mbostock:
> Do you think it would be possible to alias the script (from `foo.js?sha=1234` to `foo.1234.js`) in both cases?

Oops, that’s a bug. If we don’t fix this the built site is broken!

PR #1037, opened 2024-03-11
normalize links

### LINKED ISSUE #92: Optionally emit page links with “.html” for static file servers that don’t support implicit extensions when routing

Fil:
This introduces a **cleanUrls** option (boolean, defaults to true) that indicates whether links to pages should end with ".html" (except for directories, where the convention is that the file `dir/index.html` is served with the `dir/` path).

This option is subsumed in the definition of `config.md.normalizeLinks`, which is applied to all links generated by Markdown text and now also by links generated in the sidebar, in the pager, and in the search results. (It is not applied to links that are written directly in HTML.)

closes #92 

based on #1034

This could also help re-implement a normalization of the pager (#689 #1022)


test/output/build/space-page/index.html
```diff
@@ -23,7 +23,7 @@
     <li class="observablehq-link observablehq-link-active"><a href="./">Home</a></li>
   </ol>
   <ol>
-    <li class="observablehq-link"><a href="./a space">A page</a></li>
+    <li class="observablehq-link"><a href="./a%20space">A page</a></li>
```
Fil, line 26: this is a side benefit of normalization

Fil:
`observable deploy` tested by manually deploying with this option on.

src/markdown.ts
```diff
@@ -282,7 +282,21 @@ export interface ParseOptions {
   md?: MarkdownIt;
 }
 
-export function mdparser({markdownIt}: {markdownIt?: (md: MarkdownIt) => MarkdownIt} = {}): MarkdownIt {
+function cleanPath(href: string, clean: boolean): string {
+  if (href.startsWith("./") || href.startsWith("../")) {
+    const g = href.match(/(?<dir>[^#?]*\/)((?<file>[^/?#]*?)([.]html)?)(?<q>[?#].*)?$/)?.groups;
```
mbostock, line 287: Maybe we should use `URL` to parse, and `dirname`, `basename`, `extname`? Regular expressions…

src/preview.ts
```diff
@@ -188,12 +188,21 @@ export class PreviewServer {
...
-        // Check for the existence of the .md file first.
+        // If this path ends with .html and cleanUrls are enabled, then redirect
+        // to drop the .html.
         if (extname(path) === ".html") {
-          res.writeHead(302, {Location: join(dirname(pathname), basename(pathname, ".html")) + url.search});
-          res.end();
-          return;
+          path = path.slice(0, -".html".length);
+          pathname = pathname.slice(0, -".html".length);
+          if (config.md.normalizeLink("./hello") === "./hello") {
```
mbostock, line 196: This seems a little icky. Wouldn’t it be better to check the value of `config.md.normalizeLink(path)` instead, and then redirect to the normalized value? Otherwise we’re essentially duplicating the logic of `normalizeLink` and assuming its behavior. Or if we want to assume its behavior, then we should expose `config.cleanUrls` instead of trying to invert it by calling `md.normalizeLink`.

src/search.ts
```diff
@@ -81,7 +81,8 @@ export async function searchIndex(config: Config, effects = defaultEffects): Pro
         options: {
           fields: indexOptions.fields,
           storeFields: indexOptions.storeFields
-        }
+        },
+        cleanUrls: config.md.normalizeLink("./hello") === "./hello"
```
mbostock, line 85: Similar to above. If we expect `md.normalizeLink` to be overridden and have some other behavior, then we need to pass the normalized links to minisearch so that it renders normalized links on the client. On the other hand, if we only want to support our more limited definition of `cleanUrls`, then we shouldn’t test the behavior of `md.normalizeLink`; we should explicitly track the setting in the config.

I guess it’d be cool to respect the implementation of `md.normalizeLink`, if it’s not too much work. Perhaps we could invoke it up above when computing the `id` rather than having the client do the normalization and repeating the logic?

mbostock:
I moved all the normalization to the server-side and got rid of the `./hello` test. I also changed the preview server to respect the `normalizeLink` implementation, which simplified things quite a bit! 👍

Issue #687, opened 2024-02-06
Show better progress during deploy


mbostock:
Yuri writes:

> I deployed the site from a slow internet connection a few times and occasionally saw a large time gap with no indication of progress (the progress report would stay on the same step for ~30 seconds).

Maybe we can use clack or similar for a progress bar?

PR #1042, opened 2024-03-12
deploy from output directory instead of doing an in-memory build

mythmon:
With this PR, `observable deploy` will deploy files from the configured output directory instead of creating its own build. If the newest file in the dist directory is more than 5 minutes old, the build is considered "stale". I happy to adjust that value, or even make it configurable if we want to. If the dist directory is missing or empty the build is "missing".

~~There is a new `--build` flag that has several possible values:~~

This flag has changed. See further down the thread for more details.

|                     | fresh  | stale  | missing |
| ------------------- | ------ | ------ | ------- |
| `--build=prompt`    | deploy | prompt | prompt  |
| `--build=always`    | build  | build  | build   |
| `--build=if-needed` | deploy | build  | build   |
| `--build=never`     | deploy | error  | error   |

`--build=prompt` is the effective default, `--build=always` is similiar to the old default.


src/deploy.ts
```diff
@@ -299,8 +388,14 @@ export async function deploy(
...
 
-  // Build the project
-  await build({config}, new DeployBuildEffects(apiClient, deployId, effects));
+  // Upload the files
+  const uploadSpinner = clack.spinner();
+  uploadSpinner.start("");
+  await runAllWithConcurrencyLimit(buildFilePaths, async (path, i) => {
+    uploadSpinner.message(`${i + 1} / ${buildFilePaths!.length} ${path.slice(0, effects.outputColumns - 10)}`);
+    await apiClient.postDeployFile(deployId, join(config.output, path), path);
+  });
```
mbostock, line 397: Concurrency! Nice. Related #687 (though we still don’t show intermediate progress for each file). Also note that clack has a [`tasks`](https://github.com/natemoo-re/clack/blob/0cb17d3bed21f8c8569d91a5b1ece884ea3c7ee2/packages/prompts/src/index.ts#L801-L813) method for managing concurrent spinners.

Fil:
wow upload is so fast now!

I was able to get a deploy of pangea (a project with 300+ pages), which used to take about 350s, happen in ~49s. My baseline for this project is ~1’15”.

However I haven't been successful with the full build+deploy process:

`> yarn deploy --build=always -m"not much"`

because it seems I'm always hitting some rate limit:
■  Error: Unexpected response status 429

It is particularly frustrating because it happens after about 2‘30” of activity, often near the end of the uploads; for example:

```
◓  563 / 579 _npm/@duckdb/duckdb-wasm@1.28.1-dev99.0/+esm.js│
■  Error: Unexpected response status 429
```

(it's not always the same file, but often around the 500+ file mark.)


We'll probably need to 1. up that rate limit or check it differently so that it doesn't catch you in mid-flight, and 2. if it happens, make the client display a message indicating that the server wants some air, pause and retry, rather than fail.

Note that I (as a user) am not trying to do anything else than a normal "deploy", so I get a bad feeling of getting “blamed” or “rejected” for something I didn't do or that is out of my control, and without any way to solve the issue except retry later hoping that it might work.

Paradoxically, this is almost orthogonal to what you're doing: I guess it's just because it's faster, and the server doesn't expect that! But it's blocking.

Fil:
One thing that is not immediately clear to me (when I don't read the docs above or forget about them) is what the `--build=if-needed` option does. Rebuild only if the build is missing, or if it's stale? We could clarify by having two options: 
* `--build=if-missing-or-stale` (probably abbreviated as `--build=if-stale`, it would be a rename of `if-needed`). 
* `--build=if-missing` would add a row to the table with  (fresh = deploy | stale =deploy  | missing = build).

mythmon:
I waffled around a bit about the various `--build` options. I ended up not including a `--build=if-missing` because I couldn't think if a concrete time I'd want to use it. I'm even on the fence about `--build=if-stale`. Do you have use cases for those?

Today I'm thinking about inverting the options. something like `--on-stale=build` and `--on-missing=error`. I think that might help with combinatoric growth of the `--build=` version. What do you think about that?

Now to go make it handle 429 better.

Fil:
I think `--if-missing` makes sense as a use case, for when you want to deploy a project as quickly as possible, without having to think about its current build status. If it was not built then it needs a build prior to that, otherwise just use the build.

Splitting that option into two flags probably makes sense.

src/deploy.ts
mythmon: I think that `clack.tasks` can only handle serial tasks, and wouldn't know what to do with the concurrent ones we have here. It would be fun to get a better UI here at some point though.

mythmon:
The recent commits

- If the server sends a [429](https://http.cat/429), check for a "retry-after" header, and wait as long as instructed
- If the project has more than 300 files, slow the rate of uploads to a rate that shouldn't hit rate limits
- Split the `--build` flag to `--if-missing` and `--if-stale` flags

Fil:
Up-to-date options table:

| option | supported |
| -- | -- |
| `--if-missing` | build, cancel, **prompt** |
| `--if-stale` | build, cancel, deploy, **prompt** |

build means "build then deploy", and deploy means "deploy stale"; the default for each of these options is prompt.

mythmon:
This PR has introduced flakiness in two specific tests:

- `getResolvers(page, {root, path}) - resolves directly-attached files`
- `getResolvers(page, {root, path}) - detects file methods`

I'm a little puzzled as to why those are affected, and I haven't been able to reproduce it locally. I suspect there is some degree of test pollution from the concurrent upload.

mcglincy (approved):
Code change LGTM, assuming we can resolve flakiness in getResolvers() tests.

mbostock:
The flakiness is probably fixed in #1076 if you want to rebase on that.

src/bin/observable.ts
```diff
@@ -105,19 +105,46 @@ try {
...
+            description: `What to do if the output directory is stale: ${staleDescription}`
+          },
+          "if-missing": {
+            type: "string",
+            description: `What to do if the output directory is missing: ${missingDescription}`
           }
         }
       });
+      if (ifStale && ifStale !== "prompt" && ifStale !== "build" && ifStale !== "cancel" && ifStale !== "deploy") {
+        console.log(`Invalid --ifStale option: ${ifStale}, expected ${staleDescription}`);
```
mbostock, line 130:
```suggestion
        console.log(`Invalid --if-stale option: ${ifStale}, expected ${staleDescription}`);
```

src/bin/observable.ts
```diff
@@ -105,19 +105,46 @@ try {
...
+        process.exit(1);
+      }
+      if (ifMissing && ifMissing !== "prompt" && ifMissing !== "build" && ifMissing !== "cancel") {
+        console.log(`Invalid --ifMissing option: ${ifMissing}, expected ${missingDescription}`);
```
mbostock, line 134:
```suggestion
        console.log(`Invalid --if-missing option: ${ifMissing}, expected ${missingDescription}`);
```

src/deferred.ts
mbostock, line 1: This is only used by tests. Can we move it back into test?

mythmon:
Looks like those `getResolver` tests are still flaky with the changes from 1076.

mythmon:
Adding the undici mock for jsdelivr fixed those tests. I re-ran the last CI job 4 times without any failures, which is much more reliable than it was before that change.

One of the failed tests involve resolving a npm specifier. I'm less certain about the relation for the other test. I'm also not sure why the changes here triggered this. I'm going to merge this though, since it isn't breaking CI anymore.

PR #1044, opened 2024-03-12
save search queries (plural)

Fil:
This tries and implement the suggestions in https://github.com/observablehq/framework/pull/897#issuecomment-1974801238

I don't find this very usable. To go from one search result to the next, this requires typing:
* cmd-/  to focus the search input
* arrowUp to reactivate the previous query
* arrowRight or Enter to unselect the query in order to be allowed to arrowDown
* arrowDown to select the next search result
* Enter to follow the link.

Furthermore, there is a bit of confusion when you type something, then select it, then navigate up in the search query history. (I think I got it to work, but :sweat:).

Maybe something is lacking in the implementation.


Fil:
I still prefer #897

PR #1048, opened 2024-03-13
1.3.0

mbostock:


test/output/build/imports/foo/foo.html
```diff
@@ -21,7 +21,7 @@
 import {define} from "../_observablehq/client.js";
 import {registerFile} from "../_observablehq/stdlib.js";
 
-registerFile("../top.js", {"name":"../top.js","mimeType":"application/javascript","path":"../_file/top.a53c5d5b.js"});
+registerFile("../top.js", {"name":"../top.js","mimeType":"text/javascript","path":"../_file/top.a53c5d5b.js"});
```
Fil, line 24: mime4 returns 'text/javascript' where mime3 returns 'application/javascript'

but their documentation is not up-to-date!
https://www.npmjs.com/package/mime#api reads:
```
mime.getType('js');             // ⇨ 'application/javascript'
```

I think the source of truth is https://www.iana.org/assignments/media-types/media-types.xhtml ; it says that text/javascript is correct and obsoletes the others, so we're fine.

Fil (approved):

test/output/build/imports/foo/foo.html
mbostock: Yep, I was going to explain this but you figured it out. 👍

PR #1051, opened 2024-03-14
lastModified

### LINKED ISSUE #62: FileAttachment should expose last-modified time
How can the page know the "data snapshot time"? I feel it's important to have it (maybe to display a date somewhere on the dashboard, or use it to archive this version), and I don't think we can trust the mtime of the build _file/cache/xxx.csv

Fil:
Adds a lastModified property to registered files.

This property is computed as the modification time of, in order:
1. the file, if it exists (obvi!)
2. the cache generated by the corresponding data loader, if it exists
3. the data loader (the script that generates the data)

On build, this happens when rendering the pages, which is _after_ the (stale) data loaders have run. This means that the time is consistently given by 1 or 2—and we get to 3 only in the rare case where the data loader has failed to produce a cached artifact.

In preview, by contrast, this happens _before_ the data loader has had a chance to run. So, when the cache is stale, you get stale information—but the situation solves itself on the next page load. (It is not possible to solve this in any other way as we can't predict the future, and don't want to wait for the loader to have finished running before we send the page to the browser.)

At the same time and with the same logic (1, 2, 3), the hash of a file generated by a data loader is now based on the cached result rather than on the data loader code.

closes #62 
supersedes #697
supersedes #699

Notes:

Follows the same pattern as [`File.lastModified`](https://developer.mozilla.org/en-US/docs/Web/API/File/lastModified) (a number not a `Date` because numbers are immutable).

Does not use `touch` for unit tests, so only checks that there is (or isn't) a timestamp.


src/render.ts
```diff
@@ -30,7 +30,8 @@ export async function renderPage(page: MarkdownPage, options: RenderOptions & Re
   const sidebar = data?.sidebar !== undefined ? Boolean(data.sidebar) : options.sidebar;
   const toc = mergeToc(data?.toc, options.toc);
   const draft = Boolean(data?.draft);
-  const {files, resolveFile, resolveImport} = resolvers;
+  const {files, resolveFile, resolveImport, lastModified} = resolvers;
+
```
Fil, line 34:
```suggestion
```

src/dataloader.ts
```diff
@@ -129,11 +133,33 @@ export class LoaderResolver {
     return FileWatchers.of(this, path, watchPaths, callback);
   }
 
-  getFileHash(path: string): string {
+  // Compute file Hash and update the lastModified map. For data loaders, use
```
Fil, line 136:
```suggestion
  // Compute file hash and update the lastModified map. For data loaders, use
```

docs/javascript/files.md
```diff
@@ -32,7 +32,7 @@ volcano
 
 ### Static analysis
 
-The `FileAttachment` function can _only_ be passed a static string literal; constructing a dynamic path such as `FileAttachment("my" + "file.csv")` is invalid syntax. Static analysis is used to invoke [data loaders](../loaders) at build time, and ensures that only referenced files are included in the generated output during build. In the future [#260](https://github.com/observablehq/framework/issues/260), it will also allow content hashes for cache breaking during deploy.
```
Fil, line 35: This "in the future" comment was already obsolete.

docs/javascript/files.md
```diff
@@ -32,7 +32,7 @@ volcano
...
+The `FileAttachment` function can _only_ be passed a static string literal; constructing a dynamic path such as `FileAttachment("my" + "file.csv")` is invalid syntax. Static analysis is used to invoke [data loaders](../loaders) at build time, and ensures that only referenced files are included in the generated output during build. This also allows to include a content hash in the file name for cache breaking during deploy.
```
mbostock, line 35:
```suggestion
The `FileAttachment` function can _only_ be passed a static string literal; constructing a dynamic path such as `FileAttachment("my" + "file.csv")` is invalid syntax. Static analysis is used to invoke [data loaders](../loaders) at build time, and ensures that only referenced files are included in the generated output during build. This also allows content hashes in the file name for cache breaking during deploy.
```

docs/javascript/files.md
```diff
@@ -52,6 +52,10 @@ const frames = [
 
 None of the files in `frames` above are loaded until a [content method](#supported-formats) is invoked, for example by saying `frames[0].image()`.
 
+### Edge cases
+
```
mbostock, line 56:
```suggestion
```

docs/javascript/files.md
```diff
@@ -52,6 +52,10 @@ const frames = [
...
+Missing files don’t exhibit a lastModified property. To determine the file’s MIME type, its extension is checked against the [mime](https://www.npmjs.com/package/mime) database; it defaults to `application/octet-stream`.
```
mbostock, line 57:
```suggestion
For missing files, `file.lastModified` is undefined. The’s MIME type is determined by checking the file extension against the [`mime-db` media type database](https://github.com/jshttp/mime-db); it defaults to `application/octet-stream`.
```

src/dataloader.ts
```diff
@@ -48,6 +50,7 @@ export interface LoaderOptions {
 export class LoaderResolver {
   private readonly root: string;
   private readonly interpreters: Map<string, string[]>;
+  lastModified: Map<string, number>;
```
mbostock, line 53: A risk with this design is that you try to access the `lastModified` time from the map before it’s been initialized (because you haven’t called `getFileHash` yet). This isn’t a problem with the other sets/maps that `getResolvers` returns because those maps are fully-initialized when `getResolvers` returns, and subsequently never mutated. But it is a problem here because the map is mutable, and visible before the map is populated.

I would therefore instead prefer that the `LoaderResolver` class expose a method `getFileLastModified` that returns the (indirect) last-modified time of the given file. I also probably wouldn’t cache this information. (Note that we store `mtimeMs` in the `fileInfoCache` used by `getFileInfo`, but that’s just used to invalidate the cache — we check the current `mtimeMs` every time we read from `fileInfoCache`. Calling `statSync` should be fast.)

mbostock (changes requested):
Functionally correct, but I would like to redesign as described above.

Fil:
I've changed the logic and added a few tests.

src/client/stdlib/fileAttachment.js
```diff
@@ -28,9 +28,10 @@ async function dsv(file, delimiter, {array = false, typed = false} = {}) {
 }
 
 export class AbstractFile {
-  constructor(name, mimeType = "application/octet-stream") {
-    Object.defineProperty(this, "name", {value: `${name}`, enumerable: true});
+  constructor(name, mimeType, lastModified) {
```
mbostock, line 31: Looks like you (inadvertently?) dropped the default for `mimeType` here. I think we want to keep that (and I’m surprised there’s not a test). Also this is part of a public API so we wouldn’t want to remove the default in any case.

src/client/stdlib/fileAttachment.js
```diff
@@ -28,9 +28,10 @@ async function dsv(file, delimiter, {array = false, typed = false} = {}) {
...
     Object.defineProperty(this, "mimeType", {value: `${mimeType}`, enumerable: true});
+    Object.defineProperty(this, "name", {value: `${name}`, enumerable: true});
+    if (lastModified !== undefined) Object.defineProperty(this, "lastModified", {value: Number(lastModified), enumerable: true}); // prettier-ignore
```
mbostock, line 34: Fine to always include this property, even if undefined?

mbostock (commented):
Looks good other than the two tiny things!

src/client/stdlib/fileAttachment.js
Fil: Oops, good catch. I'll add a test!

Fil: Ah now I remember—I wanted to change this so that it would be defined in the call to registerFile instead of in the client.

So, in render.ts we'd have:

```diff
-    mimeType: mime.getType(name) ?? undefined,
+    mimeType: mime.getType(name) ?? "application/octet-stream",
```

and then we don't need this default in the client (and it becomes possible to do a "build test").

src/client/stdlib/fileAttachment.js
Fil: We still need a test because we don't want to show NaN for an undefined value?

src/client/stdlib/fileAttachment.js
Fil: I've done that now; we can roll back and add the default to client/stdlib, but then I don't know how to test.

Fil:
With this architecture I think we'll be able to do even better, for example by adding the file size (if known), image dimensions for common formats, etc.

src/client/stdlib/fileAttachment.js
mbostock: Yes, please move the default back to the client.

Fil: done

mbostock (approved):
💯

PR #1058, opened 2024-03-14
apply encodeURI to imports

### LINKED ISSUE #1055: A + in a URL is sometimes decoded as a space
This affects at least one user who reported that their site hosted on S3 was broken. `dist/_npm/d3@7.9.0/+esm.js` was 403, but `dist/_npm/d3@7.9.0/%2Besm.js` was 200.

We could (must) url-encode the `src`. But, it's going to look ugly… maybe we can find a better way. (Also we should make sure that we url-encode everything that needs to be encoded.)

Fil:
closes #1055

### DIFF
--- a/test/output/build/imports/foo/foo.html
+++ b/test/output/build/imports/foo/foo.html
@@ -10,7 +10,7 @@
 <link rel="modulepreload" href="../_observablehq/client.js">
 <link rel="modulepreload" href="../_observablehq/runtime.js">
 <link rel="modulepreload" href="../_observablehq/stdlib.js">
-<link rel="modulepreload" href="../_npm/d3@7.8.5/+esm.js">
+<link rel="modulepreload" href="../_npm/d3@7.8.5/_esm.js">
 <link rel="modulepreload" href="../_import/bar/bar.13bb8056.js">
 <link rel="modulepreload" href="../_import/top.160847a6.js">
 <link rel="modulepreload" href="../_import/foo/foo bar.b173d3de.js">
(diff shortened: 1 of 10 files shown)


test/output/build/imports/foo/foo.html
```diff
@@ -13,7 +13,7 @@
 <link rel="modulepreload" href="../_npm/d3@7.8.5/_esm.js">
 <link rel="modulepreload" href="../_import/bar/bar.13bb8056.js">
 <link rel="modulepreload" href="../_import/top.160847a6.js">
-<link rel="modulepreload" href="../_import/foo/foo bar.b173d3de.js">
+<link rel="modulepreload" href="../_import/foo/foo%20bar.b173d3de.js">
```
Fil, line 16: this is a welcome change

mbostock (changes requested):
This doesn’t help with #1055 because [`encodeURI`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURI) doesn’t escape the `+` character because it “may part of the URI syntax.” 😅 

```js
encodeURI("/_npm/d3@7.9.0/+esm.js") // "/_npm/d3@7.9.0/+esm.js"
```

`encodeURIComponent` _does_ encode the `+`, but then it also encodes slashes, so it’s not what we want either.

```js
encodeURI("/_npm/d3@7.9.0/+esm.js") // "%2F_npm%2Fd3%407.9.0%2F%2Besm.js"
```

I don’t actually know what the right solution is here. I think we should investigate whether other people have run into similar issues with S3; this might be an S3-specific problem.

mbostock:
This suggests that it’s an S3-specific bug. So we could specifically accommodate `+` in the file name if we wanted to, but we’d have to parse the URL and only rewrite `+` in the file name (not the query!).

https://stackoverflow.com/questions/38282932/amazon-s3-url-being-encoded-to-2

Maybe we do nothing and just avoid this bug in S3?

mbostock (commented):
Also, there are lots of other places where we render `src` and `href` attributes that might need this type of escaping, right? Like everything that gets passed through `normalizeLink` and all the asset attributes in `rewriteHtml`? I don’t know how we could do this universally.

Fil:
Apparently linkedom and jsdom don't care about this either:

```js
import {parseHTML} from "linkedom";
const {document} = parseHTML("<a>");
document.querySelector("a").setAttribute("href", "a b+c\"d'e$f#g&h");
console.log(`${document}`); // <a href="a b+c&quot;d'e$f#g&h"></a>

const document2 = await import("jsdom").then(({JSDOM}) => new JSDOM("").window.document);
const a = document2.createElement("a");
a.setAttribute("href", "a b+c\"d'e$f#g&h");
console.log(`${a.outerHTML}`); // <a href="a b+c&quot;d'e$f#g&amp;h"></a>
```

so yeah, let's just be careful. There are a few other characters that won't work, like : for file names, emojis (#783 ), long filenames on some OS… I don't think we can fix all of that.

PR #1060, opened 2024-03-15
lazily load DuckDB during preview

mbostock:
#844 had the unfortunate behavior that DuckDB is loaded even if it’s not used because we might need to call `registerTable`. Now we load DuckDB lazily only when we need to call `registerTable`.

Fortunately this only affects preview and not build! 😬


Fil (approved):
👍

PR #1065, opened 2024-03-15
Add support for DuckDB database files to sql and DuckDBClient.of

### LINKED ISSUE #1057: DuckDBClient.of should support the native file format
DuckDB has a native file format which can be used to create a [persistent database](https://duckdb.org/docs/connect/overview#persistent-database). In my case I have a data pipeline in  [`dbt-duckdb`](https://github.com/duckdb/dbt-duckdb) and the output is a single named file that I'd like to analyse with framework. I've got it working locally with the following addition to [`stdlib/duckdb.js`](https://github.com/observablehq/framework/blob/5ed90d864ef0ad113a57df6c1021fbecf6723daf/src/client/stdlib/duckdb.js#L256):

```js
if (/\.duckdb$/i.test(file.name)) {
    return await connection.query(`ATTACH '${file.name}' AS ${name} (READ_ONLY)`);
}
```

It's worth noting that the file extension appears to be arbitrary, and this approach would not support `.db` or `.ddb` files. It's also possible that users may want to attach [other types of database](https://duckdb.org/2024/01/26/multi-database-support-in-duckdb.html) to the client, but that's probably worth a separate issue.

Fil:
````md
---
sql:
  base: database.db
---

```sql
SELECT * FROM base.tablename
```
````

closes #1057

@bayre

### CODE REVIEW

File: src/client/stdlib/duckdb.js [hunk 1]
```diff
@@ -254,6 +254,14 @@ async function insertFile(database, name, file, options) {
         if (/\.parquet$/i.test(file.name)) {
           return await connection.query(`CREATE VIEW '${name}' AS SELECT * FROM parquet_scan('${file.name}')`);
         }
+        if (/\.(db|ddb|duckdb)$/i.test(file.name)) {
+          // https://duckdb.org/internals/storage.html#storage-header
+          const magic = await fetch(await file.url(), {headers: {range: "bytes=0-11"}})
+            .then((d) => d.blob())
+            .then((d) => d.slice(8, 12).text());
+          if (magic === "DUCK") return await connection.query(`ATTACH '${file.name}' AS ${name} (READ_ONLY)`);
+          else console.warn(`${file.name} does not appear to be a valid DuckDB database file.`);
```
mbostock, line 263: Should we just pass any file through and let `connection.query` error as needed?

Fil: Yes we should probably do that—maybe it will bork the DuckDB instance in some cases, and the user will have to reload the page, but it will be simpler (for us) and a better separation of concerns. All the cases we want to support for now work in both cases.

In the future, we could extend the sql front matter syntax to pass a type, esp. for remote files:

```yaml
sql:
  gaia:
    source: "an opaque external url"
    type: text/csv
```

More generally in this file, the code branches based on file.name or file.mimeType, depending on something that isn't specified: is it because the remote file could be served with a wrong file name and a correct mime type in some cases? It's something we'll have to clarify.

But for now, yeah, we can do the simplest. The native DuckDB errors look like this:
* `Error: Invalid Error: Opening file 'base.udb' failed with error: Failed to open file: unknown.db`
* `Error: IO Error: The file "index.md" exists, but it is not a valid DuckDB database file!`

I think they're quite expressive and understandable!

docs/lib/duckdb.md
```diff
@@ -53,7 +53,21 @@ Plot.plot({
 })
 ```
 
-For externally-hosted data, you can create an empty `DuckDBClient` and load a table from a SQL query, say using [`read_parquet`](https://duckdb.org/docs/guides/import/parquet_import) or [`read_csv`](https://duckdb.org/docs/guides/import/csv_import).
+When using file attachments, the following formats are supported for tables: [CSV](./lib/csv), [TSV](./lib/csv), [JSON](./javascript/files#json), [Apache Arrow](./lib/arrow), [Apache Parquet](./lib/arrow#apache-parquet).
+
+You can also [attach](https://duckdb.org/docs/sql/statements/attach) a complete database (saved as DuckDB file, typically using the extension `.db`, `.ddb`, or `.duckdb`):
+
+```js echo
+const db2 = await DuckDBClient.of({base: FileAttachment("quakes.db")});
```
mbostock, line 61: If you say this

```suggestion
const db2 = await DuckDBClient.of({main: FileAttachment("quakes.db")});
```

can you then say

```sql
SELECT COUNT() FROM events
```

? I think it should work since`main` is the default schema. Then maybe we could also support this syntax:

```js
const db2 = await DuckDBClient.of(FileAttachment("quakes.db"));
```

Or even:

```js
const db2 = FileAttachment("quakes.db").duckdb();
```

Fil: I couldn't get `main` to work (but it is doing something, as calling a `main.events` table in that case… errors).

mbostock: Okay, let’s just start here then. 👍

src/client/stdlib/duckdb.js
```diff
@@ -255,7 +255,8 @@ async function insertFile(database, name, file, options) {
         if (/\.parquet$/i.test(file.name)) {
           return await connection.query(`CREATE VIEW '${name}' AS SELECT * FROM parquet_scan('${file.name}')`);
         }
-        throw new Error(`unknown file type: ${file.mimeType}`);
+        // Lastly, let DuckDB guess: it may be a .duckdb (or .db or .ddb) database file.
+        return await connection.query(`ATTACH '${file.name}' AS ${name} (READ_ONLY)`);
```
mbostock, line 259: I changed my mind here and would like to be more conservative and only do this with `.db`, `.duckdb`, `.ddb`. It’ll make it easier for us to introduce new formats in the future, and encourage users to be more consistent with how they store DuckDB database files.

mbostock: (But I think we should only check the file extension; we don’t need to check the additional magic.)

mbostock (approved):

PR #1068, opened 2024-03-15
fix describe{Tables,Columns}

mbostock:
These methods broke because `db.query` now returns an Arrow table rather than an Array, and hence has no _array_.map method. (We’ve deviated from the DatabaseClient specification now, but that’s a good thing because Arrow tables are much, much faster.)

````md
```js echo
const arrowTable = Arrow.tableFromArrays({col1: [1, 2, 3], col2: ["one", "two", "three"]});
const db = await DuckDBClient.of({arrowTable});
```

```js echo
db.describeTables()
```

```js echo
db.query("SELECT * FROM arrowTable")
```
````

### DIFF
--- a/src/client/stdlib/duckdb.js
+++ b/src/client/stdlib/duckdb.js
@@ -146,18 +146,19 @@ export class DuckDBClient {
   }
 
   async describeTables() {
-    const tables = await this.query("SHOW TABLES");
-    return tables.map(({name}) => ({name}));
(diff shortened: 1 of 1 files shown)


mbostock:
Demonstration of fix:

Screenshot 2024-03-15 at 8 52 59 AM

Fil (approved):

PR #1076, opened 2024-03-15
normalize file path; deterministic resolution

### LINKED ISSUE #1066: Paths for FileAttachment don't work using a leading slash
Was having a [Slack discussion on routing](https://observable-community.slack.com/archives/C06KPR8FCE4/p1709496586037579) and am noticing that referencing a data loader in a FileAttachment doesn't work with a leading slash.

For example, if I put the example notebooks under a `demo/` folder, then the following works when running a preview:

```js
const events = FileAttachment("/data/events.json").json();
```

but it fails for me once it is deployed to Observable. I get something like `Error: File not found: /data/events.json`

Our discussion on Slack ended on imports, which do work with a leading slash, but the FileAttachment path I think still has to be referenced by a relative path. It works locally on testing, but the deployed build fails.

**Overall**: was hoping that the paths to access data loader results with FileAttachment could allow for leading-slash paths, which would help improve organization with reference to folder organization of pages.

Thanks!

mbostock:
Fixes #1066. Fixes an interdependence in the tests, and removes nondeterminism from LoaderResolvers: instead of using the cached output if it exists, the calling code decides whether it wants to consider the data loader source file or the cached output file when working with a loader-generated file.


Fil:
lgtm but we're blocked by coverage

src/build.ts
```diff
@@ -167,7 +167,7 @@ export async function build(
     const hash = createHash("sha256").update(contents).digest("hex").slice(0, 8);
     const ext = extname(file);
     const alias = `/${join("_file", dirname(file), `${basename(file, ext)}.${hash}${ext}`)}`;
-    aliases.set(loaders.resolveFilePath(file), alias);
+    aliases.set(loaders.resolvePreviewFilePath(file), alias);
```
mbostock, line 170: This is confusing because we’re getting the preview path during build, but we’re not really using — we’re using it as the key to the `aliases` map, which we’ll then replace with the hashed value.

src/dataloader.ts
```diff
@@ -130,38 +130,55 @@ export class LoaderResolver {
...
+    const entry = getFileInfo(this.root, this.getBuildFilePath(name));
+    return entry && Math.floor(entry.mtimeMs);
+  }
+
+  resolvePreviewFilePath(path: string): string {
+    return `/${join("_file", path)}?sha=${this.getPreviewFileHash(path)}`;
+  }
+
+  resolveBuildFilePath(path: string): string {
+    return `/${join("_file", path)}?sha=${this.getBuildFileHash(path)}`;
```
mbostock, line 181: Some of these methods aren’t used; I’m going to prune them.

src/resolvers.ts
```diff
@@ -115,10 +116,10 @@ export async function getResolvers(
   }
 
   // Compute the content hash.
-  for (const f of assets) hash.update(loaders.getFileHash(resolvePath(path, f)));
-  for (const f of files) hash.update(loaders.getFileHash(resolvePath(path, f)));
+  for (const f of assets) hash.update(loaders.getPreviewFileHash(resolvePath(path, f)));
+  for (const f of files) hash.update(loaders.getPreviewFileHash(resolvePath(path, f)));
   for (const i of localImports) hash.update(getModuleHash(root, resolvePath(path, i)));
-  if (page.style && isPathImport(page.style)) hash.update(loaders.getFileHash(resolvePath(path, page.style)));
+  if (page.style && isPathImport(page.style)) hash.update(loaders.getPreviewFileHash(resolvePath(path, page.style)));
```
mbostock, line 122: This is also confusing since of course resolvers are used during build. But the content hash is only used during preview, so this likewise doesn’t affect the build.

test/build-test.ts
```diff
@@ -26,7 +26,7 @@ describe("build", () => {
     if (isEmpty(path)) continue;
     const only = name.startsWith("only.");
     const skip = name.startsWith("skip.");
-    const outname = only || skip ? name.slice(5) : name;
+    const outname = name.replace(/^only\.|skip\./, "");
```
mbostock, line 29: This lets you change the definition of `only` without a surprising rename of the test. 🙂

src/resolvers.ts
```diff
@@ -236,7 +237,7 @@ export async function getResolvers(
   }
 
   function resolveFile(specifier: string): string {
-    return relativePath(path, loaders.resolveFilePath(resolvePath(path, specifier)));
+    return relativePath(path, loaders.resolvePreviewFilePath(resolvePath(path, specifier)));
```
mbostock, line 240: This is also confusing, since this code runs during build… but the difference between a “preview file path” and a “build file path” is only the hash, and the hash ultimately isn’t used in build because file references are remapped through the `aliases` map. I need to come up with some better names here.

mbostock:
Merging to unblock test breakages in other branches. A posthumous review would still be appreciated. 🙏

Fil:
posthoc review: 👍 🚀

PR #1078, opened 2024-03-16
prql

### LINKED ISSUE #1075: PRQL
It compiles to SQL, so it should be a pretty easy lift, maybe?

https://prql-lang.org/

mbostock:
Fixes #1075.

Doesn’t work with interpolated expressions, e.g.,

```prql
from gaia
filter phot_g_mean_mag < ${min}
sort {phot_g_mean_mag}
take 10
```

I’m not sure how we’d implement that… we’d need some way to represent placeholder values (such as `?`) that are appropriately substituted during the transformation.

I’m also not sure if we also want to support a `prql` tagged template literal, since that would mean loading `prql-js` in the browser (importing it dynamically, presumably?).

### DIFF
--- a/docs/sql.md
+++ b/docs/sql.md
@@ -174,3 +174,25 @@ The `sql` tag is available by default in Markdown. You can also import it explic
 ```js echo
 import {sql} from "npm:@observablehq/duckdb";
 ```
+
+## PRQL
+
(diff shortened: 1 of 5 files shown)


srenatus:
Just so you're aware -- with the next release, importing prql-js might change? I've been pursuing prql in observable-framework myself [here](https://github.com/PRQL/prql/issues/4273), and that change -- switching the browser build to ESM module, hasn't been part of a release yet AFAICT.

mbostock:
That’s great to hear @srenatus. An ESM distribution would be welcome.

Issue #928, opened 2024-02-27
Custom implicit imports


mbostock:
The config file could enumerate a set of additions to the standard library, like we do with recommended libraries, where we say that `d3` is an alias for `import * as d3 from "npm:d3"`.

mbostock:
Another example…

```js
export default {
  library: {
    Highcharts: "npm:highcharts"
  }
};
```

We’ll have to figure out the distinction between a default import and a namespace import in this syntax, though, since Highcharts is imported as

```js
import Highcharts from "npm:highcharts";
```

whereas D3 is imported as

```js
import * as d3 from "npm:d3";
```

Maybe the default export is the default, and some additional syntax is required for a namespace import?

```js
export default {
  library: {
    Highcharts: {import: "default", from: "npm:highcharts"},
    d3: {import: "*", from: "npm:d3"}
  }
};
```

Or maybe it’s just JavaScript?

```js
export default {
  library: [
    "import Highcharts from 'npm:highcharts';",
    "import * as d3 from 'npm:d3';"
  ]
};
```

Or:

```js
export default {
  library: `
    import Highcharts from "npm:highcharts";
    import * as d3 from "npm:d3";
 `
};
```

I think we’d want to restrict what sort of JavaScript we allow here though… only statically-analyzable static import declarations. And I guess if there’s overlap between the project’s declarations and the default ones, the project’s declarations take precedence. But should there be a way to remove a default declaration? 🤔

PR #1085, opened 2024-03-16
custom builtins

mbostock:
Sketch at #928. The main thing missing is the equivalent of `getImplicitInputImports`, `getImplicitDownloads`, _etc._ Also, I’d rather have this live inside the config file, but I think it’s harder to express the builtins as code if we do that?

### DIFF
--- a/docs/_builtins.js
+++ b/docs/_builtins.js
@@ -1,0 +1,3 @@
+export default {
+  confetti: () => import("npm:canvas-confetti").then((confetti) => confetti.default)
+};
--- a/docs/config.md
+++ b/docs/config.md
@@ -1,5 +1,9 @@
 # Configuration
 
+```js
+confetti();
+```
+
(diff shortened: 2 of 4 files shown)


Fil:
Without trying to implement it (so I'm probably missing why it's difficult), I imagined you could say this in a config file:

```js
const confetti = () => import("npm:canvas-confetti").then((confetti) => confetti.default);

export default {
  builtins: {confetti},
  ...
}
```

Or, adding imports and downloads as keys:

```js
const confetti = {
  module: () => import("npm:canvas-confetti").then((confetti) => confetti.default),
  imports: … ,
  downloads: …
};

export default {
  builtins: {confetti},
  ...
}
```

mbostock:
Try it in our config and you’ll see why. The config is server-side code, so we don’t want to use it to define client-side code.

Issue #692, opened 2024-02-07
Resolve assets referenced in head, header, and footer


mbostock:
A common use case is to put a favicon in the **head** config option:

```html
<link rel="apple-touch-icon" href="/favicon.png">
<link rel="icon" type="image/png" href="/favicon.png" sizes="512x512">
```

There are two problems with this approach:

1. The `favicon.png` isn’t detected as a referenced file, so unless some other code references this, it won’t be included in the built site.
2. The `/favicon.png` link will work locally, but it won’t get rewritten to a relative link on build, and so will result in a broken link if the built site isn’t served at the root of a domain.

We’ve talked about having a “public” assets folder #169 but that only addresses problem (1), not problem (2). If we resolved reference assets in the head, header, and footer, then we’d solve both problems at once: we’d include the referenced files in the built output, and we’d rewrite the links to be relative. We already do this for the **scripts** option, so it seems reasonable to extend it to these other HTML fragments. (Though it’s extra tricky because these options can be overridden by the page.)

espinielli:
This is pretty needed...I resolved to embed the logo I wanted like in hello-world [config](https://github.com/observablehq/framework/blob/3f09fab3ea186531eddeda670b27233af1bc09ce/examples/hello-world/observablehq.config.ts#L39) but that doesn't scale: think of a nav-bar with icons...

mbostock:
We should also make `npm:` specifiers work, _e.g._

```html
<link rel="stylesheet" href="npm:katex/dist/katex.min.css">
```

sebsebmc:
This would potentially be useful for site-wide OpenGraph data such as thumbnails.

jlguenego:
When doing `<link rel="stylesheet" href="npm:katex/dist/katex.min.css">` it does not import at build time the fonts.

PR #1091, opened 2024-03-17
resolve npm: assets

mbostock:
Implements https://github.com/observablehq/framework/issues/692#issuecomment-1978007735. For example:

```html
<img src="npm:leaflet/dist/images/marker-icon-2x.png">
```

Not sure how useful this is, yet, but it feels like an easy extension of what we’re already doing.


Fil (approved):
nice! it also works with:

```js
head: `<link rel="icon" href="npm:leaflet/dist/images/marker-icon-2x.png">`,
```

PR #1097, opened 2024-03-18
DuckDBClient.sql

### LINKED ISSUE #1072: Define sql connectors to live data
```yaml
---
sql:
  mytable: ${a live expression}
---
```

Fil:
implement sql fenced code from DuckDBClient.sql(…)

following @mbostock’s suggestion https://github.com/observablehq/framework/issues/1072#issuecomment-2000253228

closes #1072


mbostock:
Do we still need this if we have #1098? I prefer using the **sql** front matter option if possible.

Fil:
Yes I think so? It would be useful if you wanted to work with an array of objects computed on the client and use them in sql fenced code blocks. It's possible without this, of course.

mbostock:
Ah, and the dynamic source case you mentioned (_e.g._, driven by input); **sql** front matter only works with static sources.

mbostock (approved):

PR #1098, opened 2024-03-18
accept bare URL definitions for sql tables

### LINKED ISSUE #1071: Support for sql remote URLs
```yaml
---
sql:
  gaia: https://uwdata.github.io/mosaic-datasets/data/gaia-5m.parquet
---
```

An alternative to consider / a complement would be to allow `FileAttachment("https://uwdata.github.io/mosaic-datasets/data/gaia-5m.parquet"[, options])`; the options could indicate the extension/mimeType.

Fil:
````yaml
---
sql:
  adresses: https://static.data.gouv.fr/resources/bureaux-de-vote-et-adresses-de-leurs-electeurs/20230626-135723/table-adresses-reu.parquet
  quakes: https://earthquake.usgs.gov/earthquakes/feed/v1.0/summary/all_day.csv
  gaia: https://uwdata.github.io/mosaic-datasets/data/gaia-5m.parquet
---

```sql echo
SELECT COUNT() FROM quakes;
```
````

etc.

closes #1071

I believe this is more a bugfix than a new feature (we previously returned `/https:/static.data.gouv.fr/resources/burea……….parquet`).

### CODE REVIEW

**Review by mbostock — COMMENTED** [review]
Not sure this is the right place to do this — normally we only call resolveFile on things that we know are files (unlike resolveImport, I guess). Look for calls to isAssetPath, which we should also use here. I’ll take a look and see if we can consolidate. Good find on the fix!

**Comment by Fil** [comment]
The caller is renderFiles
https://github.com/observablehq/framework/blob/efd8ba0eadf7ba19416fa6077260db1426b909a1/src/render.ts#L96

where we build the `registerTable` client code.

**Comment by mbostock** [comment]
Yeah, we don’t want to declare a `FileAttachment` with an external URL, I imagine. Interesting that it works though.

**Review by mbostock — APPROVED** [review]
Switched to using the [FROM-first syntax](https://duckdb.org/docs/sql/query_syntax/from.html), _e.g._,

```sql
CREATE TABLE 'quakes' AS FROM 'https://earthquake.usgs.gov/earthquakes/feed/v1.0/summary/all_day.csv';
```

docs/sql.md
```diff
@@ -16,6 +16,17 @@ sql:
...
+To load externally-hosted data, you can use a full URL:
+
+```yaml
+---
+sql:
+  quakes: https://earthquake.usgs.gov/earthquakes/feed/v1.0/summary/all_day.csv
+---
+```
+
+<div class="tip">For performance and reliability, we recommend using local files rather than loading data from external servers at runtime. If needded, you can use a <a href="./loaders">data loader</a> to take a snapshot of a remote data during build.</div>
```
Fil, line 28:
```suggestion
<div class="tip">For performance and reliability, we recommend using local files rather than loading data from external servers at runtime. If needed, you can use a <a href="./loaders">data loader</a> to take a snapshot of a remote data during build.</div>
```

PR #1108, opened 2024-03-19
hint to "deploy help" in the documentation

### LINKED ISSUE #1105: document --if-missing and --if-stale
https://observablehq.com/framework/deploying should describe these new options introduced in #1042

Fil:
closes #1105


mbostock (commented):
I appreciate the additional documentation, but this feels a little out of place here. I think it would be better to expand the `observable help deploy` output as I mentioned in Slack. We could also have a command reference page, perhaps, though I like the idea that the reference is built-in to the tool itself.

Fil:
Then maybe we just say "call `npm observable help deploy` for more options"?

mbostock:
It’s `npm run observable help deploy`, but yeah. 😅

mbostock (approved):
Made it a tip, and switched to:

```sh
npm run deploy -- --help
```

PR #1113, opened 2024-03-20
file.href

### LINKED ISSUE #954: Make FileAttachment.url synchronous
It’s async because it needed to be in notebooks, but with server-side rendering in Framework we always know the file attachment URL synchronously and it would be nice if you could say `file.url` instead of needing to `await file.url()`.

mbostock:
Fixes #954.


Fil:
Would it be useful to make this discoverable in the inspector?

For example by adding a getter to `FileAttachmentImpl`:

```js
  get href() {
    return this.href;
  }
```

Capture d’écran 2024-03-20 à 08 59 51

(it would also be nice to hide the deprecated url() method, but I don't know how to do that.)

Fil (approved):

mbostock:
If we want the `file.href` to be discoverable, we simply make it enumerable here:

```js
    Object.defineProperty(this, "href", {value: href, enumerable: true});
```

I chose not to do that because I think it’s extra noise, but let me take a look and see what I think.

PR #1115, opened 2024-03-20
fix mermaid dark mode

Fil:
This PR adopts [dark](<https://observablehq.com/framework/lib/generators#dark()>) to define the color of mermaid charts.

This fixes two issues:
* when the theme is not reactive to dark mode (when it has bee defined to be always light or always dark), the function could be wrong as it was based on user preference; it is now consistent with the actual colors of the page.
* when the user preference changed, the charts did not update; now they are regenerated automatically (when generated by the template literal; if you're using the function, you still have to make your expression depend explicitly on dark).

### DIFF
--- a/src/client/stdlib/mermaid.js
+++ b/src/client/stdlib/mermaid.js
@@ -1,8 +1,12 @@
 import mer from "npm:mermaid";
+import {Generators} from "observablehq:stdlib";
 
 let nextId = 0;
-const theme = matchMedia("(prefers-color-scheme: dark)").matches ? "dark" : "neutral";
-mer.initialize({startOnLoad: false, securityLevel: "loose", theme});
+
+(async () => {
+  for await (const dark of Generators.dark())
+    mer.initialize({startOnLoad: false, securityLevel: "loose", theme: dark ? "dark" : "neutral"});
+})();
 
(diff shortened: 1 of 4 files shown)


mbostock (changes requested):
Hmm, I don’t like the idea that the `mermaid` tagged template literal behaves differently (non-responsively) than the Mermaid code blocks. That means if you use the tagged template literal directly, either in Markdown or by importing it into a JavaScript component, it won’t be responsive.

I’m not sure this is really worth the trouble to fix. But I guess it would mean that the contents of the element returned by `mermaid` could change later (be rewritten) if the theme changes?

Maybe the better solution is to define our own theme that uses CSS variables instead of hard-coded colors, and then `mermaid` doesn’t need to be responsive at all. I think it’s possible to define custom themes that could reference our `--theme-*` variables.

Update: Mermaid [requires colors to be specified as hex](https://mermaid.js.org/config/theming.html#color-and-color-calculation), so CSS custom properties presumably aren’t supported. So, I think we probably do nothing here except perhaps fix the static check to look at the computed color-scheme instead of the preferred color scheme.

Fil:
superseded by #1122

PR #1116, opened 2024-03-20
add default favicon to templates

### LINKED ISSUE #674: observablehq-create should include a favicon
We're planning to serve a default one at /favicon.ico (https://github.com/observablehq/observablehq/issues/15747), but Visnu suggested that observablehq-create could also configure one for a new project.

See slack thread https://observablehq.slack.com/archives/C062YRED4TB/p1706990280011629

Fil:
TODO:

- [ ] design an asset (and format) to use as a default favicon for projects
- [ ] apply same code to the other template

closes #674

### DIFF
--- a/templates/empty/docs/favicon.svg
+++ b/templates/empty/docs/favicon.svg
@@ -1,0 +1,1 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" viewBox="0 0 16 16" fill="none" stroke="currentColor" stroke-width="2"><path d="M2 6C2 5.44772 2.44772 5 3 5H10C10.5523 5 11 5.44772 11 6V13C11 13.5523 10.5523 14 10 14H3C2.44772 14 2 13.5523 2 13V6Z M4 2.00004L12 2.00001C13.1046 2 14 2.89544 14 4.00001V12"></path></svg>
--- a/templates/empty/observablehq.config.js.tmpl
+++ b/templates/empty/observablehq.config.js.tmpl
@@ -16,6 +16,9 @@ export default {
   //   }
   // ],
 
+  // Links to add to the head of the page, e.g. for a favicon:
+  head: '<link rel="icon" href="favicon.svg" type="image/svg">',
+
   // Some additional configuration options and their defaults:
   // theme: "default", // try "light", "dark", "slate", etc.
   // header: "", // what to show in the header (HTML)


mbostock:
Safari doesn’t support [SVG favicons](https://caniuse.com/link-icon-svg); you have to use the [nonstandard mask-icon](https://www.leereamsnyder.com/favicons-in-2021). 😞

Fil:
Capture d’écran 2024-04-02 à 20 22 08
Capture d’écran 2024-04-02 à 20 22 01

asset minified down to 493 bytes

mbostock:
Can we not use our official favicon? I like it better. I don’t like that hairline white halo in dark mode.

https://static.observablehq.com/favicon-512.0667824687f99c942a02e06e2db1a060911da0bf3606671676a255b1cf97b4fe.png

mbostock:
Hmm, checking for favicon.png feels a little magical to me — it’s not hard to set the **head** option, is it? Especially if we now show you how to do it with the default template?

Fil:
I was thinking it could be useful for a beginner who does not open the config file and removes favicon.png from docs/. But the current code is working in a good way in that situation: the generated page points to the non-existing /favicon.png, a 404 is shown in the network inspector, all the information is there and nothing is badly broken.

mbostock (changes requested):
When you’ve replaced the image with the gray version let me know and I’ll take another look.

I am also tempted to call it `observable.png` rather than `favicon.png`, and to set `sizes="32x32"`, so as to encourage authors to use a differently-named file if they want to replace it, and thereby force discovery of the **head** option if you want to change the favicon. But not a requirement.

Fil:
Yes I'm only hesitant to use a 15kB 512x512 image, when we could minify a 32x32 image for 394 bytes. (It's not in the hot path, so not a big deal.)

mbostock:
You can use `gm convert` and `optipng` to convert that 512x512 image to 32x32? I’m not opposed to having a more minified favicon. I just prefer the gray one to the black & white one.

Fil:
Done. I also tested a build, just in case :)

templates/default/observablehq.config.js.tmpl
```diff
@@ -16,6 +16,9 @@ export default {
   //   }
   // ],
 
+  // Links to add to the head of the page, e.g. for a favicon:
```
mbostock, line 19:
```suggestion
  // Content to add to the head of the page, e.g. for a favicon:
```

Not limited to links here (e.g., you could have an analytics script or meta tags).

mbostock (approved):

PR #1120, opened 2024-03-20
summary tables

### LINKED ISSUE #23: Summary data table display

Fil:
just a prototype… there's still a lot to do

closes #23


Fil:
TODO:
- [ ] detect dates since they are numbers as of apache-arrow@16 (see #1375)

PR #1122, opened 2024-03-21
fix mermaid dark theme

mbostock:
Alternative to #1115. Still requires a reload if you change your preferred color scheme, but I think that’s okay.


Fil (approved):
It's way simpler, but it won't switch dynamically at dawn :)
(It fixes the main bug, so that's fine)

PR #1124, opened 2024-03-22
don’t preload async

mbostock:
A `<script type="module" async>` probably shouldn’t be preloaded — the [`async` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script#async) implies that it’s lower priority. For example, we don’t want to preload our first-party analytics script on the Framework documentation.


Fil (approved):
👍

PR #1125, opened 2024-03-22
show GitHub stars in header

mbostock:
Screenshot 2024-03-21 at 8 13 30 PM

### DIFF
--- a/observablehq.config.ts
+++ b/observablehq.config.ts
@@ -1,6 +1,12 @@
 import {formatPrefix} from "d3-format";
 
-const {stargazers_count} = await github("/repos/observablehq/framework").catch(() => ({stargazers_count: 4242}));
+let stargazers_count: number;
+try {
+  ({stargazers_count} = await github("/repos/observablehq/framework"));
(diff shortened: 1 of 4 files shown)


observablehq.config.ts
```diff
@@ -1,3 +1,7 @@
+import {formatPrefix} from "d3-format";
+
+const {stargazers_count} = await github("/repos/observablehq/framework").catch(() => ({stargazers_count: 4242}));
```
Fil, line 3: nit: I'd default to null or undefined if the github API call fails. This might require an additional test below to avoid showing 0.0k/NaNk stars (although it could be ok too to show NaN if we can't get the information).

mbostock: I was only thinking of local development. In CI I’d want the deploy to fail if we’re not able to talk to GitHub, so that we always show the correct star count. I’ve changed the code to rethrow in CI.

Fil (approved):

mbostock:
We may want some caching if we load the config lazily on each request… #695

PR #1129, opened 2024-03-23
fix config caching

mbostock:
Previously `normalizeConfig` would return a new `Config` instance given the same spec. We want it to return the _same_ instance given the same spec, so that downstream code (such as search) can use the `Config` as a key in a `WeakMap`. This applies two additional optimizations: `readPages` is now cached, keyed on the file contents within the directory; and `parseMarkdownMetadata` is used instead of `parseMarkdown` within `readPages`, avoiding unnecessary work to compute the default page list.


docs/display-race.md
mbostock, line 1: I noticed these two test pages were inadvertently committed.

test/build-test.ts
```diff
@@ -42,7 +42,7 @@ describe("build", () => {
 
       await rm(actualDir, {recursive: true, force: true});
       if (generate) console.warn(`! generating ${expectedDir}`);
-      const config = Object.assign(await readConfig(undefined, path), {output: outputDir});
+      const config = {...await readConfig(undefined, path), output: outputDir};
```
mbostock, line 45: This change is to avoid mutating the `Config` instance: the same config is used by other tests.

test/config-test.ts
```diff
@@ -4,7 +4,7 @@ import {normalizeConfig as config, mergeToc, readConfig, setCurrentDate} from ".
 import {LoaderResolver} from "../src/dataloader.js";
 
 describe("readConfig(undefined, root)", () => {
-  before(() => setCurrentDate(new Date("2024-01-11T01:02:03")));
+  before(() => setCurrentDate(new Date("2024-01-10T16:00:00")));
```
mbostock, line 7: This change is to match the build tests — since the config is now cached, we want the footer to have the same data, so the behavior of the tests is the same independent of order.

src/search.ts
```diff
@@ -84,7 +85,7 @@ export async function searchIndex(config: Config, effects = defaultEffects): Pro
     )
   );
 
-  indexCache.set(config, {json, freshUntil: +new Date() + reindexingDelay});
+  indexCache.set(pages, {json, freshUntil: Date.now() + reindexDelay});
```
mbostock, line 88: The cache is keyed by the `pages` instead of the `config` so that if the set of pages change, the search index is recomputed.

src/files.ts
```diff
@@ -52,6 +52,11 @@ export function* visitMarkdownFiles(root: string): Generator<string> {
 export function* visitFiles(root: string): Generator<string> {
   const visited = new Set<number>();
   const queue: string[] = [(root = normalize(root))];
+  try {
+    visited.add(statSync(join(root, ".observablehq")).ino);
+  } catch {
+    // ignore the .observablehq directory, if it exists
```
mbostock, line 58: I noticed that if you have a `.md` file in your `docs/.observablehq/cache`, it was being picked up by the default pages. We don’t want that, so let’s just ignore everything in `.observablehq` under the root.

src/config.ts
```diff
@@ -88,18 +89,29 @@ export async function readDefaultConfig(root?: string): Promise<Config> {
...
 function readPages(root: string, md: MarkdownIt): Page[] {
-  const pages: Page[] = [];
+  const files: {file: string; source: string}[] = [];
+  const hash = createHash("sha256");
   for (const file of visitMarkdownFiles(root)) {
     if (file === "index.md" || file === "404.md") continue;
     const source = readFileSync(join(root, file), "utf8");
-    const parsed = parseMarkdown(source, {path: file, md});
+    files.push({file, source});
+    hash.update(file).update(source);
```
mbostock, line 101: I also tried hashing the mtime of the file instead of the source, which is slightly faster in the cached cache. But I don’t think it’s fast enough that it matters, and this felt simpler.

src/config.ts
```diff
@@ -88,18 +89,29 @@ export async function readDefaultConfig(root?: string): Promise<Config> {
   return normalizeConfig(await importConfig(tsPath), root);
 }
 
+let cachedPages: {key: string; pages: Page[]} | null = null;
```
mbostock, line 92: I’m only store a single value (the most recent value) in the cache for `readPages`, just to keep things simple. I don’t expect that there would be more than one set of pages needed simultaneously.

src/config.ts
```diff
@@ -109,7 +121,15 @@ export function setCurrentDate(date = new Date()): void {
...
 
+// The config is used as a cache key for other operations; for example the pages
+// are used as a cache key for search indexing and the previous & next links in
+// the footer. When given the same spec (because import returned the same
+// module), we want to return the same Config instance.
+const configCache = new WeakMap<any, Config>();
+
 export function normalizeConfig(spec: any = {}, defaultRoot = "docs"): Config {
+  const cachedConfig = configCache.get(spec);
+  if (cachedConfig) return cachedConfig;
```
Fil, line 132: Do you prefer this pattern over:
```
  if (configCache.has(spec)) return configCache.get(spec);
```
For performance maybe?

src/search.ts
```diff
@@ -21,15 +21,16 @@ const defaultEffects: SearchIndexEffects = {logger: console};
 const indexOptions = {
   fields: ["title", "text", "keywords"],
   storeFields: ["title"],
-  processTerm(term) {
-    return term.match(/\p{N}/gu)?.length > 6 ? null : term.slice(0, 15).toLowerCase(); // fields to return with search results
+  processTerm(term: string) {
+    return (term.match(/\p{N}/gu)?.length ?? 0) > 6 ? null : term.slice(0, 15).toLowerCase(); // fields to return with search results
```
Fil: I'm moving this comment to line 22 (the issue predates this PR, I just happened to see it here).

Fil (approved):

src/config.ts
mbostock: For TypeScript.

Previously you had declared the `WeakMap` without types, so keys and values were type `any` by default. In this case TypeScript considers the return value of `map.get` as `any`, too, and hence doesn’t require that it returns a defined value and allows `map.get(key).property` even though it is unsafe.

When we declare the cache as `WeakMap<Config["pages"], {json: string; freshUntil: number}>`, then TypeScript will enforce type checking because `map.get(key)` now returns `{json: string; freshUntil: number} | undefined`. So if we don’t use the pattern above, we get an `Object is possibly 'undefined'` type error:

Screenshot 2024-03-23 at 9 07 57 AM

TypeScript isn’t smart enough to know that if `map.has(key)` returns true, then a subsequent `map.get(key)` with the same `key` returns a defined value. (There’s no way to declare such a relationship in an interface, at least not yet.) So as a type-safe alternative, we call `map.get(key)` first, and then test the returned value which can’t change, and hence TypeScript knows that the value is defined.

PR #1131, opened 2024-03-23
log a confirmation message after a successful logout

Fil:


mbostock (commented):
Looks like the tests are strict about what gets logged, but otherwise LGTM!

mbostock (approved):

PR #1132, opened 2024-03-23
fix hello race condition

### LINKED ISSUE #563: Vi and emacs don't properly live update in preview
Ambassador Martien [reports](https://observable-community.slack.com/archives/C06E30AF761/p1705677466167639?thread_ts=1705677466.167639&cid=C06E30AF761) that when saving files using vim and emacs in their default configurations, the CLI does not properly live update and instead shows an error about "file no longer exists".

> Question: (I’m new to yarn):
> - when I use vi or emacs to edit, say, index.md , yarn says “file no longer exists: /index.md”, and the page in Safari does not update;
> - when I edit a copy of index.md and use that to overwrite index.md, yarn sees the update and updates the browser page.
>
> What editor should I use?

> I think I found it. In vi , :set nowritebackup) :
> - inode stays the same;
> - only modify and change timestamps change. Access and birth remain the same.

mbostock:
The lazy config #695 introduced a race condition where we could drop a _hello_↑ message that is sent immediately upon connection because we were waiting for the config to load. Now we register the _message_ event listener (and other listeners) synchronously on connect, moving the await to within the listener. Fixes #563.

### DIFF
--- a/src/preview.ts
+++ b/src/preview.ts
@@ -223,9 +223,9 @@ export class PreviewServer {
     }
   };
 
-  _handleConnection = async (socket: WebSocket, req: IncomingMessage) => {
+  _handleConnection = (socket: WebSocket, req: IncomingMessage) => {
     if (req.url === "/_observablehq") {
-      handleWatch(socket, req, await this._readConfig());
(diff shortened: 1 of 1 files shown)


Fil (approved):
works for real!

PR #1144, opened 2024-03-25
more precise esbuild target

### LINKED ISSUE #1140: esbuild’s keepNames uses static initializer blocks, which requires Safari 16.4+
Safari 16.4 was released March 2023 which is a bit bleeding edge, and I’d like to support older versions of Safari if it’s easy.

Ref. https://twitter.com/tbdr/status/1772168965348843542

mbostock:
Fixes #1140.

### DIFF
--- a/src/rollup.ts
+++ b/src/rollup.ts
@@ -66,7 +66,7 @@ export async function rollupClient(
       nodeResolve({resolveOnly: BUNDLED_MODULES}),
       importResolve(input, root, path),
       esbuild({
-        target: "es2022",
+        target: ["es2022", "chrome96", "firefox96", "safari16", "node18"],
         exclude: [], // don’t exclude node_modules
         keepNames,
         minify,


src/rollup.ts
```diff
@@ -66,7 +66,7 @@ export async function rollupClient(
       nodeResolve({resolveOnly: BUNDLED_MODULES}),
       importResolve(input, root, path),
       esbuild({
-        target: "es2022",
+        target: ["es2022", "chrome96", "firefox96", "safari16", "node18"],
```
mbostock, line 69: These are the latest browsers as of January 2022, with the exception of Safari which shipped September 2022.

Fil (approved):

Fil:
Are you seeing a difference in the generated files?

mbostock:
Yes.

Screenshot 2024-03-25 at 4 22 33 PM

PR #1145, opened 2024-03-26
initialize cache on create

### LINKED ISSUE #1064: npm init should probably run the initial build
Since we now self-host libraries from npm, it probably be good to pre-populate the npm cache by running `npm run build` during the `npm init @observablehq` step. That way the first page load is instant (rather than downloading a bunch of stuff from jsDelivr).

mbostock:
Fixes #1064. By running build during create, the initial loads of Framework in the preview server are instant because the npm and data loader cache are already populated. I think this is a better user experience — you’re expecting to wait during install anyway, and it means your first experience seeing pages load in Framework is representative of the (instant) experience of the cached and built project.


Fil (approved):
Much nicer experience! The 1000ms delay is to give us time to ctrl-c?

PR #1146, opened 2024-03-26
1.5.0

mbostock:


mbostock:
We’re getting 500 errors from weather.gov in CI which I can’t reproduce locally, but I figure it’s fine to cache the weather forecast for the getting started tutorial anyway and remove the external dependency.

mythmon (approved):

PR #1147, opened 2024-03-26
reload on config change

mbostock:
Followup to #695 which actively watches the config file and tells the client to *reload*↓ if the config file changes. Might be overkill, and a bit hard to test in this repo since we use `tsx watch`, too. What do you think? Seems fine to me…


Fil (approved):
Testing with `yarn tsx src/bin/observable.ts preview`: it's working great!

mbostock:
Oops, I was supposed to merge this before 1.5.1. But we haven’t created the tag yet, so it’s still fine.

PR #1155, opened 2024-03-27
fix infinite normalize loop

### LINKED ISSUE #783: Things break when a file name contains emoji

### LINKED ISSUE #1151: Path names containing spaces cause an infinite redirect loop
I think the root cause is due to the fact that an escaped path name containing spaces will never be equal to the original path name. 
 https://github.com/observablehq/framework/blob/0186815b016bc9ba38619e86cff6d68ea1dde18a/src/preview.ts#L174-L179

mbostock:
Fixes #1151. Fixes #783.

Screenshot 2024-03-27 at 9 22 24 AM

Screenshot 2024-03-27 at 9 21 50 AM

### DIFF
--- a/src/preview.ts
+++ b/src/preview.ts
@@ -172,7 +172,7 @@ export class PreviewServer {
 
         // Normalize the pathname (e.g., dropping ".html").
         const normalizedPathname = config.md.normalizeLink(pathname);
-        if (pathname !== normalizedPathname) {
+        if (url.pathname !== normalizedPathname) {
           res.writeHead(302, {Location: normalizedPathname + url.search});
           res.end();
           return;


Fil (approved):
I tried to break it but failed.

ź̶̢̳̗̮̖̲̲̼̬͆̎̔̓̊̔͋͒͋̔͠ͅa̴̳̙̭̅̀̓̓̇̒̋̔̑͛͒̐̏͝ļ̶̨̩͇̖͉̦͚̱͇̬͕́́̋̉͐̇̊͌́̒̓̏̚͝ġ̵̢̠̣̦̖̖͉̫̩͚̦̫̗͖̝̳͔̎́̇̈́̔͂͗̀̔̀̋̆̕ȯ̵̧̢̝̰̝͈̮͕̥͓̗̙̋͋̆͛̚ͅͅ

PR #1156, opened 2024-03-27
node imports

### LINKED ISSUE #360: Import from node_modules with bare module specifiers
We currently support `npm:` protocol imports for importing directly from npm (via the jsDelivr CDN). But we could also support importing from your local `node_modules` with bare module specifiers, using rollup to build an ES module as needed.

mbostock:
Fixes #360.


mbostock:
Probably the big thing missing here is that we need to set the “browser” condition instead of the “node” condition for [conditional exports](https://nodejs.org/api/packages.html#conditional-exports). Unfortunately I don’t think Node exposes any functionality for resolving a require with custom conditions; `require.resolve` always uses the “node” and “import” conditions, which means we’ll then attempt to bundle the wrong entry for the browser. Probably what we need to do is use [@rollup/plugin-node-resolve](https://www.npmjs.com/package/@rollup/plugin-node-resolve), or at least inspect its implementation to learn how to do it.

Fil:
Another case where things don't work:
`yarn add --dev @duckdb/duckdb-wasm@1.28.1-dev173.0`

````
```js echo
const duck = import ("@duckdb/duckdb-wasm");
```
````

results in
`RuntimeError: Failed to fetch dynamically imported module: http://127.0.0.1:3000/_node/@duckdb/duckdb-wasm@1.28.1-dev173.0/dist/duckdb-node.cjs`

the **.cjs** extension is a first blocker; the console says:
Failed to load module script: Expected a JavaScript module script but the server responded with a **MIME type of "application/octet-stream"**. Strict MIME type checking is enforced for module scripts per HTML spec.

**Comment by mbostock** [comment]
Yup, that’s the node entry point rather than the browser entry point.

**Comment by mbostock** [comment]
Okay, now using @rollup/plugin-node-resolve! 🚀

**Comment by mbostock** [comment]
Argh, it doesn’t work on windows for some reason. ☹️

**Comment by mbostock** [comment]
Okay, it was my mistake. 😌 Fixed now!

**Comment by Fil** [comment]
yoohoo!

**Comment by Fil** [comment]
I can't find how to load the DuckDB wasm this way.

```js
import * as duckdb from "@duckdb/duckdb-wasm";  // works
import duckdb_wasm from '@duckdb/duckdb-wasm/dist/duckdb-mvp.wasm'; // doesn't work
```

mbostock:
You can’t `import` a `.wasm`; you use `import.meta.resolve` instead.

Fil:
Yes I was just getting to the point where this seems to work:

````
```js echo
import * as duckdb from "@duckdb/duckdb-wasm";
const MANUAL_BUNDLES = {
    mvp: {
        mainModule: import.meta.resolve('@duckdb/duckdb-wasm/dist/duckdb-mvp.wasm'),
        mainWorker: import.meta.resolve('@duckdb/duckdb-wasm/dist/duckdb-browser-mvp.worker.js'),
    },
    eh: {
        mainModule: import.meta.resolve('@duckdb/duckdb-wasm/dist/duckdb-eh.wasm'),
        mainWorker: import.meta.resolve('@duckdb/duckdb-wasm/dist/duckdb-browser-eh.worker.js'),    },
};
// Select a bundle based on browser checks
const bundle = await duckdb.selectBundle(MANUAL_BUNDLES);
// Instantiate the asynchronous version of DuckDB-Wasm
const worker = new Worker(bundle.mainWorker);
const logger = new duckdb.ConsoleLogger();
const db = new duckdb.AsyncDuckDB(logger, worker);
await db.instantiate(bundle.mainModule, bundle.pthreadWorker);
```
````

I'll approve the PR and continue tinkering with DuckDB :)

**Review by Fil — APPROVED** [review]

**Comment by mbostock** [comment]
Seems to work! I tested this:

````md
```js echo
const connection = await db.connect();
try {
  display(await connection.query(`SELECT 1 + 2`));
} finally {
  await connection.close();
}
```
````

**Comment by mbostock** [comment]
Confirmed that this works with Yarn’s **resolutions**, so you can for example tell @duckdb/duckdb-wasm to use the latest version of apache-arrow like so in `package.json`:

```json
  "resolutions": {
    "@duckdb/duckdb-wasm/apache-arrow": "15.0.2"
  }
```

src/node.ts
```diff
@@ -0,0 +1,115 @@
...
+  const {name, path = "."} = parseNpmSpecifier(spec);
+  const require = createRequire(pathToFileURL(op.join(packageRoot, "/")));
+  const pathResolution = require.resolve(spec);
+  let packageResolution = pathResolution;
+  do {
+    const p = op.dirname(packageResolution);
+    if (p === packageResolution) throw new Error(`unable to resolve package.json: ${spec}`);
+    packageResolution = p;
+  } while (!existsSync(op.join(packageResolution, "package.json")));
+  const {version} = JSON.parse(await readFile(op.join(packageResolution, "package.json"), "utf-8"));
```
mbostock, line 34: For posterity, here is an alternative approach to finding the `package.json` that I discovered:

https://github.com/nodejs/node/issues/33460#issuecomment-919184789

I guess there are some cases of a “nested” `package.json` that this doesn’t handle, and so their approach is to trim the relative path to the entry point from the fully-resolved path. I think our approach is fine for now, though. There’s also this:

https://github.com/sindresorhus/package-up

mbostock: I’ve switched this branch to use [pkg-dir](https://github.com/sindresorhus/pkg-dir).

PR #1160, opened 2024-03-28
browser entry point

Fil:
Re:  https://github.com/observablehq/framework/pull/1156#issuecomment-2024391100.

Instead of taking at face value what pathResolution returns, we find package.json then see if the "browser" ~~condition~~  entry point is present and update pathResolution as a consequence. The [spec](https://github.com/defunctzombie/package-browser-field-spec) indicates that there are two possibilities to cover:
- a string (found in _e.g._ "supports-color@7.2.0"), 
- a replacement map (found in _e.g._ "asap@2.0.6").

My tests use two modules that are already here by "chance", and will break if any of these modules gets updated (for example supports-color@9 does not use the browser field anymore), but we can fix that when it happens. (Alternatively, we could add these precise versions in devDependencies just for these tests?)

I checked what [@rollup/plugin-node-resolve](https://www.npmjs.com/package/@rollup/plugin-node-resolve) plugin does: it looks at the entry with the ~~condition~~ key in [resolvePackageTarget](https://github.com/rollup/plugins/blob/2a19079892f0bef53b557da965339cdef0a13a93/packages/node-resolve/src/package/resolvePackageTarget.ts#L145). The chain of calls: < [resolvePackageImportsExports](https://github.com/rollup/plugins/blob/master/packages/node-resolve/src/package/resolvePackageImportsExports.ts#L96) < [resolvePackageImports](https://github.com/rollup/plugins/blob/master/packages/node-resolve/src/package/resolvePackageImports.ts#L14) < [resolveImportSpecifiers](https://github.com/rollup/plugins/blob/2a19079892f0bef53b557da965339cdef0a13a93/packages/node-resolve/src/resolveImportSpecifiers.js#L37) < [resolveImportSpecifiers](https://github.com/rollup/plugins/blob/2a19079892f0bef53b557da965339cdef0a13a93/packages/node-resolve/src/resolveImportSpecifiers.js#L108)

### DIFF
--- a/src/node.ts
+++ b/src/node.ts
@@ -23,14 +23,25 @@ const bundlePromises = new Map<string, Promise<void>>();
 async function resolveNodeImportInternal(root: string, packageRoot: string, spec: string): Promise<string> {
   const specifier = parseNpmSpecifier(spec);
   const require = createRequire(pathToFileURL(op.join(packageRoot, "/")));
-  const pathResolution = require.resolve(spec);
+  let pathResolution = require.resolve(spec);
   let packageResolution = pathResolution;
   do {
     const p = op.dirname(packageResolution);
     if (p === packageResolution) throw new Error(`unable to resolve package.json: ${spec}`);
     packageResolution = p;
   } while (!existsSync(op.join(packageResolution, "package.json")));
-  const {version} = JSON.parse(await readFile(op.join(packageResolution, "package.json"), "utf-8"));
+  const {version, browser} = JSON.parse(await readFile(op.join(packageResolution, "package.json"), "utf-8"));
+  if (browser) {
+    if (typeof browser === "string") pathResolution = op.join(packageResolution, browser);
+    else if (typeof browser === "object") {
+      for (const [key, value] of Object.entries(browser)) {
+        if (typeof value === "string" && op.join(packageResolution, key) === pathResolution) {
+          pathResolution = op.join(packageResolution, value);
+          break;
+        }
+      }
+    }
+  }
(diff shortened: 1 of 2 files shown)


mbostock (changes requested):
This is using the `browser` entry point (or “main field”) which is different from the `browser` condition within the `exports` object, which is what I was referring to. The latter is the modern standard that is the top priority to support. I suppose we should also support the `browser` field too as a fallback (and maybe `jsdelivr` and `unpkg` fields, like d3-require does?). It sure is complicated! At least there are modern standards for this, but it’ll be a bit of a mess to try to make everything work.

mbostock:
Also I wanted to see if there is a way to use Rollup’s node-resolve plugin to figure some of this stuff out for us. I think it might be possible, but I’m not sure. Anyway I appreciate your help cracking this nut!

Fil:
yup, I also feel this is something that should be left to rollup.

mbostock:
Superseded by 50a3e7d11a680fcb2fc24cd8b8de5a1f0809e177.

mbostock:
(I’ll work on porting the tests now.)

mbostock:
Hmm, I don’t have a good way of porting the tests because we no longer return the resolved path to the entry point — we just pass the entry name to rollup and let it figure it out. So the default entry name is `.` which we rename to `index.js` (because we need the `.js` extension, and I’m assuming no one would need both `.` and `index.js` to refer to different things). We’d have to test the transpiled contents, I think. Maybe we need to hand-craft some packages to test.

Issue #1036, opened 2024-03-11
Expose front matter (meta data) to the page?


Fil:
Exposing meta data to the page could help in many scenarios. For a concrete example, suppose that I want to show an author's name in a page.

If I defined it in the front matter as

```md
---
author: Fil
date: 2024-03-11
keywords: [ spam, ham ]
---
```

and could somehow reuse these in the page with something like `${meta.author}`, it would help a lot when organizing a complex website.

(Initially a comment in #178)

Issue #180, opened 2023-11-15
An origin option for features that need absolute URLs


Fil:
For anything that needs absolute URLs (#168, #178, #179).

~~The 404.html page (#174) should also somehow be aware of the server's root (does not need the **host** option, but the **base**) if we want to use the same file to serve 404s both at paths containing any sub/paths/ from the home page.~~(or maybe not, see https://github.com/observablehq/cli/pull/194#issuecomment-1817414701).

See also discussion under #42.

mbostock:
Does the existing **base** option work for this?

Fil:
We could choose to make it work, for example when this option is specified as a URL it could be applied to all pages, not only the 404. But we'd need to change the relative paths (as they would now be relative to the base, not to the page). It feels like a different concern (or concept), in that what we need here is the URL of the root of the site. "host", as this was titled, is not enough.

mbostock:
I didn’t mean to add a `<base>` everywhere. I meant to use the **base** value to compute an absolute path as needed. But not enough if you need the origin (host) too I guess.

Fil:
Oh I see… yes I think you're correct: if the base is an absolute URL, it would have that slightly new meaning (URL of the root of the site) and still be compatible with how we use it for 404.html.

mbostock:
> it would have that slightly new meaning (URL of the root of the site)

I thought that’s what it meant. 😅 The docs say “the base path when serving the site.” Or are you saying that “base” would then have “https://…” in it? I didn’t mean that. I just meant enough to compute the absolute path. If we need the absolute URL including the origin then we should add a new config property such as **origin**.

Fil:
ok, so the formula would be `root url = origin + base`, with origin = 'https://user.observablehq.cloud' and base = '/project-name/'? 

(this is what this issue was about originally [but not explicit enough, and the option was called **host**]).

Issue #179, opened 2023-11-15
Generate link rel=canonical


Fil:
With the **host** and **origin** options provided (#180, #42), we can derive the page’s canonical URL (which is a good thing to have on static websites that might be served under different circumstances). Maybe a good idea in that case to automatically add the corresponding link[rel=canonical].

To determine this URL we also need to know the page’s path; this can be done if **head** is a function that receives the path as an argument (see #1161).

Issue #178, opened 2023-11-15
Support open graph


Fil:
Design and implement a system (template? hook? options in frontmatter?) that allows the developer to add proper tags so that their page has the relevant information to create a "card" or "summary" on social platforms. Usually:

- the title
- an og:image (typically 1200 &times; 630)
- a description
- the canonical url (#179)

This needs the **origin** and **base** options (#180).

Fil:
Exposing meta data to the page could help in more scenarios. For a concrete example, suppose that I want to show an author's name in a page.

If I defined it in the front matter as

```md
---
author: Fil
---
```

and could somehow reuse it in the page with something like `${meta.author}`, it would help a lot when organizing a complex website.

pettiross:
OG images would also let us show thumbnails for projects and pages on the platform.

Fil:
related: #168

pettiross:
+1 to using the front matter for this. We could expand on the front matter + `observable.config.ts` combination we have today so project-level defaults for OG metalabels could be set once and overridden on specific pages.

mbostock:
Is this already done with the **head** option (and the ability to reference local assets in the head)?

Fil:
Not quite yet. I'd like to explore a possibility for the **head** option (and header/footer) to be specified as a function, that would receive as arguments some meta data about the project config (including the **origin** option #180), and the current page (front matter, title, and possibly contents/files, if the user wants to insert an excerpt / summary or pick the first image from the contents) #1036.

mbostock:
I’m still not sure what this needs. Why not specify the **head** front matter option? Do we just want shorthand so you can specify a **description** front matter option (and maybe **image**) instead that gets promoted to the corresponding `meta` elements in the head?

Fil:
The way I imagine this, the configuration **head** option, if specified as a function (#1255) would receive meta data, from the page's front matter (or first H1 if the title is not defined as front matter), and would build the og "card" from these values — it seems like a common convenience in CMS.

I did not even consider typing everything explicitly, although it (almost) works:

```
---
head:
  <meta property="og:title" content="Page title"/>
  <meta property="og:description" content="This is a nice page." />
  <meta property="og:image" content="heroimage.png"/>
  <meta property="og:image:width" content="1200"/>
  <meta property="og:image:height" content="630"/>
---
```

[...]

PR #1161, opened 2024-03-28
support open graph (exploration)

Fil:
This branch (which is probably not going to be a PR) explores what's needed to support open graph (#178) and rel-canonical (#179). It raises a few questions:

- [x] support an **origin** option (#180)
- [x] allow the user to add their own custom values in front matter (part of #1036)
- [ ] `["meta[property='og:image']", "content"]` must be an _absolute_ URL, which means we need a proper way either to apply the origin+base transform after the HTML has been produced, or conversely that we might want to process a file with an absolute url that corresponds to the site as fileattachments? (There is also `og:image:secure_url`)
- [ ] Should #1083 also have treated a head/header/footer value passed as front-matter? (see #1255).

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -174,6 +174,10 @@ export default {
 
 The base path when serving the site. Currently this only affects the custom 404 page, if any.
 
+## origin
+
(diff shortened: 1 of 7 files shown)


Fil:
see [discussion](https://github.com/observablehq/framework/discussions/1199#discussioncomment-10616803)

PR #1165, opened 2024-03-29
implicit /+esm if no extension; fix /+esm conflict

### LINKED ISSUE #1163: Adding /+esm to an import can crash the preview server
The problem is that you can’t have a file and a directory with the same name. So if you start with this:

```js
import confetti from "npm:canvas-confetti@1/dist/confetti.module.mjs";
```

And then you change it to:

```js
import confetti from "npm:canvas-confetti@1/dist/confetti.module.mjs/+esm";
```

You’ll get:

```
Error: EEXIST: file already exists, mkdir 'docs/.observablehq/cache/_npm/canvas-confetti@1.9.2/dist/confetti.module.mjs'
```

Maybe we could change our rewriting to handle this more gracefully? One possibility is the `_esm` moves to the front instead of the end:

```
/_npm/canvas-confetti@1/_esm/dist/confetti.module.mjs
```

And maybe the default case remains unchanged (`/_esm.js`).

mbostock:
In the process of documenting imports, I noticed we don’t handle specifying named entry points well. This fixes importing a named entry point such as:

```js
import mime from "npm:mime/lite";
```

Without this, the import would fail because the `.js` extension is needed to set the MIME type correctly, and also because jsDelivr doesn’t support loading a named entry point without `/+esm` ([e.g.](https://cdn.jsdelivr.net/npm/mime@4.0.1/lite)). You could workaround it like so:

```js
import mime from "npm:mime/lite/+esm";
```

But I figure that it would be better if it “just worked” rather than knowing you need to add `/+esm`, just like the default `npm:mime` without a path works.

I also thought about implicitly adding `/+esm` for paths with JavaScript extensions, such as:

```js
import "npm:katex/contrib/mhchem/mhchem.js";
```

However, since the import could already be valid as-is, I figure it’s better to opt-in to the `/+esm` explicitly. And there are also cases where `/+esm` is broken, such as `echarts`, so we’d have to invent a way to opt-out if we add `/+esm` implicitly.

```js
import "npm:echarts/dist/echarts.esm.min.js";
```

~~In the latest commit, I’ve also tacked on a fix #1163, which is to change how we’re rewriting `/+esm`. Instead of replacing it with `/_esm.js`, we now move it to a leading `/_esm`. This is the same in the default case _e.g._ `npm:d3` resolves to the jsDelivr path `/npm/d3/+esm` and the local path `/_npm/d3/_esm.js`, but in a more complicated case such as `npm:d3@7.8.5/dist/d3.js` which is `/npm/d3@7.8.5/dist/d3.js/+esm`, it is now `/_npm/d3@7.8.5/_esm/dist/d3.js` which avoids the problem of `/_npm/d3@7.8.5/dist/d3.js` potentially needing to be both a file and a directory.~~

I’ve now changed the fix for #1163 to be smaller, which is instead of always replace `/+esm` with `/_esm.js`, in every case except for the top-level `/+esm`, it gets replaced with `._esm.js` instead. The exception for the top-level path is to produce `/_npm/d3@7.8.5/_esm.js` instead of `/_npm/d3@7.8.5._esm.js`. In some was the latter would be preferred and more readable, but I think it’s better for everything to live under `/_npm/d3@7.8.5/` so that you can more easily clear the cache for a specific package. Perhaps a further improvement would change it to `/_npm/d3/7.8.5._esm.js`? At any rate I think that’s an aesthetic decision that I would prefer to defer to future work.

### DIFF
--- a/src/npm.ts
+++ b/src/npm.ts
@@ -1,6 +1,6 @@
 import {existsSync} from "node:fs";
 import {mkdir, readFile, readdir, writeFile} from "node:fs/promises";
-import {dirname, join} from "node:path/posix";
+import {dirname, extname, join} from "node:path/posix";
 import type {CallExpression} from "acorn";
 import {simple} from "acorn-walk";
 import {rsort, satisfies} from "semver";
@@ -249,7 +249,13 @@ export async function resolveNpmImport(root: string, specifier: string): Promise
       ? "dist/echarts.esm.min.js"
       : "+esm"
   } = parseNpmSpecifier(specifier);
-  return `/_npm/${name}@${await resolveNpmVersion(root, {name, range})}/${path.replace(/\+esm$/, "_esm.js")}`;
+  return `/_npm/${name}@${await resolveNpmVersion(root, {name, range})}/${
+    extname(path)
+      ? path // preserve the existing extension
+      : path.endsWith("/+esm") || path === "+esm" || !path
+      ? `${path.slice(0, -"+esm".length)}_esm.js` // remove +esm; add _esm.js
+      : `${path}/_esm.js` // add /_esm.js
+  }`;
 }
 
 /**
--- a/src/npm.ts
+++ b/src/npm.ts
@@ -269,7 +269,18 @@ export async function resolveNpmImports(root: string, path: string): Promise<Imp
  */
 export function extractNpmSpecifier(path: string): string {
   if (!path.startsWith("/_npm/")) throw new Error(`invalid npm path: ${path}`);
-  return path.replace(/^\/_npm\//, "").replace(/\/_esm\.js$/, "/+esm");
+  const parts = path.split("/");
+  parts.splice(0, 2); // delete /_npm
+  const i = parts[0].startsWith("@") ? 2 : 1;
+  if (parts[i] === "_esm" || (i === parts.length - 1 && parts[i] === "_esm.js")) {
+    parts.splice(i, 1); // delete _esm
+    parts.push("+esm"); // append +esm
+  }
+  // TODO If we added an extension to the path, we should remove it here, but
+  // we can’t distinguish between the case where the extension was added by us
+  // (as in `npm:mime/lite`) and the case where it was added by the user (as
+  // in `npm:mime/lite.js/+esm`).
+  return parts.join("/");
 }
 
 /**
--- a/src/npm.ts
+++ b/src/npm.ts
@@ -250,7 +250,17 @@ export async function resolveNpmImport(root: string, specifier: string): Promise
       : "+esm"
   } = parseNpmSpecifier(specifier);
   const version = await resolveNpmVersion(root, {name, range});
-  return fromJsDelivrPath(`/npm/${name}@${version}/${path}`);
+  return fromJsDelivrPath(
+    `/npm/${name}@${version}/${
+      extname(path) || // npm:foo/bar.js
+      path === "" || // npm:foo/
+      path.endsWith("/") || // npm:foo/bar/
+      path === "+esm" || // npm:foo/+esm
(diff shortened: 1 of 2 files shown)


mbostock:
Hmm, there’s a problem with the new scheme: it loses information when we add the implicit `.js` extension. When I have a local path such as

```
/_npm/mime@4.0.1/_esm/lite.js
```

I can’t tell which of the following jsDelivr paths is intended:

```
/npm/mime@4.0.1/lite/+esm
/npm/mime@4.0.1/lite.js/+esm
```

I guess maybe I need to use one of these forms instead for the implicit extension case? But I don’t really like either of them…

```
/_npm/mime@4.0.1/_esm/lite/.js
/_npm/mime@4.0.1/_esm/lite_.js
```

Hmmmmm. 🤔

Fil:
If we take a step back, I think the trouble comes from supporting two APIs to require a module by specifier:
* 1. importing `npm:specifier` in markdown
* 2. passing _npm/specifier URL to preview.

The "bare node module" API (`node.ts`) is simpler, in that it only has to support the first case; the preview URL is a pass-thru of the resolved specifier, directly to the cache.

This means that the map from specifier to local url/cache name does not need to be invertible. In particular it can use a hashing function to differentiate resources that might have similar names. This is what I've explored in #1167.

It leaves us one thing to solve, which is how to find the proper resource on the CDN when we have a specifier. But here I'm getting a bit lost with all the cases.

````
```js
import mime from "npm:mime/lite";
import * as echarts from "npm:echarts/dist/echarts.esm.min.js";
import * as d3 from "npm:d3";
import tex from "npm:@observablehq/tex";
import "npm:katex/contrib/mhchem/mhchem.js";
```
````

**Comment by mbostock** [comment]
The `_node` implementation is simpler because in effect you _always_ get `/+esm`; ESM bundling applies automatically to [any JavaScript import](https://github.com/observablehq/framework/blob/0e6919d2a308d9e12aff43af8f6770add1bf302d/src/javascript/imports.ts#L122-L124) and there’s no way to opt-out of it. Whereas with `_npm` we expose the underlying jsDelivr API so you can opt-in to `/+esm` if you want, but with the exception of the default entry point you otherwise just get the files as-is. And hence only in the `_npm` case do we have the problem that the same resource (file) can be requested either with or without `/+esm`, which leads to the crash when we try to create a directory and a file of the same name already exists, or vice versa.

The mapping from jsDelivr path to Framework path (or equivalently npm specifier, which we can then map to a Framework path) has to be invertible because jsDelivr serves `/+esm` with imports to other jsDelivr paths. When we request `/npm/d3-array@3.2.4/+esm` from jsDelivr, it gives us an import path to `/npm/internmap@2.0.3/+esm`; so in addition to mapping `d3-array` to a jsDelivr path, we need the inverse mapping back to `internmap@2.0.3/+esm` to understand what jsDelivr’s path represents, and to store it with a consistent Framework path in case the user also wants to import `internmap` directly.

Strictly speaking we don’t need to invert a jsDelivr path to an npm specifier; we just need to invert to a Framework path. But I think it’s nice from a logging perspective to show the npm specifier since the user will for the most part think in terms of npm specifiers instead of Framework paths.

**Comment by mbostock** [comment]
I think I’ll try this…

```
/npm/mime@4.0.1/+esm         → /_npm/mime@4.0.1/_esm.js
/npm/mime@4.0.1/lite/+esm    → /_npm/mime@4.0.1/lite._esm.js
/npm/mime@4.0.1/lite.js/+esm → /_npm/mime@4.0.1/lite.js._esm.js
```

mbostock:
Take another look, @Fil? 🙏

Fil:
OK to fix the name later.

Fil (approved):

PR #1166, opened 2024-03-29
more import docs; format=esm platform=browser

### LINKED ISSUE #1162: Support for bare (node) imports is not yet documented
Follow-up to #1156.

mbostock:
Fixes #1162. Also explicitly sets esbuild’s **format** to _esm_ and **platform** to _browser_, since we don’t want it guessing.

Re. **platform**, it defaults to _browser_ so setting it explicitly should have no effect, but feels like a good idea for our usage.

Re. **format**, my understanding is that if **minify** is set to true then **bundle** is implicitly set to true and **format** defaults to _esm_ (because **platform** is *browser*); however we don’t always set **minify** to true (and I noticed when I turned it off in development) in which case **bundle** defaults to false (and rollup does the bundling) and then esbuild doesn’t do any format conversion so we could end up with CommonJS. 😅

### DIFF
--- a/docs/javascript/imports.md
+++ b/docs/javascript/imports.md
@@ -35,7 +35,7 @@ Framework provides a variety of ways to import. When you reference `d3`, `Inputs
 - [local paths](#local-imports), or
 - [remote URLs](#url-imports).
 
-With the exception of remote URL imports, imported modules are bundled with your project, improving performance and security. In some cases, such as stylesheets and WebAssembly modules, you may need to specify additional files to download via [`import.meta.resolve`](#dynamic-imports).
+With the exception of remote URL imports, imported modules are bundled with your project, improving performance, security, and stability. In some cases, such as stylesheets and WebAssembly modules, you may need to specify additional files to download via [`import.meta.resolve`](#dynamic-imports).
 
 ## npm imports
 
@@ -63,9 +63,9 @@ import confetti from "npm:canvas-confetti/dist/confetti.module.mjs";
 
 ### Self-hosting of npm imports
 
-Framework downloads `npm:` imports from jsDelivr during preview and build. This improves performance and security of your built site by removing runtime dependencies on external sites. It also improves performance during local preview by only downloading libraries once.
+Framework downloads `npm:` imports from jsDelivr during preview and build. This improves performance, security, and stability of your built site by removing runtime dependencies on external sites.
 
-Downloads from npm are cached in `.observablehq/cache/_npm` within your [source root](../config#root) (`docs` by default). You can clear the cache and restart the server to re-fetch the latest versions of libraries from npm.
+Downloads from npm are cached in `.observablehq/cache/_npm` within your [source root](../config#root) (`docs` by default). An imported module is downloaded from jsDelivr only if it is not already in the cache. You can clear the cache and restart the server to re-fetch the latest versions of libraries from npm.
 
 Self-hosting of `npm:` imports applies to transitive static and [dynamic imports](#dynamic-imports). In addition to downloading modules, Framework downloads supporting files as needed for [recommended libraries](./imports#implicit-imports) and [`import.meta.resolve`](#import-resolutions). For example, [DuckDB](../lib/duckdb) needs WebAssembly modules, and [KaTeX](../lib/tex) needs a stylesheet and fonts. For dynamic imports and `import.meta.resolve`, Framework is only able to self-host import specifiers that are static string literals.
 
--- a/docs/javascript/imports.md
+++ b/docs/javascript/imports.md
@@ -27,7 +27,7 @@ With the exception of remote URL imports, imports are self-hosted: imported modu
 
 Framework downloads `npm:` imports, as above, from the [npm package registry](https://www.npmjs.com/) via the [jsDelivr CDN](https://www.jsdelivr.com/esm). Unlike [imports from `node_modules`](#node-imports), you don’t have to install `npm:` imports — just import, and the cloud shall provide. 😌
 
-By default, npm imports resolve to the latest version of the given package. Imported versions are resolved on build, or during preview after you clear your npm cache and restart the server.
+By default, npm imports resolve to the latest version of the given package. Imported versions are resolved on build, or during preview after you clear your [npm cache](#self-hosting-of-npm-imports) and restart the server.
 
 To load an earlier or specific version of a package, add a [semver range](https://docs.npmjs.com/about-semantic-versioning). For example, to load major version 1 of `canvas-confetti`:
 
--- a/docs/javascript/imports.md
+++ b/docs/javascript/imports.md
@@ -168,7 +168,7 @@ An import of canvas-confetti is preloaded as:
 <link rel="modulepreload" href="/_npm/canvas-confetti@1.9.2/_esm.js">
 ```
 
-Module preloading applies to transitive dependencies, too. For example, `npm:d3-array` has two dependencies (`isoformat` and `internmap`) which are preloaded as:
+Module preloading applies to transitive dependencies, too. For example, d3-array depends on isoformat and internmap, which together are preloaded as:
 
 ```js
 import "npm:d3-array";
--- a/docs/javascript/imports.md
+++ b/docs/javascript/imports.md
@@ -1,6 +1,6 @@
 # JavaScript: Imports
 
-You can load a library using an [`import` statement](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import). For example, to load [canvas-confetti](https://github.com/catdad/canvas-confetti) from [npm](https://www.npmjs.com/package/canvas-confetti):
+You can load a library using an [`import` statement](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import). For example, to load [canvas-confetti](https://github.com/catdad/canvas-confetti) from npm:
 
 ```js echo
 import confetti from "npm:canvas-confetti";
@@ -15,11 +15,8 @@ import {default as confetti} from "npm:canvas-confetti";
 Depending on the package, you may want to import specific named exports, or to import everything as a namespace. For example:
 
 ```js run=false
-import {rollup} from "npm:d3-array";
-```
-
-```js run=false
-import * as d3 from "npm:d3";
+import {rollup} from "npm:d3-array"; // a single named import
+import * as d3 from "npm:d3"; // a namespace import
 ```
 
 Imported symbols can be referenced in any code block or inline expression — not only in the code block that declares the import.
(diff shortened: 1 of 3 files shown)


docs/javascript/imports.md
```diff
@@ -1,38 +1,49 @@
...
+To load a different entry point, specify the desired path. For example, to load the `lite` entry point to the `mime` package:
 
-```js run=false
-import confetti from "npm:canvas-confetti@1";
+```js echo
+import mime from "npm:mime/lite";
 ```
 
-To load a specific entry point, add a slash and the desired path:
+To instruct jsDelivr to bundle the specified code into an ES module, you can add `/+esm` to the end of the path like so:
```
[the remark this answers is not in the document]
mbostock: Yep, I’m still working on this. I’ll ping when it’s ready for review.

docs/javascript/imports.md
```diff
@@ -1,38 +1,49 @@
...
+
+With the exception of remote URL imports, imports are self-hosted during preview and build: imported modules are downloaded and bundled with your project, improving performance and security. Self-hosting also allows you to control when you upgrade an imported library.
+
 ## npm imports
 
-An `npm:` specifier, as shown above, denotes that the imported library will be loaded from npm rather than from `node_modules`. You don’t have to install imported libraries beforehand — just import, and the cloud shall provide. (This convention is also [used by Deno](https://docs.deno.com/runtime/manual/node/npm_specifiers).)
+Framework downloads `npm:` imports, as above, from the [npm package registry](https://www.npmjs.com/) via the [jsDelivr CDN](https://www.jsdelivr.com/esm). Unlike [imports from `node_modules`](#node-imports), you don’t have to install `npm:` imports — just import, and the cloud shall provide. 😌
+
+By default, npm imports resolve to the latest version of the given package. Since imports are [self-hosted](#self-hosting-of-npm-imports), the resolved version can only change when you re-build your site, or during preview when you clear your npm cache and restart the server.
```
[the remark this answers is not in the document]
mbostock: The cache is documented in the “self-hosting of npm imports” section.

docs/javascript/imports.md
```diff
@@ -1,118 +1,194 @@
...
-You can use an imported library anywhere on the page.
+```js run=false
+import {default as confetti} from "npm:canvas-confetti";
+```
+
+Depending on the package, you may want to import specific named exports, or to import everything as a namespace. For example:
+
+```js run=false
+import {rollup} from "npm:d3-array"; // a single named import
+import * as d3 from "npm:d3"; // a namespace import
```
Fil, line 19: This shows two changes. I understand it makes sense to only import from d3-array when you need only rollup, but it makes the example more difficult to understand. (People might wonder how d3-array is different from d3, and how they could tell.)

docs/javascript/imports.md
```diff
@@ -1,118 +1,194 @@
...
+
 ## npm imports
 
-An `npm:` specifier, as shown above, denotes that the imported library will be loaded from npm rather than from `node_modules`. You don’t have to install imported libraries beforehand — just import, and the cloud shall provide. (This convention is also [used by Deno](https://docs.deno.com/runtime/manual/node/npm_specifiers).)
+Framework downloads `npm:` imports from the [npm package registry](https://www.npmjs.com/) via the [jsDelivr CDN](https://www.jsdelivr.com/esm). Unlike [imports from `node_modules`](#node-imports), you don’t have to install `npm:` imports — just import, and the cloud shall provide. 😌
 
-Under the hood, `npm:` imports are powered by [jsDelivr’s esm.run CDN](https://www.jsdelivr.com/esm). The import above is thus equivalent to:
+By default, npm imports resolve to the latest version of the given package. Imported versions are resolved on build, or during preview after you clear your [npm cache](#self-hosting-of-npm-imports) and restart the server. To load an earlier or specific version of a package, add a [semver range](https://docs.npmjs.com/about-semantic-versioning). For example, to load major version 1 of canvas-confetti:
```
Fil, line 44:
```suggestion
By default, npm imports resolve to the latest version of the given package. Imported versions are resolved on build or during preview and cached in your [npm cache](#self-hosting-of-npm-imports). To load an earlier or specific version of a package, add a [semver range](https://docs.npmjs.com/about-semantic-versioning). For example, to load major version 1 of canvas-confetti:
```

“Imported versions are resolved on build, or during preview” reads as if build did not use the npm cache and always did a fresh resolution.

docs/javascript/imports.md
```diff
@@ -1,118 +1,194 @@
...
-To load a specific entry point, add a slash and the desired path:
+Similarly, to load the file `dist/confetti.module.mjs` from canvas-confetti:
 
 ```js run=false
-import confetti from "npm:canvas-confetti@1/dist/confetti.module.mjs";
+import confetti from "npm:canvas-confetti/dist/confetti.module.mjs";
 ```
 
-If you’re having difficulty getting an import working, it may help to browse the package and see what files are available as well as what’s exported in the `package.json`. You can browse the contents of a published module via jsDelivr; for example, for `canvas-confetti` see <https://cdn.jsdelivr.net/npm/canvas-confetti/>.
+<div class="tip">If you’re having difficulty importing, it may help to browse the package and see what files are available, and what’s exported in the <code>package.json</code>. You can browse the contents of a published module via jsDelivr; for example, for canvas-confetti see <a href="https://cdn.jsdelivr.net/npm/canvas-confetti/">https://cdn.jsdelivr.net/npm/canvas-confetti/</a>.</div>
```
Fil, line 62:
```suggestion
<div class="tip">If you’re having difficulty importing, it may help to browse the package and see what files are available, and what’s exported in the <code>package.json</code>. You can browse the contents of a published module via jsDelivr; for example, see <a href="https://cdn.jsdelivr.net/npm/canvas-confetti/">https://cdn.jsdelivr.net/npm/canvas-confetti/</a>.</div>
```

docs/javascript/imports.md
```diff
@@ -1,118 +1,194 @@
...
 
 ### Self-hosting of npm imports
 
-Framework automatically downloads `npm:` imports from jsDelivr during preview and build. This improves performance and security of your built site by removing external code dependencies. It also improves performance during local preview by only downloading libraries once.
+Framework downloads `npm:` imports from jsDelivr during preview and build. This improves performance and security of your built site by removing runtime dependencies on external sites. It also improves performance during local preview by only downloading libraries once.
```
Fil, line 66:
```suggestion
Framework downloads `npm:` imports from jsDelivr during preview and build. This improves performance and security of your built site and of the local preview by removing runtime dependencies on external sites.
```

docs/javascript/imports.md
```diff
@@ -1,118 +1,194 @@
...
 
-<div class="tip">You can load a library from a CDN at runtime by importing a URL. However, we recommend self-hosted <code>npm:</code> to improve performance and security, and to improve reliability by letting you control when you upgrade.</div>
+Imports from `node_modules` are cached in `.observablehq/cache/_node` within your [source root](../config#root) (`docs` by default). You shouldn’t need to clear this cache as it is automatically managed, but feel free to clear it you like.
 
-Transitive static imports are also registered as module preloads (using `<link rel="modulepreload">`), such that the requests happen in parallel and as early as possible, rather than being chained. This dramatically improves performance on page load. Framework also preloads `npm:` imports for [`FileAttachment`](./files) methods, such as `d3-dsv` for [CSV](../lib/csv).
+Framework (via esbuild) automatically converts CommonJS to ES modules. However, not all Node-compatible packages are usable in the browser; node imports are only supported for browser-compatible modules that do not rely on Node-specific APIs.
```
Fil, line 92:
```suggestion
Framework (via [esbuild](https://esbuild.github.io/)) automatically converts CommonJS to ES modules. However, not all Node-compatible packages are usable in the browser; node imports are only supported for browser-compatible modules that do not rely on Node-specific APIs.
```

docs/javascript/imports.md
```diff
@@ -1,118 +1,194 @@
...
-```js run=false
-import confetti from "https://cdn.jsdelivr.net/npm/canvas-confetti@1.9.2/+esm";
+An import of canvas-confetti is preloaded as:
+
+```html run=false
+<link rel="modulepreload" href="/_npm/canvas-confetti@1.9.2/_esm.js">
 ```
 
-Version resolution locks the version of imported libraries so you don’t have to worry about new releases breaking your built site in the future. At the same time, you’ll conveniently get the latest version of libraries during local development and the next time you build.
+Module preloading applies to transitive dependencies, too. For example, `npm:d3-array` has two dependencies which are preloaded as:
```
Fil, line 171:
```suggestion
Module preloading applies to transitive dependencies, too. For example, `npm:d3-array` has two dependencies (`isoformat` and `internmap`) which are preloaded as:
```

docs/javascript/imports.md
```diff
@@ -141,7 +219,7 @@ Click on any of the imported symbols below to learn more.
 
 ## Require
 
-If you’re familiar with Observable notebooks, you may have noticed that we don’t mention `require` above. We recommend that you avoid `require` as the underlying Asynchronous Module Definition (AMD) convention has been made obsolete by standard imports in JavaScript, and AMD tends to be implemented inconsistently by libraries.
+If you’re familiar with Observable notebooks, you be familiar with `require`. We recommend that you avoid `require` as the underlying Asynchronous Module Definition (AMD) convention has been made obsolete by standard imports in JavaScript, and AMD tends to be implemented inconsistently.
```
Fil, line 222:
```suggestion
If you’re familiar with Observable notebooks, you may be familiar with `require`. We recommend that you avoid `require` as the underlying Asynchronous Module Definition (AMD) convention has been made obsolete by standard imports in JavaScript, and AMD tends to be implemented inconsistently.
```

docs/javascript/imports.md
mbostock: The point is “depending on the package” in the line above. It depends on what you need too, of course. But I wouldn’t recommend `import {rollup} from "npm:d3"` since you would pull in all 30 packages even though you just need d3-array. And I wouldn’t recommend `import * as d3 from "npm:d3-array"` because it gives the impression that all of D3 is available. So we could say `import * as array from "npm:d3-array"` here, but is that better? I can’t think of a situation where I would use that over specific named imports.

Fil: maybe reorder and adapt the comments:

```suggestion
import * as d3 from "npm:d3"; // import all of D3 with a namespace import
import {rollup} from "npm:d3-array"; // a single named import from a specific module
```

docs/javascript/imports.md
mbostock: It doesn’t remove the dependency from preview, though, since you still have the dependency the first time you import the library (before it is cached). Also I think “runtime dependency” only makes sense in the context of build (where “runtime” refers to running the built site); during preview, everything is runtime.

docs/javascript/imports.md
mbostock: I ordered it to match the sentence above “you may want to import specific named exports, or to import everything as a namespace.” I can reorder that sentence too but I don’t understand why it’s easier to understand if the namespace is first? If anything I think importing the default export as shown in the canvas-confetti examples is closer to importing a namespace.

mbostock: I changed the namespace comment to `// import everything as a namespace`. Does that help?

Fil (approved):
lgtm!

docs/javascript/imports.md
mbostock: Okay, I removed this line, and instead added “An imported module is downloaded from jsDelivr only if it is not already in the cache” to the next paragraph, which I think is better because it explains a little more the nature of our caching: it never expires, and versions are resolved against what’s in the cache. Although I don’t say this so explicitly… and I don’t want to explain “satisfying” semantic versions _etc._ 😓

PR #1170, opened 2024-03-29
Add interactive version of PMMS dashboard

eagereyes:
This adds a (simple) interactive version of the PMMS dashboard, as described in this blog post: https://observablehq.com/blog/how-to-add-interactivity-observable-framework-dashboard


mbostock (commented):
Should we add `sidebar: false` since now there are two pages, so the sidebar will appear by default?

eagereyes:
Okay, done! I've also added "(interactive)" to the interactive version, so it makes some sense in the sidebar when it is opened. It makes the title a bit long, but since we want to link to it directly we probably want the full title there (not just "PMMS, interactive version" or similar).

mbostock (approved):

PR #1171, opened 2024-03-29
Add direct links to Observable settings in docs

mythmon:
ObservableHQ recently added support for direct links to workspace pages. It will either prompt the user for the workspace they want to use, or directly navigate to it in the case they only have one workspace.

### DIFF
--- a/docs/deploying.md
+++ b/docs/deploying.md
@@ -68,7 +68,7 @@ When deploying automatically, you can’t sign-in in your browser the way you di
 
 To create an API key:
 
-1. Open the [API Key settings](https://observablehq.com/select-workspace?next=api-keys-settings) for your ObservableHQ workspace.
+1. Open the [API Key settings](https://observablehq.com/select-workspace?next=api-keys-settings) for your Observable workspace.
 2. Click **New API Key**.
 3. Check the **Deploy new versions of projects** checkbox.
 4. Give your key a description, such as “Deploy via GitHub Actions”.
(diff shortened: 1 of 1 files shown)


mythmon:
CC @tophtucker

docs/deploying.md
```diff
@@ -68,13 +68,11 @@ When deploying automatically, you can’t sign-in in your browser the way you di
...
 To create an API key:
 
-1. Go to [observablehq.com](https://observablehq.com).
-2. In the left sidebar, click **Settings**.
-3. In the right sidebar, click **API / Notebook Keys**.
-4. Click **New API Key**.
-5. Check the **Deploy new versions of projects** checkbox.
-6. Give your key a description, such as “Deploy via GitHub Actions”.
-7. Click **Create API Key**.
+1. Open the [API Key settings](https://observablehq.com/select-workspace?next=api-keys-settings) for your ObservableHQ workspace.
```
mbostock, line 71:
```suggestion
1. Open the [API Key settings](https://observablehq.com/select-workspace?next=api-keys-settings) for your Observable workspace.
```

We never use “ObservableHQ” to refer to Observable. The `observablehq` name is used solely to avoid conflict in shared, strict global namespaces that we can’t control (e.g., npm scope, DNS, social media handles), not part of our natural language brand identity.

mbostock (changes requested):
Nice, this is helpful!

Maybe `/select-workspace` could auto-focus your “current” workspace by default so you can hit Enter to continue? Mostly helpful for users like me that have dozens of workspaces, so low priority though.

mbostock (approved):

PR #1172, opened 2024-03-29
FileAttachment.dsv

mbostock:
Adds a `FileAttachment.dsv` method which takes a **delimiter** option, so you’re not limited to CSV/TSV.


Fil (approved):
👍

chphmh:
Can this delimiter option also be used when using Inputs.file({accept: ".csv"}) as file source?

mbostock:
@chphmh Yes. Please open a [discussion](https://github.com/observablehq/framework/discussions) if you’re running into trouble or would like help. Thank you!

PR #1177, opened 2024-03-30
jquery-ui shim

mbostock:
I don’t think we want to do this, but sharing just as a pedagogical example. This makes the following imports work:

```js
import $ from "npm:jquery";
import "npm:jquery-ui";
```

This is quite a bit simpler than the current alternative:

````md
<link rel="stylesheet" href="npm:jquery-ui/dist/themes/base/jquery-ui.css">

```js
import $ from "npm:jquery";
self.jQuery = $;
await import("npm:jquery-ui/dist/jquery-ui.js/+esm");
```
````

The downsides are:

* This only works when the semver range is not specified (or `latest` maybe?)
* This only works when the path is not specified
* This hard-codes the `base` theme, which may not be wanted

In general, I think it’s nice if we can make things “just work”, but maybe inserting a special shim is a little too far.

I suggest we close this PR but might be useful as a future technique.

### DIFF
--- a/docs/lib/jquery-ui.md
+++ b/docs/lib/jquery-ui.md
@@ -1,0 +1,27 @@
+# jQuery UI
+
+```js echo
+import $ from "npm:jquery";
+import "npm:jquery-ui";
+```
+
+<p>
+  <label for="amount">Price range:</label>
+  ${value.join("–")}
+</p>
+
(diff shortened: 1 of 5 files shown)


Fil:
superseded by #1178

PR #1178, opened 2024-03-30
minimal jquery-ui support

mbostock:
Alternative to #1177 that makes jquery-ui a little bit easier to use, and is minimal effort for us.


Fil:
I merged #1174 and added a line of intro

Fil (approved):
It's still a bit of work for the user, but it shows the way. People might want to write and import their own wrapper, or help jquery-ui release an esm.

PR #1200, opened 2024-04-03
support link_shared access level

cinxmo:
This PR supports the "link_shared" project access level. This is to be merged once the rest of the feature is feature-complete.


CobusT:
Perhaps we should reconsider whether we want to do this. There are things we want to do in the platform UI that we may not want to replicate here. For example, we are talking about requiring a password, and adding anough warnings etc. to the UI. I think it is totally fair to NOT show this option as part of a CLI deploy.

PR #1207, opened 2024-04-06
reset after rejection

### LINKED ISSUE #1206: Errors aren’t cleared when a cell doesn’t display anything
For example, given this:

````md
```js echo
const a = 1;
```
```js echo
const a = 2;
```
```js echo
a;
```
````

You should expect that removing the duplicate definition of `a` would clear the error, but it doesn’t, because the cell that errored doesn’t display anything.

### PR DESCRIPTION
**Pull request by mbostock**
Fixes #1206. I tested with the following page:

````md
# Test

```js echo
const a = Promise.resolve(1);
```
```js echo
const b = new Promise((resolve) => setTimeout(() => resolve(2), 2000));
```
```js echo
a + b;
```
${a + b}
````

Which I manually edited to:

````md
# Test

```js echo
const a = Promise.reject(1);
```
```js echo
const b = new Promise((resolve) => setTimeout(() => resolve(2), 1000));
```
```js echo
a + b;
```
${a + b}
````

If you toggle back and forth between the two states you can reproduce the bug and confirm the fix.

### DIFF
--- a/src/client/main.js
+++ b/src/client/main.js
@@ -27,8 +27,10 @@ export function define(cell) {
   cellsById.get(id)?.variables.forEach((v) => v.delete());
   cellsById.set(id, {cell, variables});
   const root = document.querySelector(`#cell-${id}`);
-  const rejected = (error) => (clear(root), console.error(error), root.append(inspectError(error)));
-  const v = main.variable({_node: root, rejected}, {shadow: {}}); // _node for visibility promise
+  const loading = root.classList.contains("observablehq--loading");
+  const pending = () => reset(root, loading);
+  const rejected = (error) => reject(root, error);
+  const v = main.variable({_node: root, pending, rejected}, {shadow: {}}); // _node for visibility promise
   if (inputs.includes("display") || inputs.includes("view")) {
     let displayVersion = -1; // the variable._version of currently-displayed values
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/client/main.js [hunk 1]
```diff
@@ -27,8 +27,10 @@ export function define(cell) {
   cellsById.get(id)?.variables.forEach((v) => v.delete());
   cellsById.set(id, {cell, variables});
   const root = document.querySelector(`#cell-${id}`);
-  const rejected = (error) => (clear(root), console.error(error), root.append(inspectError(error)));
-  const v = main.variable({_node: root, rejected}, {shadow: {}}); // _node for visibility promise
+  const loading = root.classList.contains("observablehq--loading");
```
mbostock, line 30: This is set here for expression code blocks:

https://github.com/observablehq/framework/blob/5b78e40b9e112476775f4781e17510e9608902e3/src/markdown.ts#L111

And here for inline expressions:

https://github.com/observablehq/framework/blob/5b78e40b9e112476775f4781e17510e9608902e3/src/markdown.ts#L265

I suppose we could expose this more explicitly (like we do for the `inline` property), whether there’s an implicit display and hence whether we should show a loading indicator. But it seemed reasonable to just preserve whatever was there before when clearing the error.

Fil (approved):

PR #1216, opened 2024-04-09
page loaders

### LINKED ISSUE #931: Server-Side Rendering (SSR)
Thanks for the great work on Framework!

For even faster (perceived) page loads it would be valuable to render initial views of charts and other content on the server, then "hydrate" them on the client once corresponding runtime cells are ready. For example, I'd like to publish static SVG as part of the initial page content, which can then be replaced with "live" content upon load.

I realize there be dragons here, especially with arbitrary code that may involve randomness, generators, or otherwise cause strange changes of content upon hydration, and so on. Nevertheless, I see a lot of value in being able to knowingly _opt-in_ to SSR.

I imagine one would have to instantiate a runtime during server-side compilation then extract resulting content. We do something similar in [Living Papers](https://github.com/uwdata/living-papers/blob/main/packages/compiler/src/transforms/convert/convert-component.js) using Puppeteer to extract and screenshot runtime-generated content, though a lighter weight approach (e.g., using jsdom and node canvas?) may be preferable and cover common use cases.

It looks like [some work in this direction](https://github.com/observablehq/framework/pull/413) is ongoing, though I'm not sure of the planned scope. So, just wanted to share a plug for this!

mbostock:
A _page loader_ is like a data loader, but for pages instead of files or assets: it allows you to generate Markdown dynamically (at build time) rather than being limited to static Markdown. Page loaders allow server-side rendering #931 and are particularly useful in conjunction with parameterized pages #1523.

(Unlike earlier versions of this PR, this implementation is still Markdown-specific. In the future, we can explore authoring pages in HTML instead of Markdown if desired, but this should be treated as an orthogonal problem.)

TODO

- [x] Page loaders
- [x] Adopt parameterized page loaders in place of `generate-themes` script
- [ ] Documentation
- [ ] Tests

Fixes #931.


mbostock:
We could maybe call this “page generator” (and “generated pages”) instead.

Fil:
Yay, I wanted to explore something like this too. In my idea (that I hadn't really tried to implement yet) parseMarkdown would have been one of the input helpers, instead of something that is applied by default to all (transformed) pages.

Something like this:

```
md  -> parseMarkdown -> parsed page
mdx -> parseMdx      -> (same)
js  -> runJs         -> (same)
```

Whereas this PR treats md as the pivot format:

```
js ---|
md ---|
mdx --|
       \-> md -> parseMarkdown -> parsed page
```

The only advantage to my approach is that the generated page wouldn't go through markdown at all if your input format was not md, thus making sure there are no surprises after your page generator has run. (However, outputting html with no line breaks should guarantee that too.)

This PR's approach is much better:
* well “contained” in terms of API, and simple to reason about if you want to write a new page format generator.
* You can even run the page generators independently from Framework with `node home.page.js > home.md`; like you do with data loaders.
* It does not require the page format generator to be aware of how Framework works: you could even use pandoc.

mbostock:
Indeed that was my original plan which you can see in the first commit, and I’d still like to support that, too — particularly for JSX which is compiled to JavaScript that needs to be evaluated on the server and then serializes to HTML. I was originally thinking of generating JSON using the `RenderPage` interface, but that won’t quite work as-is because `parseMarkdown` calls `parseJavaScript` internally and I don’t want to serialize the entire AST for JavaScript; so I think I’ll pull the `parseJavaScript` out and have `parseMarkdown` simply return the unparsed source for code blocks. I also want to support HTML as an extension, perhaps with `<script type="observablehq">` for code blocks? But not a high priority.

mbostock:
Now that we’ve landed JSX and React separately #1429, I want to reboot this PR to focus just on page loaders.

I want to stick with Markdown as the serialized representation of the page for now. We do need a serialized representation because we want to be able to write page loaders in other languages (_e.g._, `index.md.py` for a Python page loader) just like we do with data loaders. I think we could also support HTML as a serialized representation (_e.g._, `index.html.py`) and even static HTML pages (_e.g._, `index.html` instead of `index.md`). But we’d need to design that representation (for example how do we represent reactive code blocks, probably `<script type="observablehq">` elements?) and that can be a future enhancement.

PR #1217, opened 2024-04-10
EIA example static version

allisonhorst:
This updates the EIA example dashboard as follows: 

- Adds static files (JSON) with EIA data (baHourly, countryInterchange, usOverview)
- Removes any data access w/ EIA API
- Updates tooltips in area chart to local TZ (to match x-axis scale set with `type: "time"`)
- Updates references to "Today" and "Yesterday" with actual dates, since no longer live


mbostock (commented):
The pattern I like to use is that we have data loaders alongside cached output of the data loader. So, you’d have `country-interchange.csv.js` and you’d have `country-interchange.csv`, the latter of which is generated like so:

```
node docs/data/country-interchange.csv.js > docs/data/country-interchange.csv
```

In other words, `country-interchange.csv.js` would only be used manually: you can run the command above whenever you want to update the snapshot. That way the data is easy to update, even if it’s only updated manually, and you can revert back to automatic updates simply by deleting the snapshot. I used this technique with the new deck.gl example:

https://github.com/observablehq/framework/blob/main/docs/data/dft-road-collisions.csv.sh

And the checked-in snapshot:

https://github.com/observablehq/framework/blob/main/docs/data/dft-road-collisions.csv

Can we use that technique here? That would minimize the changes needed to the data loaders. Though we’d still need to replace “Today” and “Yesterday” with actual dates as appropriate. You can even seed the snapshots with the last-generated ones that are currently live (_e.g._, [`country-interchange.f8bf51ca.csv`](https://observablehq.com/framework/examples/eia/_file/data/country-interchange.f8bf51ca.csv)).

allisonhorst:
Yeah that's way better. Will do!

allisonhorst:
Will file a separate PR using Mike's approach above.

PR #1218, opened 2024-04-10
Eia dashboard static version

allisonhorst:
Updates the EIA dashboard: 

- Manually generates data snapshots to docs/data from data loaders 
- Data loaders updated to use v2 of EIA API
- Updates from "Yesterday / Today" to actual date since it's not live
- Minor cleanup


examples/eia/docs/components/charts.js
```diff
@@ -110,9 +111,14 @@ export function countryInterchangeChart(width, height, usDemandGenForecast, coun
...
         fill: "name",
-        tip: true
-      }),
-      Plot.ruleY([0], {strokeOpacity: 0.3})
+        tip: true,
+        title: (d) =>
+          `Country: ${d.name}\nDate: ${d3.timeFormat("%-d %b")(d.date)} ${d.date.toLocaleString("en-US", {
+            hour: "numeric",
+            hour12: true
+          })}\nExported: ${d.value / 1000} GWh`
```
mbostock, line 119: Feels a little funny to mix-and-match d3-time-format and date.toLocaleString here. Also, the `hour12: true` should be implied for the en-US locale. Also, we probably want to use number.toLocaleString to display the number, too, in case there are thousands.

```suggestion
          `Country: ${d.name}\nDate: ${d.date.toLocaleString("en-US", {
            month: "short",
            day: "numeric",
            hour: "numeric"
          })}\nExported: ${(d.value / 1000).toLocaleString("en-US")} GWh`
```

It’s not exactly the same date format (`Apr 10, 9 AM` instead of `10 Apr 9 AM`), but I think the new format is also more common and readable.

examples/eia/docs/components/charts.js
```diff
@@ -110,9 +111,14 @@ export function countryInterchangeChart(width, height, usDemandGenForecast, coun
...
+      })
+      //Plot.ruleY([0], {strokeOpacity: 0.3})
```
mbostock, line 121:
```suggestion
```

examples/eia/docs/data/country-interchange.csv.js
```diff
@@ -1,30 +1,27 @@
+import "dotenv/config";
 import * as d3 from "d3";
 
+// Access key from .env
+const {EIA_KEY} = process.env;
+
 const end = d3.timeDay.offset(d3.timeHour(new Date()), 1);
 const start = d3.timeHour(d3.utcDay.offset(end, -7));
-const convertDate = d3.timeFormat("%m%d%Y %H:%M:%S"); // Convert dates into the format the EIA web API expects
+const convertDate = d3.timeFormat("%Y-%m-%dT%H"); // Convert dates into the format the EIA web API expects
```
mbostock, line 9: I expect these should all be using `d3.utc` instead of `d3.time`? I don’t think it should matter much and it’s a little confusing since in CI we hard-code the America/Los_Angeles time zone. Anyway, fine to leave it as-is, too.

examples/eia/docs/index.md
```diff
@@ -121,9 +112,10 @@ hoursAgoInput.querySelector("input[type=number]").remove();
 ```
 
 ```js
-// Establish current hour and relative day
+// Get current date in readable format
+const formatDate = d3.utcFormat("%B %d, %Y");
 const currentHour = new Date(endHour.getTime() - hoursAgo * MS_IN_AN_HOUR);
-const relativeDay = () => currentHour.getDate() === endHour.getDate() ? "Today" : "Yesterday";
+const currentDate = d3.timeFormat("%-d %b %Y")(currentHour)
```
mbostock, line 118:
```suggestion
const currentDate = d3.timeFormat("%-d %b %Y")(currentHour);
```

examples/eia/docs/index.md
```diff
@@ -37,24 +34,18 @@ const eiaBARef = FileAttachment("data/eia-bia-reference.csv").csv({typed: true})
 const us = await FileAttachment("data/us-states.json").json();
 const nation = us.features.find(({id}) => id === "nation");
 const statemesh = us.features.find(({id}) => id === "statemesh");
```
mbostock, line 36: The `await` here causes the entire block to wait for this file to load, rather than letting the promises resolve independently. One way to let them resolve independently is to make them all promises:

```suggestion
const us = FileAttachment("data/us-states.json").json();
const nation = us.then((us) => us.features.find(({id}) => id === "nation"));
const statemesh = us.then((us) => us.features.find(({id}) => id === "statemesh"));
```

mbostock (changes requested):
A nice improvement! Glad we are using the supported APIs now. A few small suggestions.

allisonhorst:
Thanks for all the feedback & suggestions @mbostock! Accepted the suggested changes.

examples/eia/docs/components/charts.js
```diff
@@ -1,4 +1,5 @@
 import * as Plot from "npm:@observablehq/plot";
+import * as d3 from "npm:d3";
```
mbostock, line 2:
```suggestion
```

This is no longer needed.

mbostock (approved):

Fil (approved):

examples/eia/node
mbostock, line 1: Looks like this file was spuriously committed. I’ll delete it.

PR #1226, opened 2024-04-11
Handle created deploy status, in case deploy hasn't yet updated to pending.

mcglincy:
I'll add a test case.


mythmon (changes requested):
Waiting for the promised test case

mythmon (approved):

PR #1230, opened 2024-04-12
interactive react button

Fil:
Capture d’écran 2024-04-12 à 15 47 14

Advantage of the button: it works in the built page.


mbostock:
> Advantage of the button: it works in the built page.

What do you mean? The DIV works now, too, both in preview and in build. Did you find a bug?

Fil:
No, not a bug; but the main call to action here is "Try changing the background color to x" which you can't do in the build. And the counter is very tiny.

mbostock:
This page is going to be completely overhauled in #1216 so I don’t think we should worry too much about what’s here.

PR #1238, opened 2024-04-13
Check most-recent source modification time when deciding to do a build.

### LINKED ISSUE #1214: Running deploy without ever running build unexpectedly deploys Hello Framework
From @Fil: if you never call yarn build, but just yarn deploy immediately after you start editing files, it deploys "Hello Framework", with no warning that the build is stale.

mcglincy:
Fixes #1214

Also introduces --build and --no-build deploy flags. Uses Clack `confirm` where possible, and switches "youngest" / "age" variables to prefer least-recent and most-recent modification time millis.

### DIFF
--- a/src/bin/observable.ts
+++ b/src/bin/observable.ts
@@ -108,8 +108,9 @@ try {
     case "deploy": {
       const missingDescription = "one of 'build', 'cancel', or 'prompt' (the default)";
       const staleDescription = "one of 'build', 'cancel', 'deploy', or 'prompt' (the default)";
+      const olderDescription = "one of 'build', 'cancel', 'deploy', or 'prompt' (the default)";
       const {
-        values: {config, root, message, "if-stale": ifStale, "if-missing": ifMissing}
+        values: {config, root, message, "if-stale": ifStale, "if-missing": ifMissing, "if-older": ifOlder}
       } = helpArgs(command, {
         options: {
           ...CONFIG_OPTION,
@@ -124,6 +125,10 @@ try {
           "if-missing": {
             type: "string",
             description: `What to do if the output directory is missing: ${missingDescription}`
+          },
+          "if-older": {
+            type: "string",
+            description: `What to do if the output files are older than source files: ${olderDescription}`
           }
         }
       });
@@ -135,6 +140,10 @@ try {
         console.log(`Invalid --if-missing option: ${ifMissing}, expected ${missingDescription}`);
         process.exit(1);
       }
+      if (ifOlder && ifOlder !== "prompt" && ifOlder !== "build" && ifOlder !== "cancel" && ifOlder !== "deploy") {
+        console.log(`Invalid --if-older option: ${ifOlder}, expected ${olderDescription}`);
+        process.exit(1);
+      }
       if (!process.stdin.isTTY && (ifStale === "prompt" || ifMissing === "prompt")) {
         throw new CliError("Cannot prompt for input in non-interactive mode");
       }
@@ -144,7 +153,8 @@ try {
           config: await readConfig(config, root),
           message,
           ifBuildMissing: (ifMissing ?? "prompt") as "prompt" | "build" | "cancel",
-          ifBuildStale: (ifStale ?? "prompt") as "prompt" | "build" | "cancel" | "deploy"
+          ifBuildStale: (ifStale ?? "prompt") as "prompt" | "build" | "cancel" | "deploy",
+          ifBuildOlder: (ifOlder ?? "prompt") as "prompt" | "build" | "cancel" | "deploy"
         })
       );
       break;
--- a/src/bin/observable.ts
+++ b/src/bin/observable.ts
@@ -106,24 +106,14 @@ try {
       break;
     }
     case "deploy": {
-      const missingDescription = "one of 'build', 'cancel', or 'prompt' (the default)";
-      const staleDescription = "one of 'build', 'cancel', 'deploy', or 'prompt' (the default)";
       const {
-        values: {config, root, message, "if-stale": ifStale, "if-missing": ifMissing, build, "no-build": noBuild}
+        values: {config, root, message, build, "no-build": noBuild}
       } = helpArgs(command, {
         options: {
           ...CONFIG_OPTION,
           message: {
             type: "string",
             short: "m"
-          },
-          "if-stale": {
-            type: "string",
-            description: `What to do if the output directory is stale: ${staleDescription}`
-          },
-          "if-missing": {
-            type: "string",
-            description: `What to do if the output directory is missing: ${missingDescription}`
           },
           build: {
             type: "boolean",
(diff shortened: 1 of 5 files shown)


src/deploy.ts
```diff
@@ -445,6 +456,31 @@ export async function deploy(
...
+  let oldestAge = -Infinity;
+  const nowMs = Date.now();
+
+  try {
+    for await (const file of effects.visitFiles(config.root, {ignoreObservable: false})) {
+      let stat: Stats;
+      const joinedPath = join(config.root, file);
+      try {
+        stat = await effects.stat(joinedPath);
+        if (stat?.isFile()) {
```
mbostock, line 469: This shouldn’t be necessary:

- `effects.stat` always returns a `Promise<Stat>`, so the `?.` shouldn’t be needed
- `effects.visitFiles` only visits files, so the `stat.isFile` shouldn’t be needed

I suppose technically `visitFiles` can return symlinks (and other weird things like block devices maybe?) but that isn’t really intended. As you might guess from the name `visitFiles`, I only intended to enumerate files.

src/deploy.ts
```diff
@@ -445,6 +456,31 @@ export async function deploy(
   Telemetry.record({event: "deploy", step: "finish"});
 }
 
+async function findOldestSourceAge(effects: DeployEffects, config: Config): Promise<number> {
+  let oldestAge = -Infinity;
+  const nowMs = Date.now();
```
mbostock, line 461: I would prefer if this function returned the oldest modification time rather than the “age” relative to now. And same for `findBuildFiles`. That would let us use the same _now_ in both cases, but also I think it’s easier to reason about timestamps than relative times. That would slightly increase the scope of this PR though.

src/deploy.ts
```diff
@@ -445,6 +456,31 @@ export async function deploy(
...
+      try {
+        stat = await effects.stat(joinedPath);
+        if (stat?.isFile()) {
+          if (nowMs - stat.mtimeMs > oldestAge) {
+            oldestAge = nowMs - stat.mtimeMs;
+          }
+        }
+      } catch (error) {
+        // ignore
+      }
```
mbostock, line 476: Do we want to ignore these here? Seems like we should just let them throw? (The ceremony around catching errors and re-wrapping them in `CliError` does seem somewhat cumbersome in `findBuildFiles` so I can understand not wanting to repeat that here, but I’m wondering if that’s really necessary to produce a nice error message, or if we could just let things fail a bit more generically if these errors are going to be very rare.)

test/deploy-test.ts
```diff
@@ -103,21 +108,30 @@ class MockDeployEffects extends MockAuthEffects implements DeployEffects {
     this.deployConfig = config;
   }
 
-  *visitFiles(path: string) {
-    yield* visitFiles(path);
+  *visitFiles(path: string, {ignoreObservable = true} = {}) {
+    yield* visitFiles(path, {ignoreObservable});
```
mbostock, line 112: We probably shouldn’t repeat the default here (in case it changes). Would this work instead?

```suggestion
  *visitFiles(path: string, options?: {ignoreObservable?: boolean}) {
    yield* visitFiles(path, options);
```

mbostock (changes requested):
Thanks for fixing this Matt! 🙏 Some small requests.

And a bigger one (for the future): I was thinking we might also redesign the prompts a bit. In the stale case, I think we should still prompt the user with something like: 

> Your sources have changed since the last time you built. Would you like to re-build before deploying?
> Yes (default) / No

And in the case that the build files are newer, we would _always_ prompt regardless of age, saying:

> You last built this project NN minutes ago. Would you like to re-build before deploy?
> Yes / No

where the default depends on whether _NN_ is more or less than 5 minutes. So you always have the choice of whether to rebuild. You can bypass these prompts with the appropriate command-line flag. (We probably need a new command-line flag that’s just `--build` to force a build, and `--no-build` to force deploying what’s already built, rather than the more complex `--if-stale` and `--if-missing` flags.)

Also, I don’t think we need to provide Cancel as an explicit option, since you can always ^C; just use Clack’s `confirm`.

☝️ But making those changes would increase the scope of this PR, so I think we should land this first, and then revisit the prompts to make them a little friendlier (IMO).

mcglincy:
Working on some prompts clean-up. I may wait to switch the ages to modification times until tomorrow morning over coffee.

mbostock (commented):
Thanks again, Matt. I’d still prefer the more direct `--build` and `--no-build` flags to the ones here, but if you want to stick with this approach then I think we should extend the meaning of the existing `--if-stale` flag to apply to the “older” case rather than introduce a new, separate `--if-older` flag. I think it is reasonable to consider the build “stale” if either the source files are newer _or_ if the build is older than NN minutes.

mcglincy:
Sure, that makes sense. I'll tweak it more tomorrow.

src/bin/observable.ts
```diff
@@ -144,7 +152,8 @@ try {
           config: await readConfig(config, root),
           message,
           ifBuildMissing: (ifMissing ?? "prompt") as "prompt" | "build" | "cancel",
-          ifBuildStale: (ifStale ?? "prompt") as "prompt" | "build" | "cancel" | "deploy"
+          ifBuildStale: (ifStale ?? "prompt") as "prompt" | "build" | "cancel" | "deploy",
+          force: build ? "build" : noBuild ? "deploy" : null
```
mbostock, line 155: We need to use the tokens parsing like we do for `--no-open` so that `--build` and `--no-build` have equal precedence and instead it is order which determines what happens. I’ll push a commit to implement this.

We might also error if both `--build` or `--no-build` is specified and `--if-missing` or `--if-stale` is specified, or log a warning saying that `--if-missing` or `--if-stale` is ignored if `--build` or --no-build` is specified? And maybe even log a deprecation notice for `--if-missing` and `--if-stale` entirely?

src/deploy.ts
```diff
@@ -199,77 +204,79 @@ export async function deploy(
...
       throw error;
     }
   }
 
-  if (!doBuild && youngestAge > BUILD_AGE_WARNING_MS) {
-    switch (ifBuildStale) {
-      case "cancel":
+  if (!doBuild) {
+    const mostRecentSourceMtimeMs = await findMostRecentSourceMtimeMs(effects, config);
+    const buildAge = new Date().getTime() - leastRecentBuildMtimeMs;
```
mbostock, line 237:
```suggestion
    const buildAge = Date.now() - leastRecentBuildMtimeMs;
```

src/deploy.ts
```diff
@@ -199,77 +204,79 @@ export async function deploy(
 
   let buildFilePaths: string[] | null = null;
 
-  let doBuild = false;
-  let youngestAge;
+  let doBuild = force === "build";
```
mbostock, line 207: I think we could simplify this as:

```js
  const doBuild =
    force === "build"
      ? true // always build
      : force === "deploy"
      ? false // never build, and error if build is missing
      : effects.isTty
      ? await askWhetherToBuild() // prompt the user
      : isBuildMissing(); // automatically build if missing
```

This would eliminate the arbitrary 5-minute definition of stale in favor of always prompting when possible (and otherwise only building if the build is missing), and you can bypass the prompt using `--build` or `--no-build` as appropriate. It’s more complicated if we want to be backwards-compatible with `--if-stale` and `--if-missing`, though. 😓

mbostock (changes requested):
This is great, Matt. 👍 

I want to fix the parsing when `--build` and `--no-build` are present (like we do for `--no-open`). I will push a commit.

I also think maybe these arguments shouldn’t be allowed in combination with `--if-stale` and `--if-missing` so that we can reduce the number of states. Or even we should drop `--if-stale` and `--if-missing` entirely given the cost of continuing to support them, even though it wouldn’t be backwards compatible (per my comment on `doBuild`). We can look at the telemetry but I think usage is small enough that we’re being too generous by putting this much effort into continuing to support them.

src/bin/observable.ts
mbostock: I’ve done the first part in 83c07ca.

mbostock: And the second part is addressed in 0859004.

mbostock (approved):

PR #1244, opened 2024-04-15
1.6.0

### LINKED ISSUE #1237: footer: null renders as “null”

mbostock:
Also fixes #1237.


mbostock:
I would like to merge #1245 first, please. 🙏

Fil (approved):

Fil:
(we'll merge when we release so the documentation does not display 1.6.0 too early)

PR #1253, opened 2024-04-17
change default source root from `docs` to `src`

### LINKED ISSUE #752: Rename default “docs” root to “pages”?
Since documentation is not our primary use case, “docs” isn’t a great default. But maybe it’s too disruptive to change now.

Fil:
To make the transition easier, when we normalize the config, we automatically default back to `root="docs"` if `pages/index.md` does not exist and `docs/index.md` exists. And in that case, we show a warning:

> **Deprecation warning:** this project seems to be using the docs folder as its default markdown directory. The default directory is now pages. Please rename docs to pages, or explicitly add root: "docs" in the configuration file.

(this happens in the first commit https://github.com/observablehq/framework/pull/1253/commits/0597a143581778f3153b21f53b0a8360e9109479)

I thought about keeping `docs/` for the Framework documentation, but decided against it: if we like this change, let’s own it.

Note that I have no personal preference: for me `docs` is as good as `pages`. I'm putting up this PR only because I want to see #752 resolved—positively or negatively. This will probably result in many conflits in open PRs, nothing painful but it would be good to decide quickly.

closes #752

supersedes https://github.com/observablehq/framework/compare/main...mythmon/240325/default-root-pages

(documentation changes are easier to read in https://github.com/observablehq/framework/pull/1253/commits/167e6045de4d5ddeae3b94f54bdb8087e25b433e)


Fil:
I'm good with any of docs, src, pages… it's just a convention. Sure, it projects the "typical use case", but I don't think it's crucial to change.

yurivish:
I've built a number of framework projects and my first step is always to rename `docs` to `pages` and would strongly support this change.

Regardless of whether this change goes through I would suggest that the following gitignore rules be moved from the top-level `.gitignore` into a `.gitignore`  file in the pages directory to simplify the process of configuring the directory name, since right now if you change the directory name you have to edit these entries, but if they were relative to the directory that would no longer be needed.

https://github.com/observablehq/framework/blob/main/.gitignore#L6-L8

```
/docs/.observablehq/dist/
/docs/theme/*.md
/docs/themes.md
```

mbostock:
@yurivish Thanks for the feedback. I think those files are specific to our docs, but it does make sense to put a `.gitignore` in the source root with the following rule:

```
.observablehq/cache/
```

Rather than putting this in the top-level `.gitignore`:

```
docs/.observablehq/cache/
```

Would that work for you Yuri? Also what do you think about the name `src` instead of `pages` (or `docs`)?

yurivish:
Oh, oops, yes – I meant that line in the `.gitignore`, thanks. 

I'd be fine with `src` and would no longer change the default if it was `src`. I just really dislike `docs` because in the context of software development it typically means "documentation". I used `pages` for consistency with the `pages` key in the Framework config and because that's what seemed like the most descriptive name for what the directory contains, but either `pages` or `src` works for me.

mbostock:
Here’s what the warning now looks like if the **root** option is missing and the `docs` folder exists:

Screenshot 2024-04-23 at 10 56 27 AM

And if the `docs` folder does not exist (a bit hard to see because it’s drowned out by request logs):

Screenshot 2024-04-23 at 10 57 18 AM

mbostock (approved):
Okay, I changed the default to `src` (if `docs` doesn’t exist) instead of `pages`, and fixed a few more stale references to `docs`. I also simplified how the default is computed: the default of `docs` or `src` only applies when there is no other default (_i.e._, it does _not_ apply if you specify the `--root` command-line flag); and it doesn’t check for `index.md` and instead only checks if `docs` exists.

Also, we now issue a warning whenever the **root** option is unset, essentially making this option required while retaining backwards-compatibility. And the examples and templates now all set `root: "src"` to make the behavior explicit. This should allow us to remove the check for `docs` at some point in the future.

PR #1255, opened 2024-04-17
head, header and footer can be specified as a function

### LINKED ISSUE #56: Footer “edit me” link
If a project is hosted on GitHub (or other hosting service), it’d be nice if we could automatically generate an “edit me” link in the footer.

Fil:
head, header and footer can be specified as a function that receives the page’s meta data (path, title and normalized front matter), and returns an HTML fragment (string).

closes #56

Partially addresses #1036, but I'm not sure how we can (or if we should) expose the meta data more generally:
* to the page renderer, so that we could pass the values to be rendered SSR ([#931](https://github.com/observablehq/framework/issues/931)); probably in post-processing?
* to the page runtime (via a `define("meta", …)`)?

Partially addresses #1161.

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -154,17 +154,21 @@ Whether to show the previous & next links in the footer; defaults to true. The p
 
 ## head
 
-An HTML fragment to add to the head. Defaults to the empty string.
+An HTML fragment to add to the head. Defaults to the empty string. If specified as a function, receives an object with the page’s `title`, (front-matter) `data`, and `path`, and must return a string.
 
 ## header
 
-An HTML fragment to add to the header. Defaults to the empty string.
+An HTML fragment to add to the header. Defaults to the empty string. If specified as a function, receives an object with the page’s `title`, (front-matter) `data`, and `path`, and must return a string.
 
 ## footer
 
-An HTML fragment to add to the footer. Defaults to “Built with Observable.”
+An HTML fragment to add to the footer. Defaults to “Built with Observable.” If specified as a function, receives an object with the page’s `title`, (front-matter) `data`, and `path`, and must return a string.
 
-head, header and footer can be specified as strings, or as functions that receive as arguments the page’s title, front matter, and path, and return a string.
+For example, the following adds a link to the bottom of each page:
+
+```js run=false
+footer: ({path}) => `<a href="https://github.com/example/test/blob/main/src${path}.md?plain=1">view source</a>`,
+```
 
 ## base
 
--- a/docs/config.md
+++ b/docs/config.md
@@ -164,6 +164,8 @@ An HTML fragment to add to the header. Defaults to the empty string.
 
 An HTML fragment to add to the footer. Defaults to “Built with Observable.”
 
+head, header and footer can be specified as strings, or as functions that receive as arguments the page’s title, front matter, and path, and return a string.
+
 ## base
 
 The base path when serving the site. Currently this only affects the custom 404 page, if any.
--- a/src/config.ts
+++ b/src/config.ts
@@ -10,6 +10,7 @@ import wrapAnsi from "wrap-ansi";
 import {LoaderResolver} from "./dataloader.js";
 import {visitMarkdownFiles} from "./files.js";
 import {formatIsoDate, formatLocaleDate} from "./format.js";
+import type {FrontMatter} from "./frontMatter.js";
 import {createMarkdownIt, parseMarkdownMetadata} from "./markdown.js";
 import {isAssetPath, parseRelativeUrl, resolvePath} from "./path.js";
 import {resolveTheme} from "./theme.js";
@@ -43,6 +44,11 @@ export interface Script {
   type: string | null;
 }
 
+/**
+ * A function that generates a page fragment such as head, header or footer.
+ */
+type PageFragmentFunction = (title: string | null, data: FrontMatter, path: string) => string;
+
 export interface Config {
   root: string; // defaults to src
   output: string; // defaults to dist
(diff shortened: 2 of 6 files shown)


mbostock:
Why is it necessary to remove the strict normalization of the front matter? I don’t think we want to encourage users to put anything they want in the front matter; that will make it harder for us to introduce new options in the future without breaking projects. What is the use case for adding arbitrary things into the front matter?

Fil:
The idea is coming from metadata that I would like to both display in the page, and consume outside.

For example, author (to create a byline), source (which I use in pangea to refer to the original notebook, when applicable), tags, date_published, og_image, thumbnail, etc.

Some of these might go in the head (to inform opengraph tags), some in the indexation / listings (typically, the tags), yet others in the footer (e.g. tags with links to the tags pages.

Maybe it's the wrong approach and we should reserve front matter for “official business”, but then we'd need another mechanism that would be more extensible in user land. Maybe it's just a "meta" key in front matter that is not normalized, and made available both for SSR and for client-side JS?

Anyway I can dumb down this PR so that it only does the "head as a function" part and finally closes #56, and we continue the discussion on the extensibility separately.

mbostock:
This also overlaps with the page loader concept #1216 which allows the entire page (not just the head, header, footer) to be dynamically generated.

Fil:
rebased

src/config.ts
```diff
@@ -43,6 +44,11 @@ export interface Script {
   type: string | null;
 }
 
+/**
+ * A function that generates a page fragment such as head, header or footer.
+ */
+type PageFragmentFunction = (title: string | null, data: FrontMatter, path: string) => string;
```
mbostock, line 50: Would rather use named arguments here (`({title, data, path})`), not positional (`(title, data, path)`), so that it’s easier for us to add more arguments in the future without it becoming unwieldy.

Also this type is referenced by `Config` and hence should be exported.

src/config.ts
```diff
@@ -247,6 +253,10 @@ function getPathNormalizer(spec: unknown = true): (path: string) => string {
   };
 }
 
+function pageFragment(spec: unknown) {
```
mbostock, line 256:
```suggestion
function pageFragment(spec: unknown): PageFragmentFunction | string | null {
```

src/markdown.ts
```diff
@@ -358,23 +359,26 @@ function getHead(data: FrontMatter, options: ParseOptions): string | null {
...
   {path, [key]: defaultValue}: ParseOptions
 ): string | null {
   return data[key] !== undefined
     ? data[key]
       ? String(data[key])
       : null
+    : typeof defaultValue === "function"
+    ? rewriteHtmlPaths(defaultValue(title, data, path), path)
     : defaultValue != null
     ? rewriteHtmlPaths(defaultValue, path)
```
mbostock, line 383: Would rather combine this like so:

```suggestion
    : defaultValue != null
    ? rewriteHtmlPaths(typeof defaultValue === "function" ? defaultValue(title, data, path) : defaultValue, path)
```

test/input/build/fragments/index.md
```diff
@@ -0,0 +1,10 @@
+---
+author: "Ignored Anonymous"
+title: Testing fragment functions
+date: 2024-04-18
+keywords: ["very", "much"]
```
mbostock, line 5: FWIW, in YAML I believe you can write this as 

```suggestion
keywords: [very, much]
```

docs/config.md
```diff
@@ -164,6 +164,8 @@ An HTML fragment to add to the header. Defaults to the empty string.
 
 An HTML fragment to add to the footer. Defaults to “Built with Observable.”
 
+head, header and footer can be specified as strings, or as functions that receive as arguments the page’s title, front matter, and path, and return a string.
```
mbostock, line 167: I think this needs to be repeated for each option. Otherwise you’re likely to miss it when you’re looking at **head** or **header** — you won’t know that this sentence applies to multiple options. It would also be nice to include at least one example, though maybe that doesn’t need to be replicated.

mbostock (changes requested):

src/config.ts
```diff
@@ -47,7 +47,15 @@ export interface Script {
...
-type PageFragmentFunction = (title: string | null, data: FrontMatter, path: string) => string;
+export type PageFragmentFunction = ({
+  title,
+  data,
+  path
+}: {
+  title: string | null;
+  data: FrontMatter;
+  path: string;
+}) => string;
```
mbostock, line 58:
```suggestion
}) => string | null;
```

mbostock (changes requested):

mbostock (approved):

Fil:
showing "undefined" or "null" was not very pretty (but … informative!) :)

dleeftink:
I've tried defining a dynamic `head` function by following the footer example, to no avail. When testing it for `header` and `footer`, rather than evaluating the function it returns the function as string in the rendered HTML (when in dev preview, at least). Maybe I am doing something wrong?

Fil:
@dleeftink What version of framework are you using?

dleeftink:
@Fil an older one it seems (1.5.1), updated now tot 1.9.0 and no issues with the main `observablehq.config.js`. Just wondering how to override on a per file basis? Adding an arrow function to the YAML frontmatter of a page like doesn't seem right:

```
footer: ()=>`test`
```

dleeftink:
@Fil an older one it seems (1.5.1), updated now tot 1.9.0 and no issues with the main `observablehq.config.js`. Just wondering how to override on a per file basis? Adding an arrow function to the YAML frontmatter of a page like so doesn't seem right:

```
footer: ()=>`test`
```

dleeftink:
Also, how about making the per-page overrides configurable to replace (`append=false`) or append (`append=true`) to the globally defined functions in `config.js`? This way pages don't interfere with the global settings if only minor additions are required. Something like:

```yaml
header: ({append = true}) => `<i>some subtitle</i>`
```

Maybe overrides can even be configured to replace/append to specific metadata fields, for instance to append a date to a globally defined title or a metadata page description. Not sure what that interface would like tho as it involves creating per-field flags for potentially nested frontmatter items.

Fil:
The current logic works the other way around: the function you define in `observablehq.config.ts` is downstream from the page, not upstream — it receives as input the page's meta data (slug, title and front-matter). One aspect we could make better is to allow "non normalized" front matter fields, so you could _e.g._ define _categories: x, y, z_ in the front matter and use those in the footer. If you want to open a new issue, please try to provide details of the use case you have in mind.

PR #1262, opened 2024-04-18
fix elliptic import

mbostock:
This fixes an import of the [elliptic](https://github.com/indutny/elliptic) package (and its dependencies). This comes from a [Slack request](https://observable-community.slack.com/archives/C03SEMPUF6F/p1713346102969629) which I debugged. I’ll comment on the changes below.

I tested manually with:

````md
# Elliptic

[Elliptic](https://github.com/indutny/elliptic) is an implementation of “fast elliptic curve cryptography in plain JavaScript.”

```js echo
import elliptic from "elliptic";
const {ec: EC} = elliptic;

// Create and initialize EC context
// (better do it once and reuse it)
const ec = new EC("secp256k1");

// Generate keys
const key = ec.genKeyPair();

// Sign the message's hash (input must be an array, or a hex-string)
const msgHash = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10];
const signature = key.sign(msgHash);

// Export DER encoded signature in Array
const derSign = signature.toDER();

display(derSign);
```
````

(Yes, I should probably write a test, but this isn’t especially high-value work and it’s all a hack anyway.)

### DIFF
--- a/src/node.ts
+++ b/src/node.ts
@@ -102,7 +102,7 @@ async function bundle(
       ...(isBadCommonJs(input) ? [(virtual as any)({"-": shimCommonJs(input, require)})] : []),
       importResolve(input, cacheRoot, packageRoot),
       nodeResolve({browser: true, rootDir: packageRoot}),
-      json(),
+      (json as any)(),
       (commonjs as any)({
         esmExternals: true,
         requireReturnsDefault: "preferred"
--- a/src/node.ts
+++ b/src/node.ts
@@ -138,7 +138,7 @@ function importResolve(input: string, cacheRoot: string, packageRoot: string): P
     return typeof specifier !== "string" || // AST node?
       isNodeBuiltin(specifier) || // node built-in, e.g., "node:fs" or "fs"
       isPathImport(specifier) || // relative path, e.g., ./foo.js
-      /^\0?[\w-]+:/.test(specifier) || // windows file path, https: URL, \x00node-resove:, etc.
+      /^\0?[\w-]+:/.test(specifier) || // windows file path, https: URL, \x00node-resolve:, etc.
       specifier === input // entry point
       ? null // don’t do any additional resolution
       : {id: await resolveNodeImportInternal(cacheRoot, packageRoot, specifier), external: true}; // resolve bare import
--- a/src/node.ts
+++ b/src/node.ts
@@ -5,6 +5,7 @@ import op from "node:path";
 import {extname, join} from "node:path/posix";
 import {pathToFileURL} from "node:url";
 import commonjs from "@rollup/plugin-commonjs";
+import json from "@rollup/plugin-json";
 import {nodeResolve} from "@rollup/plugin-node-resolve";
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/node.ts [hunk 1]
```diff
@@ -101,7 +102,11 @@ async function bundle(
       ...(isBadCommonJs(input) ? [(virtual as any)({"-": shimCommonJs(input, require)})] : []),
       importResolve(input, cacheRoot, packageRoot),
       nodeResolve({browser: true, rootDir: packageRoot}),
-      (commonjs as any)({esmExternals: true}),
+      json(),
```
mbostock, line 105: This is needed for CommonJS packages that `require("./package.json")` (which seems like a reasonable and easy thing for us to support).

src/node.ts
```diff
@@ -101,7 +102,11 @@ async function bundle(
...
+      (commonjs as any)({
+        esmExternals: true,
+        requireReturnsDefault: "preferred"
```
mbostock, line 108: This was the only value of the property that allowed us to import elliptic without breaking imports of react. 😞 I don’t really understand this, but it’s described here:

https://github.com/rollup/plugins/tree/master/packages/commonjs/#requirereturnsdefault

My hand-wavy understanding is that we ideally want the default `requireReturnsDefault: false`. However, this assumes that the required module is compatible with cjs-module-lexer, and thus Rollup is able to convert the CommonJS exports into named exports. But often cjs-module-lexer doesn’t work, and so we can’t know whether a `require("foo")` will be converted into a default export or named exports. The *preferred* setting lets us use the default export is used if it’s available, and otherwise we use named exports. (The *auto* setting didn’t work in the case of the [brorand](https://www.npmjs.com/package/brorand) package, whose CommonJS is converted into both a named export and a default export.)

I’m sure there remain plenty of cases where you still can’t import CommonJS, but this seems like a more compatible setting than the default.

src/node.ts
```diff
@@ -133,7 +138,7 @@ function importResolve(input: string, cacheRoot: string, packageRoot: string): P
     return typeof specifier !== "string" || // AST node?
       isNodeBuiltin(specifier) || // node built-in, e.g., "node:fs" or "fs"
       isPathImport(specifier) || // relative path, e.g., ./foo.js
-      /^\w+:/.test(specifier) || // windows file path, https: URL, etc.
+      /^\0?[\w-]+:/.test(specifier) || // windows file path, https: URL, \x00node-resove:, etc.
```
mbostock, line 141: For reasons I really don’t understand, the node-resolve plugin tries to resolve things like `"\x00node-resolve:empty.js"`. If we don’t return null here, we’ll try to resolve the id as a node package, which obviously will fail. I think this expanded pattern will match the weird internal names used by node-resolve now. 🤷

Fil, line 141:
```suggestion
      /^\0?[\w-]+:/.test(specifier) || // windows file path, https: URL, \x00node-resolve:, etc.
```

Fil (approved):
I fixed a comment :)

PR #1267, opened 2024-04-19
better messaging when deploying large files

mythmon:
- **Provide more errors when failing to upload a file**
- **Add pre-deploy check for large files**

image

### DIFF
--- a/src/bin/observable.ts
+++ b/src/bin/observable.ts
@@ -200,7 +200,7 @@ try {
     }
   }
 } catch (error: any) {
-  const wrapWidth = process.stdout.columns ?? 100;
+  const wrapWidth = process.stdout.columns ?? 80;
   const bugMessage = "If you think this is a bug, please file an issue at";
   const bugUrl = "https://github.com/observablehq/framework/issues";
   const clackBugMessage = () => {
--- a/src/bin/observable.ts
+++ b/src/bin/observable.ts
@@ -108,7 +108,7 @@ try {
     }
     case "deploy": {
       const {
-        values: {config, root, message, build, "no-build": noBuild}
+        values: {config, root, message, build, "no-build": noBuild, "allow-large-files": allowLargeFiles}
       } = helpArgs(command, {
         options: {
           ...CONFIG_OPTION,
@@ -123,6 +123,10 @@ try {
           "no-build": {
             type: "boolean",
             description: "Don’t build before deploying; deploy as is"
+          },
+          "allow-large-files": {
+            type: "boolean",
+            description: "Don't warn about large files"
           }
         }
       });
@@ -130,7 +134,8 @@ try {
         deploy.deploy({
           config: await readConfig(config, root),
           message,
-          force: build ? "build" : noBuild ? "deploy" : null
+          force: build ? "build" : noBuild ? "deploy" : null,
+          allowLargeFiles
         })
       );
       break;
--- a/src/bin/observable.ts
+++ b/src/bin/observable.ts
@@ -2,6 +2,7 @@
 import type {ParseArgsConfig} from "node:util";
 import {parseArgs} from "node:util";
 import * as clack from "@clack/prompts";
+import wrapAnsi from "wrap-ansi";
 import {readConfig} from "../config.js";
 import {CliError} from "../error.js";
 import {faint, link, red} from "../tty.js";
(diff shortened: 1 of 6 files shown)


mythmon:
> Coverage for lines (79.97%) does not meet threshold (80%) for src/deploy.ts

So close! I'll add some more tests for the new messages.

src/bin/observable.ts
```diff
@@ -122,14 +123,19 @@ try {
           "no-build": {
             type: "boolean",
             description: "Don’t build before deploying; deploy as is"
+          },
+          "allow-large-files": {
+            type: "boolean",
+            description: "Don't warn about large files"
```
Fil, line 129:
```suggestion
            description: "Don’t warn about large files"
```

src/deploy.ts
```diff
@@ -247,6 +257,55 @@ export async function deploy(
...
+          "The following files are too large and may fail to deploy:",
+          ...warnings.map(({path, size}) => `  - ${path} (${formatByteSize(size)})`)
+        ];
+        clack.log.warning(wrapAnsi(output.join("\n"), effects.outputColumns - 3));
+      }
+      clack.log.info(
+        wrapAnsi(
+          "Observable does not allow deploying files larger than 50MB. " +
+            "You can pass --allow-large-files to bypass this warning, " +
+            "though the deploy will still likely fail.",
```
Fil, line 292:
```suggestion
            "You can pass --allow-large-files to suppress this warning, " +
            "though the deploy will still likely fail.",
```

src/format.ts
```diff
@@ -13,3 +13,15 @@ export function formatLocaleDate(date: Date, locale: Intl.LocalesArgument = "en-
...
+export function formatByteSize(n: number): string {
+  const suffixes = ["bytes", "KB", "MB", "GB"];
+  for (const suffix of suffixes) {
+    if (n < 1300) {
+      if (suffix === "bytes") return `${n} ${suffix}`;
+      return `${n > 100 ? n.toFixed(0) : n.toFixed(1)} ${suffix}`;
+    }
+    n /= 1000;
+  }
+  return `${n.toFixed(1)} TiB`;
```
Fil, line 26:
```suggestion
  return `${n.toFixed(1)} TB`;
```

test/format-test.ts
```diff
@@ -21,3 +21,20 @@ describe("formatLocaleDate", () => {
...
+    assert.strictEqual(formatByteSize(1200), "1200 bytes");
+    assert.strictEqual(formatByteSize(1_500), "1.5 KB");
+    assert.strictEqual(formatByteSize(3_000), "3.0 KB");
+    assert.strictEqual(formatByteSize(400_000), "400 KB");
+    assert.strictEqual(formatByteSize(50_000_000), "50.0 MB");
+    assert.strictEqual(formatByteSize(1_250_000_000), "1250 MB");
+    assert.strictEqual(formatByteSize(1_500_000_000), "1.5 GB");
+    assert.strictEqual(formatByteSize(3_000_000_000), "3.0 GB");
+    assert.strictEqual(formatByteSize(600_000_000_000), "600 GB");
+    assert.strictEqual(formatByteSize(60_598_160), "60.6 MB");
```
Fil, line 38:
```suggestion
    assert.strictEqual(formatByteSize(60_598_160), "60.6 MB");
    assert.strictEqual(formatByteSize(60_598_160_000), "60.6 TB");
```

Fil (commented):
The messaging is better than a raw error, for sure, but the story that we're building around `--allow-large-files` now feels very messy.

From the pov of a user I don't understand how it helps me to suppress a warning and try again, if it's going to fail. Is the system recommending that I force-try? And if it fails again, should I try again (and again…), until it passes?

Because the message conveys uncertainty, about something that (again, as a naive user) should be a clear rule, with a binary resolution (a file is either too large, or it's not), I'm starting to wonder if I should trust this, or if I'm missing something obvious.

Are the file sizes, or the size checks, different on client and server? If so, can we fix that? 

If there is uncertainty that can't be fixed (because payloads are measured differently, maybe?), can we not just stop the user conservatively a few kB below the threshold, before we're in the "random" area, so as to make the script appear more "solid"?

(I would also love to be on the future pro plan where I can share large parquet or video files without having to rely on another provider.)

src/bin/observable.ts
```diff
@@ -199,29 +205,37 @@ try {
     }
   }
 } catch (error: any) {
+  const wrapWidth = process.stdout.columns ?? 100;
```
mbostock, line 208:
```suggestion
  const wrapWidth = process.stdout.columns ?? 80;
```

We should centralize this, but elsewhere we use 80, and I think that’s more readable than 100.

mbostock:
Thoughtful review, @Fil. 🙏 

I believe the thinking behind the initial warning is that we will save the user time by letting them know upfront that the deploy is likely to fail because it contains large files. However, _likely_ is the best the client can do now; the client can’t know for sure that the deploy will fail because the client’s logic may drift with the server’s logic in the future. And so we don’t want a hard error on the client since (in the future) the server might allow the deploy to succeed. But the `--allow-large-files` isn’t very friendly in the sense that it provides a recommendation (now) that is still likely to fail.

So here are some alternatives to consider:

1. Don’t do the initial warning, don’t add the `--allow-large-files` flag, and just show a nicer error message if an upload fails. This is the minimal fix to the problem in that it provides a clear indication of why the deploy failed, and makes clear that the only thing the user can do to fix the problem is to remove the large file (or make it smaller).<br><br> Since the server is solely responsible for the logic, the client won’t issue a spurious warning if the server would allow the upload; the server entirely controls what is allowed. This also has the advantage that other errors, such as hitting your upload quota, would be handled the same way as a large file. The only downside is that it could waste some of the user’s time deploying a project that we could reasonably anticipate would fail.

2. Do a preflight check by asking the server whether the upload is allowed. This could be done on a file-by-file basis (immediately prior to uploading each file) or in bulk before the deploy starts. This allows the client & server to check that the deploy _would_ succeed much faster than doing the actual deploy. However, this is of course more work, since we need to define a new API endpoint. It could also make deploys slightly slower in the common case, if there are more roundtrips.

3. As a variation of (2), just fix the upload of an individual file to be fail-fast. I think currently we allow you to upload a large file and it doesn’t fail until the end? But I haven’t checked and I could be wrong.

My recommendation if we are unsure to chose between 1, 2, & 3 is that we do 1 since it is the minimal work, and then we can followup with 2 or 3 later as a further improvement.

mbostock:
> The idea in that case is that, to enjoy (future) larger files, you'd have to upgrade your client—which I think is fair game.

How would the user know to upgrade to fix the problem though? That seems like an unlikely solution to the presented error, and asking the user to upgrade when it won’t fix the problem seems misleading (as unlikely to succeed as asking them to `--allow-large-files`, and more work).

Fil:
I like your suggestion 2, but think we can make it simpler initially:

* Query a new API endpoint for the size limit (/api/preflight/v1), sending user authentication and total file size. Default response is {size: 50_000_000, message: “Your project’s files are limited to 50MB”}
* If a file is larger than the size limit, show the message and block the deploy.

We don’t have to implement a complicated API endpoint now to make this work—for now it could even be a dummy static file, since it's the same answer for everyone. (We could develop this on the client even before we set up a dummy endpoint, by having a conditional on 404 => default message. "dirty", but makes sequencing easier.)

When we have something more interesting to say than "50MB", we implement the API endpoint to reflect a message depending on the new server policy, on the user’s tier, current total quota, and any other considerations. (The API is versioned, future smarter clients will use /api/preflight/v2, which accepts a files manifest, and returns a more elaborate list of instructions as to which files must be transmitted, and more adequate computation of quota on the server.)

Fil (approved):
Thanks! I tested by inflating `horse.mp4`:

```
■  Error: While uploading _file/horse.xxxxxxx.mp4 (616 MB): Unexpected response status 413 for POST https://api.observablehq.com/cli/deploy/xxxxxxxx/file
```

PR #1268, opened 2024-04-21
fix remote imports docs

russbiggs:
This fixes a minor typo/confusing sentence in the `Remote imports` section of of the imports documentation page. I've made an assumption on what was originally intended, but its a bit unclear as it seems two different thoughts got jumbled into one.

This changes:

`Unlike npm: and node_modules imports, remove imports are not be self-hosted;...`

to 

`Unlike npm: and node_modules imports, remote imports will not be self-hosted;`


mbostock (approved):
Thanks!

PR #1270, opened 2024-04-21
fix local script resolution

Fil:
with this configuration file:

```
export default {
  scripts: [{type: "module", async: true, src: "hello.js"}]
};

```

we expect to see this for a page in the root:

```
<script type="module" async="" src="./_import/hello.xxxx.js"></script>
```

and this for a page in a subfolder:

```
<script type="module" async="" src="../_import/hello.xxxx.js"></script>
```

and the `_import/hello.xxxx.js` should be built.

In other words, local scripts should resolve correctly, including in subfolders. Local scripts should be built.

Instead we're seeing:

```
<script type="module" async src="../../../hello.js"></script>
```

and the `hello.js` file is not built.

(I'm only adding a missing test for now.)


mbostock:
Yeah, looks like we’re not resolving the paths to local scripts as intended. I can fix now.

mbostock (approved):
I fixed this by piggybacking on the existing head rendering, rather than having a separate path for rendering scripts. Now when you call `parseMarkdown`, the **scripts** config option is baked-in to the returned `page.head`, and as a result you get the same resolve behavior as if you had used the **head** config option (with the distinction that the **scripts** config option will merge with the **head** front matter option, too).

A side effect of this change is that we now appropriately preload local scripts declared via the **scripts** config option, too, and support both JavaScript and non-JavaScript scripts. Of course, there’s not a strong reason to use the **scripts** config option over the **head** config option, either… so we could perhaps deprecate **scripts** entirely. But at least now it’s not much work to support.

Fil:
Thanks for fixing. Agree, we should deprecate **scripts** since it has no added value and just makes things more noisy.

PR #1273, opened 2024-04-21
unpin parquet-wasm

mbostock:
The latest [0.6.0 release of parquet-wasm](https://github.com/kylebarron/parquet-wasm/releases/tag/v0.6.0) fixes the conditional exports so this mostly works out of the box. There’s a slight wrinkle that the default path to the `.wasm` file is incorrect when loaded via the `/+esm` entry point, but we can fix that by passing in the URL to the `.wasm` file explicitly.

Thanks @kylebarron for the fixes!

### DIFF
--- a/examples/api/docs/index.md
+++ b/examples/api/docs/index.md
@@ -1,6 +1,4 @@
 ```js
-import "npm:apache-arrow";
-import "npm:parquet-wasm/esm/arrow1.js";
 import {ApiHeatmap} from "./components/apiHeatmap.js";
 import {ApiHistogram} from "./components/apiHistogram.js";
 ```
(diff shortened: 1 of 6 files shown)


examples/api/docs/index.md
```diff
@@ -1,6 +1,4 @@
 ```js
-import "npm:apache-arrow";
-import "npm:parquet-wasm/esm/arrow1.js";
```
mbostock, line 3: This was to preload Arrow and parquet-wasm, but we do the preloading automatically now so this isn’t needed.

Fil (approved):

mbostock:
That’s right. I also think that jsDelivr might be smart enough to rewrite the path if parquet-wasm used `import.meta.resolve` instead of `import.meta.url`, but I’m not sure.

PR #1276, opened 2024-04-22
add github template

Fil:
Several pages & data loaders. We might want more or less, and better, but we have a base.

An additional clack dialog helps to populate `.env`.

A few options (including limits) are described in the README. They could be shown, commented out, in the default .env.


Fil:
Is there a way we could show the data loaders logs during the "Initializing Framework cache..." phase? My github template data loaders take a few minutes to run (277s.), and i'd like to display progress. (The alternative is to issue a warning to say "this might need to run a few minutes as it crunches the data").

Fil:
superseded by the https://github.com/observablehq/oss-analytics/ project

PR #1278, opened 2024-04-24
use file manifests during uploads

mythmon:
This changes the deploy command to first upload a manifest of files in the output directory. The manifest contains the path, size, and hash of each file found. The server then sends back instructions to either upload the file, skip it (because it already exists on the server), or an error (such as a file being too large).

The manifest process is mostly invisible to the user. The upload spinner briefly shows that it is preparing a manifest, and the final total shows files that were skipped because they already exist.

I added a new pair of flags, `--manifest` and `--no-manifest`, which use the new and old upload strategies, respectively. The default is to use manifests. My thinking here was that we should merge this PR with the flag, and validate it in CI for a few days. If there is a problem, we can use the `--no-manifest` option while we fix it. I expect that for the next release we should remove the flag and only have one upload strategy.

This depends on server side changes that haven't been deployed yet. I'll update this PR once they are ready to test on observablehq.com.

### DIFF
--- a/src/bin/observable.ts
+++ b/src/bin/observable.ts
@@ -108,7 +108,7 @@ try {
     }
     case "deploy": {
       const {
-        values: {config, root, message, build, "no-build": noBuild, manifest}
+        values: {config, root, message, build}
       } = helpArgs(command, {
         options: {
           ...CONFIG_OPTION,
@@ -123,17 +123,14 @@ try {
           "no-build": {
             type: "boolean",
             description: "Don’t build before deploying; deploy as is"
-          },
-          manifest: {type: "boolean", description: "Use the new upload strategy"},
-          "no-manifest": {type: "boolean", description: "Use the old upload strategy"}
+          }
         }
       });
       await import("../deploy.js").then(async (deploy) =>
         deploy.deploy({
           config: await readConfig(config, root),
           message,
-          force: build ? "build" : noBuild ? "deploy" : null,
-          manifest
+          force: build === true ? "build" : build === false ? "deploy" : null
         })
       );
       break;
--- a/src/bin/observable.ts
+++ b/src/bin/observable.ts
@@ -108,7 +108,7 @@ try {
     }
     case "deploy": {
       const {
-        values: {config, root, message, build, "no-build": noBuild}
+        values: {config, root, message, build, "no-build": noBuild, manifest}
       } = helpArgs(command, {
         options: {
           ...CONFIG_OPTION,
@@ -123,14 +123,17 @@ try {
           "no-build": {
             type: "boolean",
             description: "Don’t build before deploying; deploy as is"
-          }
+          },
+          manifest: {type: "boolean", description: "Use the new upload strategy"},
+          "no-manifest": {type: "boolean", description: "Use the old upload strategy"}
         }
       });
       await import("../deploy.js").then(async (deploy) =>
         deploy.deploy({
           config: await readConfig(config, root),
           message,
-          force: build ? "build" : noBuild ? "deploy" : null
+          force: build ? "build" : noBuild ? "deploy" : null,
+          manifest
         })
       );
       break;
--- a/src/deploy.ts
+++ b/src/deploy.ts
@@ -1,7 +1,6 @@
 import {createHash} from "node:crypto";
-import {createReadStream} from "node:fs";
 import type {Stats} from "node:fs";
-import {stat} from "node:fs/promises";
+import {readFile, stat} from "node:fs/promises";
 import {join} from "node:path/posix";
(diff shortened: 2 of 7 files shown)


mcglincy (approved):
PR still LGTM.

src/bin/observable.ts
```diff
@@ -123,14 +123,17 @@ try {
...
+          manifest: {type: "boolean", description: "Use the new upload strategy"},
+          "no-manifest": {type: "boolean", description: "Use the old upload strategy"}
         }
       });
       await import("../deploy.js").then(async (deploy) =>
         deploy.deploy({
           config: await readConfig(config, root),
           message,
-          force: build ? "build" : noBuild ? "deploy" : null
+          force: build ? "build" : noBuild ? "deploy" : null,
```
mbostock, line 135: The `--build` and `--no-build` arguments are already consolidated by the `helpArgs` helper. (Apologies for not noticing and fixing this earlier when I implemented this feature.) So, instead of pulling out the `noBuild` value, you can say:

```suggestion
          force: build === true ? "build" : build === false ? "deploy" : null,
```

src/bin/observable.ts
```diff
@@ -123,14 +123,17 @@ try {
           "no-build": {
             type: "boolean",
             description: "Don’t build before deploying; deploy as is"
-          }
+          },
+          manifest: {type: "boolean", description: "Use the new upload strategy"},
+          "no-manifest": {type: "boolean", description: "Use the old upload strategy"}
```
mbostock, line 128: If these are for internal use only, and we plan to remove them before releasing, can you remove the descriptions or change them to “[internal use only]”? And add a “TODO remove me” comment? But I don’t really want to merge this to main if we don’t want to release this feature yet because 1.7 is imminent (and in general I prefer if we only merge to main what’s ready to release).

src/deploy.ts
```diff
@@ -609,3 +660,15 @@ function formatAge(age: number): string {
...
+async function sha512File(path: string): Promise<string> {
+  const hasher = createHash("sha512");
+  const stream = createReadStream(path);
+  const promise = new Promise<void>((resolve, reject) => {
+    hasher.once("finish", resolve);
+    hasher.once("error", reject);
+  });
+  stream.pipe(hasher);
+  await promise;
+  return (hasher.read() as Buffer).toString("base64");
```
mbostock, line 673: How important is it to use streaming here? Can we simplify like so (which is a pattern I use elsewhere):

```suggestion
  return createHash("sha512").update(await readFile(path)).digest("base64");
```

mbostock (changes requested):

src/deploy.ts
mbostock: Alternatively, I feel like this would be cleaner:

```ts
async function sha512File(path: string): Promise<string> {
  return new Promise((resolve, reject) => {
    const hash = createHash("sha512");
    createReadStream(path)
      .pipe(hash)
      .on("error", reject)
      .on("finish", () => resolve(hash.digest("hex")));
  });
}
```

mythmon: I think the main difference is that this is working in parallel, with up to 8 files being read at once. The most constrained place we're likely to be deploying projects is in CI. Github [allocates](https://docs.github.com/en/actions/using-github-hosted-runners/about-larger-runners/about-larger-runners#specifications-for-general-larger-runners) 4GB of memory per core, and we run at most two parallel hashers per core. We can optimize for files no larger than 50MB, so it should be no problem to read the whole file into memory.

src/bin/observable.ts
mythmon: I don't have any concrete reasons to think this isn't ready, I'm just being cautious. In the name of tempo, I removed the flag to disable the new feature, and we can fix it if there is a problem.

src/deploy.ts
mbostock: Both readFile and createReadStream should be parallelizable.

mythmon: Yes, but createReadStream would only keep a fixed window in memory at any given time.  readFile will read the whole file into memory. If we had more memory pressure that could be a problem, but it doesn't matter here.

mcglincy:
Sorry, I realize I clicked approve on the wrong PR... thinking this was the server-side one I was re-reviewing. Apologies for noise.

src/deploy.ts
mbostock: I’m cool with you using the alternative `createReadStream` version I posted above if you prefer it.

mythmon: Since their isn't much memory pressure. I'm happy with the readFile version, it's what I put in my latest commit. It's simple and small enough that I can inline it into the function instead of wanting to put it in another function, which is nice.

PR #1283, opened 2024-04-25
never close the sidebar when the mouse is over it

### LINKED ISSUE #1281: search links aren't clickable on a small screen
On a small screen, the sidebar can be revealed by focussing on the search input (with cmd-K). When you mousedown on a link, that input's onblur triggers and hides the sidebar, and the click is ignored. (There is no bug if you navigate with the keyboard and hit enter, because this is happening during focus.)

Fil:
closes #1281


mbostock:
Do we need `:focus-within`, too? Or instead?

Fil:
Ah yes that's better: it solves the bug on click, and the (same) bug when you try to tab to the search results links.

👍

mbostock (changes requested):
Okay, I’m cool with `:focus-within`. But there are probably a dozen other rules that you’ll simultaneously need to update to make this work consistently, though; see #419 and 5ec8f95595a5637bbe3ec553c8a058937f14d3f9 which did this previously. But we reverted this approach in #581. Also note that Firefox ESR doesn’t support `:has`, so there’s basically no way to do this correctly in Firefox at the moment; see #836.

Fil:
Oh right. I didn't find a dozen, only this one:

```
.observablehq-sidebar-open ~ #observablehq-sidebar-backdrop,
#observablehq-sidebar-toggle:checked ~ #observablehq-sidebar-backdrop {
  display: initial;
}
```

to which we need to add a rule so that the backdrop is present while tabbing (but visually it makes no difference with the default, since the backdrop is transparent).

Am I missing other cases?

mbostock:
Doesn’t everything that currently checks `#observablehq-sidebar-toggle:checked` and `#observablehq-sidebar-toggle:indeterminate` now also need to check `#observablehq-sidebar:focus-within` since now there’s another way that the sidebar can be open?

Fil:
OK… Now I see them!

It was already an oversight in #543 : the` .observablehq-sidebar-open` selector was already missing when on a large screen you had hidden the side bar (with ctrl-B, maybe), and opened it with ctrl-K. The visual difference was about a sliver of border-right. I hope I caught them all this time :)

Fil:
I realized that, since `:focus-within` is true when the bar is opened with ctrl-K, we can simplify this quite a bit. We only need `.observablehq-sidebar-open` to open the side bar so we can focus the input — then it has :focus-within, and this selector is redundant.

src/client/search-init.ts
```diff
@@ -10,22 +10,20 @@ input.addEventListener("focus", load, {once: true});
...
   ) {
     // Force the sidebar to be temporarily open while the search input is
-    // focused. (We can’t use :focus-within because the sidebar isn’t focusable
-    // while it is invisible, and we don’t want to keep the sidebar open
-    // persistently after you blur the search input.)
-    toggle.classList.add("observablehq-sidebar-open");
+    // focused. The class makes the sidebar visible so the input can be focused,
+    // then the :focus-within selector takes over and keeps it open until the
+    // user blurs the input.
+    sidebar.classList.add("observablehq-sidebar-open");
```
mbostock, line 23: Could you do this by temporarily toggling the checkbox? Then you may not need the `observablehq-sidebar-open` class at all?

Fil: yes!

Note: the code I pushed now assumes that we use 'visibility:hidden' to hide the side bar, and uses getComputedStyle to know what state we're in. We could skip this step and click on the toggle in any case (even when it's doing the opposite and starts closing the bar), because thanks to the _transition_ it is still visible. But I felt it fragile to rely on a css transition.

src/client/search-init.ts
```diff
@@ -10,22 +10,21 @@ input.addEventListener("focus", load, {once: true});
...
   ) {
     // Force the sidebar to be temporarily open while the search input is
-    // focused. (We can’t use :focus-within because the sidebar isn’t focusable
-    // while it is invisible, and we don’t want to keep the sidebar open
-    // persistently after you blur the search input.)
-    toggle.classList.add("observablehq-sidebar-open");
+    // focused. We click on the toggle if necessary to make the input visible so
+    // it be focused. Then the :focus-within selector takes over and keeps it
+    // open until the user blurs the input.
+    const check = getComputedStyle(input).visibility === "hidden";
```
mbostock, line 23: Would it be simpler to just look at `toggle.checked` here? As in, force the sidebar open if it’s not already forced open?

Fil: Yes this works too. I'm also tempted to write this like so:

```
    if (!toggle.checked) toggle.click(), input.focus(), toggle.click();
    else input.focus();
```

it duplicates the main action (focus), but it's much clearer in a sense.

mbostock (commented):
I tested this on Firefox ESR 😓 to see what happens when `:has()` is unsupported. It works great even so. The backdrop isn’t visible, but we don’t actually need the backdrop to be visible here because the backdrop is only used to hide the sidebar when you click on the backdrop, and in this state where the sidebar is only visible because it is focused, the sidebar will automatically hide itself when it loses focus by clicking outside the sidebar. So we could even remove the line that makes the backdrop visible when the sidebar is focused — but I think it’s fine to keep it (maybe in the future the backdrop has partial transparency to fade the page content, _e.g._).

I’m going to double-check the logic to make sure I haven’t missed anything and will then approve.

mbostock (changes requested):
Okay, I found a bug in this approach, which is what I alluded to earlier where there are other conditions that we need to check. You can reproduce the bug by setting the window width to ~1091px and then comparing the difference in behavior between focusing the sidebar with Command-K and opening it with Command-B.

https://github.com/observablehq/framework/assets/230541/0f92e729-53c8-44c8-87d7-5e0d7ef6f1fe

When the sidebar is opened with Command-B, the sidebar toggle is checked. This triggers several style rules that affect the sidebar, the center element, and the toc element. You can see these rules by looking for `#observablehq-sidebar-toggle:checked`.

When the sidebar is focused and implicitly opened with Command-K, the sidebar toggle is not checked, so only the rules with `#observablehq-sidebar:focus-within` apply. This leads to the difference in behavior. I think we want the same behavior regardless of how the sidebar is opened (or focused).

One solution to this would be the approach taken in #419 where we add quite a few more `#observablehq-sidebar:focus-within` rules so that focusing the sidebar has the equivalent effect to checking the sidebar toggle. However, this won’t work with Firefox ESR because it will rely on `:has()` — we’ll have to be careful that we don’t break things in Firefox ESR (but also accept that this approach won’t entirely work on Firefox ESR).

The alternative solution would be to abandon the `:focus-within` approach, and simply open the sidebar explicitly when it is focused via Command-K (as if you had hit Command-B). This is perhaps a worse user experience since it means that Command-K followed by Esc will no longer implicitly close the sidebar on small screens, but you can also use Command-B to close the sidebar, so I don’t think that’s a big deal?

Fil:
Oh but I like that the behavior is different, because cmd-B opens the side bar "for ever", whereas cmd-K opens it "temporarily". This difference is reflected naturally with the side bar being fixed in place in the first case (with a rearrangement of the page's structure), and floating over the contents in the second case (similar to a modal that will be dismissed). ctrl-K is faster, and does not provoke recalculations when the contents depends on width.

mbostock:
The good news is that we didn’t actually need anymore `:has()` selectors to implement the additional `:focus-within` rules — that was primarily needed in #419 because the sidebar toggle moved in position when the sidebar is opened or closed, but since #586 we have separate toggles inside and outside the sidebar. I’ve pushed the fix and tested that it works even on Firefox ESR.

mbostock:
> Oh but I like that the behavior is different, because cmd-B opens the side bar "for ever", whereas cmd-K opens it "temporarily".

Yes, I’m aware that you like it. I’m just pointing out that it presents additional challenges. But I’ve figured out how to overcome them.

mbostock:
Maybe you’re right though… maybe the visual distinction is helpful because it communicates the behavior? 🤔

PR #1291, opened 2024-04-28
page stats

### LINKED ISSUE #1243: `observable build` should print a page size summary at the end
We know the size of the page, as well as the size of its transitive static references (files and modules); it’d be great to print out a summary table and warn if the sizes look big. For example, Next.js does something like this:

```
Route (app)                               Size     First Load JS
┌ ○ /                                     5.3 kB         89.5 kB
├ ○ /_not-found                           885 B          85.1 kB
└ ○ /about                                137 B          84.4 kB
+ First Load JS shared by all             84.2 kB
  ├ chunks/184-d3bb186aac44da98.js        28.9 kB
  ├ chunks/30b509c0-f3503c24f98f3936.js   53.4 kB
  └ other shared chunks (total)
 
 
○  (Static)  prerendered as static content
```

mbostock:
Fixes #1243. Supersedes #320.

Screenshot 2024-04-28 at 1 22 16 PM

I find it a little hard to read the outputs since the number of fractional digits changes even within units, and the decimal points aren’t aligned. So, I’d like to make that a bit better. I’m also not counting stylesheets, but those should probably be counted too? And is it confusing that “imports” only refers to JavaScript (when you can also import CSS)?

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,6 +66,7 @@
     "cross-env": "^7.0.3",
     "cross-spawn": "^7.0.3",
     "d3-array": "^3.2.4",
+    "d3-hierarchy": "^3.1.2",
     "esbuild": "^0.20.1",
     "fast-array-diff": "^1.1.0",
     "gray-matter": "^4.0.3",
--- a/src/build.ts
+++ b/src/build.ts
@@ -5,6 +5,7 @@ import {basename, dirname, extname, join} from "node:path/posix";
 import type {Config} from "./config.js";
 import {CliError, isEnoent} from "./error.js";
 import {getClientPath, prepareOutput, visitMarkdownFiles} from "./files.js";
+import {formatByteSize} from "./format.js";
 import {getModuleHash} from "./javascript/module.js";
 import {transpileModule} from "./javascript/transpile.js";
 import type {Logger, Writer} from "./logger.js";
@@ -21,7 +22,7 @@ import {bundleStyles, rollupClient} from "./rollup.js";
 import {searchIndex} from "./search.js";
 import {Telemetry} from "./telemetry.js";
 import {tree} from "./tree.js";
-import {faint, yellow} from "./tty.js";
+import {faint, green, red, yellow} from "./tty.js";
 
 export interface BuildOptions {
   config: Config;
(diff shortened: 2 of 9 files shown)


Fil:
I ran it on pangea (obviously), and got a hunch that it would be nicer if:
- the pages were sorted by (total) weight in each directory, rather than by slug, so I could spot the heaviest pages without having to read through the whole list
- the Imports sizes were computed on the gziped payload. One example I have is a file which transfers 600kB (network wise), but is listed as having 1.96 MB of Imports. This also matters for Files (at least some file types that compress well in general: it's not necessarily a bad thing to have a huge json file if it compresses well, and switching to an apparently smaller csv file might be unnecessary)
- the Files column was switched with the Imports column, since files requests usually come after the imports have been transferred

I also got a slight sense of getting too much information (given the number of pages). A possibility would be to just give the names of the pages that fall under a certain threshold that we consider "fast" and don't need to give details for.

In any case this is a net improvement over the current situation, so I'd take it as it is. But seeing it in action gives more ideas:

For example, I'd love to reuse/repurpose this info somehow to manage my project: to highlight the pages that are at risk to be slow, those that have suddenly ballooned, or to see how this project has evolved across time.

A possibility could be to dump the raw info into a timestamped `file-size-report.json` file in `docs/.observablehq/`, and have a “reporter” script that people can adjust to their needs. A default reporter would run after each build, but reporters could also run independently. We could even have a reporter that creates a hidden page in the project, with visual & interactive analysis on all the available build reports.

I'd use a reporter that summarizes the crucial info, for example by listing the slow outliers: pages that might need some 👀 , or the top pages according to some heuristic, or pages that have "exploded" if a previous report was available in `docs/.observablehq/`.

Also I'm imagining a small chart (box-plot, maybe), showing the distribution of sizes, with a constant frame of reference that allows me to also compare at a glance between builds and across projects.

mbostock:
Here’s what I’d like to do:

- Swap imports and files column
- Print summary below the tree (median and max page size?)
- Show count of pages in each folder (recursively)
- Maybe improve formatting slightly?

I want to keep the alphabetical order (easier to find the page you’re looking for; matches directory listings so likely more familiar order) and not implement gzip size (too much work; possibly slow).

mbostock:
Slightly fancier layout now…

Screenshot 2024-04-28 at 7 55 00 PM

Fil (approved):

PR #1293, opened 2024-04-29
link headings with id

### LINKED ISSUE #1292: Auto-link HTML H2 elements and include them in the TOC
The [Table of contents](https://observablehq.com/framework/config#toc) supports H2 elements. However, only the markdown version `##` gets added. Any HTML `<h2></h2>` get ignored.

mbostock:
Fixes #1292. Instead of relying on markdown-it-anchor to create permalinks, do it in HTML when the heading has an `id` attribute. This means you still have to specify the `id` attribute manually to enable linking (and appear in the TOC), but that seems reasonable to me and gives more control.

### DIFF
--- a/src/client/toc.ts
+++ b/src/client/toc.ts
@@ -2,9 +2,8 @@ const toc = document.querySelector<HTMLElement>("#observablehq-toc");
 if (toc) {
   const highlight = toc.appendChild(document.createElement("div"));
   highlight.classList.add("observablehq-secondary-link-highlight");
-  const headings = Array.from(document.querySelector("#observablehq-main")!.querySelectorAll(toc.dataset.selector!))
-    .filter((e) => e.querySelector("a.observablehq-header-anchor"))
-    .reverse();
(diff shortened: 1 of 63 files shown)


src/client/toc.ts
```diff
@@ -13,7 +12,7 @@ if (toc) {
     // If there’s a location.hash, highlight that if it’s at the top of the viewport.
     if (location.hash) {
       for (const heading of headings) {
-        const hash = heading.querySelector<HTMLAnchorElement>("a[href]")?.hash;
+        const hash = `#${heading.id}`;
```
mbostock, line 15: I wonder if this needs any encoding? 🤔

Fil: I think so — since we compare it with location.hash we should probably use

```suggestion
        const hash =  new URL(`#${heading.id}`, location).hash;
```

mbostock: Added `encodeURI`. 👍

Fil (approved):

PR #1295, opened 2024-04-29
Add an --id flag to deploy for an already-created deploy.

mcglincy:
I still need to verify backend API handling of a deploy skipping postDeploy, specifically around possible errors returned from specifying a non-existent deploy ID, or an ID for a project/deploy the user doesn't have permissions to deploy to.


src/deploy.ts
```diff
@@ -342,24 +343,25 @@ export async function deploy(
...
-      if (error.statusCode === 404) {
-        throw new CliError(`Project @${deployTarget.workspace.login}/${deployTarget.project.slug} not found.`, {
-          cause: error
-        });
-      } else if (error.statusCode === 403) {
-        throw new CliError(
-          `You don't have permission to deploy to @${deployTarget.workspace.login}/${deployTarget.project.slug}.`,
-          {cause: error}
-        );
+  if (!deployId) {
```
mythmon, line 346: This should have an else, and of a deployOd was specified, query the server early to check that it is valid and the user has access. Checking early will enable us to give quicker and more helpful error messages.

mythmon (changes requested):
Looks good, but I think we should hide `--id` from the deploy command's help

src/bin/observable.ts
```diff
@@ -292,6 +298,12 @@ function helpArgs<T extends DescribableParseArgsConfig>(
 
   // Log automatic help.
   if ((result.values as any).help) {
+    // omit hidden flags from help
+    for (const key of Object.keys(options)) {
+      if (options[key].hidden) {
+        delete options[key];
```
mbostock, line 304: This is mutating the passed-in `options`, which might be surprising (and is generally a pattern I try to avoid). Maybe we could use a pattern like this instead?

```js
const publicOptions = Object.fromEntries(Object.entries(options).filter(([, option]) => !option.hidden));
```

And then replace references to `options` with `publicOptions` as appropriate?

mcglincy: fwiw I only mutated because we are process.exit-ing a few lines later, but I agree it could be surprising. Changed!

mythmon (commented):
Looking at this again with fresh eyes, it seems strange that we call go through the possibility of creating a project when `deployId` is passed. I'm not sure how we would get in this state, but if a project doesn't have a `deploy.json` file, we should still be able to deploy with `--id`, without prompting for the project.

What do you think? Is there an edge case I'm missing?

mcglincy:
I think you’re right - a deploy couldn’t be created if a project didn’t
already exist.

src/deploy.ts
```diff
@@ -78,7 +79,7 @@ type DeployTargetInfo =
 
 /** Deploy a project to ObservableHQ */
 export async function deploy(
-  {config, message, force, deployPollInterval = DEPLOY_POLL_INTERVAL_MS, maxConcurrency}: DeployOptions,
+  {config, message, force, deployPollInterval = DEPLOY_POLL_INTERVAL_MS, maxConcurrency, deployId}: DeployOptions,
```
mootari, line 82: > but if a project doesn't have a deploy.json file, we should still be able to deploy with --id, without prompting for the project.

Could we read the deploy config in observable.ts and pass it down to deploy.deploy()? And if we have a deployId, override the config with a stub config containing only the id?

mcglincy:
I am drastically refactoring the deploy() monolith to clearly separate the flows for deploy id provided vs. not. Will push a new commit for that later today.

src/deploy.ts
```diff
@@ -77,31 +78,66 @@ type DeployTargetInfo =
...
-  if (deployConfig.projectSlug && !deployConfig.projectSlug.match(/^[a-z0-9-]+$/)) {
-    throw new CliError(
-      `Found invalid project slug in ${join(config.root, ".observablehq", "deploy.json")}: ${deployConfig.projectSlug}.`
-    );
+  let deployInfo;
+  if (deployOptions.deployId) {
+    deployInfo = await continueExistingDeploy(deployOptions, effects, deployOptions.deployId);
+  } else {
+    deployInfo = await startNewDeploy(deployOptions, effects);
   }
```
mythmon, line 90: The usual style around here is to use a ternary to make it clear that both branches assign a value:

```suggestion
  const deployInfo = deployOptions.deployId 
    ? await continueExistingDeploy(deployOptions, effects, deployOptions.deployId)
    : await startNewDeploy(deployOptions, effects);
```

src/deploy.ts
```diff
@@ -77,31 +78,66 @@ type DeployTargetInfo =
...
-      }.`
-    );
-  }
-  if (deployConfig.projectSlug && !deployConfig.projectSlug.match(/^[a-z0-9-]+$/)) {
-    throw new CliError(
-      `Found invalid project slug in ${join(config.root, ".observablehq", "deploy.json")}: ${deployConfig.projectSlug}.`
-    );
+  let deployInfo;
+  if (deployOptions.deployId) {
+    deployInfo = await continueExistingDeploy(deployOptions, effects, deployOptions.deployId);
```
mythmon, line 87: Why does `deployId` need to be passed separately if it is already in `deployOptions`?

mythmon (commented):
I'm still not convinced this improves the readability of the flow, but I'm happy to go along with it if it helps you. A couple notes we could change, and maybe we want to consider this as a class with methods, so we don't have to pass things around as much? That kind of clashes with the prevailing style though.

src/deploy.ts
mcglincy: it doesn't; I was just taking advantage of the type narrowing from the if check.

src/deploy.ts
```diff
@@ -77,30 +78,62 @@ type DeployTargetInfo =
...
+  Telemetry.record({event: "deploy", step: "finish"});
+}
+
+async function continueExistingDeploy(
+  deployOptions: DeployOptions,
+  effects: DeployEffects
+): Promise<GetDeployResponse> {
+  const {apiClient} = await getApiClientAndCurrentUser(effects);
+
+  const deployId = deployOptions.deployId!;
```
mythmon, line 99: You could keep the type safety that extra parameter was giving you like this:/

```suggestion
async function continueExistingDeploy(
  deployOptions: DeployOptions & {deployId: string},
  effects: DeployEffects
): Promise<GetDeployResponse> {
  const {apiClient} = await getApiClientAndCurrentUser(effects);

  const {deployId} = deployOptions;
```

mcglincy: ah nice, thanks - will switch to that

src/bin/observable.ts
```diff
@@ -292,16 +298,18 @@ function helpArgs<T extends DescribableParseArgsConfig>(
...
+        publicOptions
       )
         .map(([name, {default: def}]) => ` [--${name}${def === undefined ? "" : `=${def}`}]`)
         .join("")}`
     );
-    if (Object.values(options).some((spec) => spec.description)) {
+    if (Object.values(publicOptions).some((spec) => spec.description)) {
       console.log();
-      for (const [long, spec] of Object.entries(options)) {
+      for (const [long, spec] of Object.entries(publicOptions)) {
```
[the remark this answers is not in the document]
mcglincy: This PR has already stretched far beyond what I wanted. Let's punt additional tweaks to followup PRs.

src/deploy.ts
mcglincy: Check out my updated handling of deploy id provided vs. not.

mcglincy:
Created a new Deployer class that handles state, to avoid plumbing deployOptions & effects everywhere.

mythmon (approved):

PR #1296, opened 2024-04-29
deprecate the scripts config option

### LINKED ISSUE #1272: deprecate the scripts option
- [ ] remove it from documentation
- [ ] add a warning to say "instead write the script tag directly in the **head** option

re: #1270

Fil:
I could not find how to inform the user with typescript, since we don't surface types that would be usable in the configuration file.

closes #1272


src/config.ts
```diff
@@ -275,6 +275,9 @@ function normalizeScript(spec: unknown): Script {
   const src = String(script.src);
   const async = script.async === undefined ? false : Boolean(script.async);
   const type = script.type == null ? null : String(script.type);
+  console.warn(
```
Fil, line 278: this shows the warning once for each of the scripts; we could use a different logic to avoid this, but it seemed convenient to squirrel this warning here rather than set up a dedicated function more upstream?

mbostock (approved):

PR #1300, opened 2024-05-01
workaround Safari loading bug

### LINKED ISSUE #1299: Loading indicator may get stuck in Safari
A render bug in Safari (tested in 17.4.1) causes the loading spinner pseudo element to remain visible after the class `observablehq--loading` was removed from the span.

The issue only appears under very specific circumstances. To reproduce, create a page with the following content:
```js
${Plot.plot()}
${Inputs.table([{a: 1}])}
```
Note that the element in the following screenshot no longer animates:
Screenshot 2024-05-02 at 00 29 10

While the content between the two interpolated elements does not appear to matter, the composition does. Neither only two plots or two tables will produce it, nor will a table with only a header.

Other contributing factors are that the pseudo element is displayed inline and animated.

A workaround is to "reset" the element, e.g. via
```css
span[class=""]::before { transform: initial }
```

There is no Webkit issue that I'm aware of, but the bug has been a topic on StackOverflow for years (for example [this one](https://stackoverflow.com/questions/3485365/how-can-i-force-webkit-to-redraw-repaint-to-propagate-style-changes/3485654)).

mbostock:
Fixes #1299. Removing the DOM element seems to avoid the bug where Safari doesn’t detect that the `:before` pseudoelement should be removed. (There are other workarounds, but this one seems to be probably the cleanest alternative approach without having to do a Safari-specific kludge.)

### DIFF
--- a/src/client/main.js
+++ b/src/client/main.js
@@ -27,7 +27,7 @@ export function define(cell) {
   cellsById.get(id)?.variables.forEach((v) => v.delete());
   cellsById.set(id, {cell, variables});
   const root = document.querySelector(`#cell-${id}`);
-  const loading = root.classList.contains("observablehq--loading");
+  const loading = root.querySelector(".observablehq-loading");
   const pending = () => reset(root, loading);
   const rejected = (error) => reject(root, error);
   const v = main.variable({_node: root, pending, rejected}, {shadow: {}}); // _node for visibility promise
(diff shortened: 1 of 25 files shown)


mbostock:
This would be nice to include in 1.7.1 today-ish, if it looks good. Thanks @Fil @mootari!

Fil (approved):

PR #1303, opened 2024-05-03
Snowflake data loader example

mbostock:
This is the first of several Observable Framework examples that demonstrate data loaders for specific data sources. When appropriately configured, it will produce an output like this:

Screenshot 2024-05-03 at 9 41 13 AM

For Snowflake, I’ve included a small wrapper library that makes it easier to manage connections.


examples/loader-snowflake/src/data/snowflake.ts
```diff
@@ -0,0 +1,68 @@
...
+
+async function connect(options: ConnectionOptions): Promise<Connection> {
+  const connection = (snowflake as any).createConnection(options);
+  await new Promise<void>((resolve, reject) => {
+    connection.connect((error) => {
+      if (error) return reject(error);
+      resolve();
+    });
+  });
+  return connection;
```
visnup, line 45: Is it worth creating a connection pool instead of a single connection? https://docs.snowflake.com/en/developer-guide/node-js/nodejs-driver-connect#label-create-connection-pool what's the expected concurrency behavior when people have multiple snowflake data loaders?

mbostock: I don’t think that’s necessary for this example, no. Users can always add that themselves if they want it, and I’d rather keep this example simple. In our experience, most data loaders issue a single query and only use a single connection. So there would be no benefit to having a connection pool. If you have multiple data loaders, then run in separate processes and wouldn’t be able to share a connection pool. Framework currently runs data loaders serially but we might parallelize in the future.

examples/loader-snowflake/package.json
```diff
@@ -0,0 +1,27 @@
...
+    "dev": "observable preview",
+    "deploy": "observable deploy",
+    "observable": "observable"
+  },
+  "dependencies": {
+    "@observablehq/framework": "^1.7.0",
+    "d3-dsv": "^3.0.1",
+    "d3-time": "^3.1.0",
+    "dotenv": "^16.4.5",
+    "snowflake-sdk": "~1.9.0"
```
visnup, line 16: ~1.9.0? is the ~ intentional? also, latest version of snowflake-idk is currently 1.10.1 released 3 weeks ago.

examples/loader-snowflake/src/data/api-requests.csv.ts
```diff
@@ -0,0 +1,21 @@
...
+import {run} from "./snowflake.js";
+
+const end = utcDay();
+const start = utcDay.offset(end, -1);
+
+process.stdout.write(
+  csvFormat(
+    await run((query) =>
+      query(
+        `SELECT normalized_path AS "path", COUNT(*) AS "count"
```
visnup, line 12: I'm a little torn on specifying explicitly lowercase column names here. on one hand it plays better with downstream usage but on the other it can lead to a lot of tedious repetition in sql queries here. I guess I'm just hoping people won't cargo cult it when it might not apply to them.

visnup (approved):

examples/loader-snowflake/package.json
mbostock: Yes. The latest version of the snowflake-sdk generates a `snowflake.log` file by default and emits a bunch of cruft to stdout that breaks data loaders. Also the types in `@types/snowflake-sdk` are wrong because `snowflake-sdk` is distributed as CommonJS so `import * as snowflake from "snowflake-sdk"` doesn’t work.

examples/loader-snowflake/src/data/api-requests.csv.ts
mbostock: The point here is the data loader controls what the output shape is. It shouldn’t be a requirement that the data loader passes through the column names as-is.

mbostock:
Merged into #1306.

PR #1312, opened 2024-05-07
use actions/setup-python

mbostock:
No idea if this will actually work until we test it in CI, but I think we should probably be using [setup-python](https://github.com/actions/setup-python?tab=readme-ov-file) to use Python in GitHub Actions. I believe this will enable better caching and be more representative of how users should do it, too.


allisonhorst (approved):
this sounds & looks good to me 👍

mbostock:
It appears to work! 👍

PR #1321, opened 2024-05-09
Elixir dataloaders

kish10:
Added a default interpreter for ".exs" (Elixir script) files, and an example ".exs" dataloader.


Fil:
Thanks for sharing the example. Closing as I don't think we're going to support this interpreter out of the box for now.

PR #1323, opened 2024-05-10
index examples

Fil:
[This is far from ready—only sharing for discussion.] This indexes the examples together with the rest of the documentation. An issue is that the results are often too high, so before we do this we might want to add a moderator/multiplicating factor to the indexing score, and activate it for examples with a low value (_e.g._ 0.1).

We also need to know what we want to index: currently it is the contents of the page itself, but do we want to include the README.md of examples? This would allow to include the data loader examples. We will have to remove the boilerplate about how to run a project.

See https://github.com/observablehq/framework/pull/1306#issuecomment-2104630355

### DIFF
--- a/docs/index-examples.ts
+++ b/docs/index-examples.ts
@@ -1,0 +1,50 @@
+import {existsSync, readdirSync, statSync} from "node:fs";
+import {readFile, writeFile} from "node:fs/promises";
+import {join} from "node:path/posix";
+import he from "he";
+import {readConfig} from "../src/config.js";
+import {visitMarkdownFiles} from "../src/files.js";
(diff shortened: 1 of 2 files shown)


mbostock:
Maybe we could have two sections of search results, one for our docs, and one for our examples? I’m also thinking that basically all the lib pages and maybe the Inputs pages should be moved to examples too… (and we should add some Plot examples).

Fil:
See also the multi-search in pangea (which seems like a better approach).

mbostock:
Ya, I don’t like that this creates stub pages for the examples and then does client-side redirect. I’d rather have the search link directly to the external example. Maybe e.g. we could have a config option that provides additional results and keywords to include in the search index?

PR #1324, opened 2024-05-10
document npm update

Fil:
For the first part I haven't put in the yarn equivalent, to keep it light. But maybe it's better to be complete?

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -134,6 +134,14 @@ This command will ask you a series of questions in order to initialize your new 
 
 And that’s it! Your new project is ready to go. 🎉
 
+<div class="tip" label="Info">
+
+Under the hood, this command installs the `@observablehq/framework` package and runs:
+
+<pre>npm exec <span class="win">"</span>@observablehq/framework<span class="win">"</span> -- create</pre>
+
+</div>
+
 ## 2. Develop
 
 Next, `cd` into your new project folder.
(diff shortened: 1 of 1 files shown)


docs/getting-started.md
```diff
@@ -134,6 +134,14 @@ This command will ask you a series of questions in order to initialize your new
 
 And that’s it! Your new project is ready to go. 🎉
 
+<div class="tip" label="Info">
```
mbostock, line 137:
```suggestion
<div class="note">
```

I’d rather not introduce a fourth category of note here; I use `note` for “additional technical details” type notes.

docs/getting-started.md
```diff
@@ -550,10 +558,22 @@ json.dump(forecast, sys.stdout)
...
 To write the data loader in R, name it <code>forecast.json.R</code>. Or as shell script, <code>forecast.json.sh</code>. You get the idea. See [Data loaders: Routing](./loaders#routing) for more. The beauty of this approach is that you can leverage the strengths (and libraries) of multiple languages, and still get instant updates in the browser as you develop.
 
-### Deploying automatically
+### Deploy automatically
 
 You can schedule builds and deploy your project automatically on commit, or on a schedule. See [deploying](./deploying) for more details.
 
+### Update Framework
+
+To update Framework to the latest [release](https://github.com/observablehq/framework/releases), cd into the project directory and run:
```
mbostock, line 567:
```suggestion
To update Framework to the [latest release](https://github.com/observablehq/framework/releases), `cd` into the project directory and run:
```

docs/getting-started.md
```diff
@@ -550,10 +558,22 @@ json.dump(forecast, sys.stdout)
...
+
+<pre data-copy>npm update <span class="win">"</span>@observablehq/framework<span class="win">"</span></pre>
+
+or
+
+<pre data-copy>yarn upgrade <span class="win">"</span>@observablehq/framework<span class="win">"</span></pre>
```
mbostock, line 573: Since Framework has a fair number of dependencies, would it be better to have people run a bare `npm update` instead of just updating the `@observablehq/framework` package? I also find that if you just upgrade packages individually, you often get multiple versions of packages, and hence it’s generally better to upgrade everything together.

mbostock (changes requested):

mbostock (approved):

PR #1331, opened 2024-05-13
fix #1330 by limiting to a version of nwsapi that we know works

### LINKED ISSUE #1330: crash due to a regression in nwsapi (“Unexpected error: 'h1:not(:first-of-type),…' is not a valid selector”)
see https://github.com/dperini/nwsapi/issues/114

this breaks the build of any project that has nwsapi@2.2.10 (2.2.9 is OK).

Fil:
(patch by @mootari)

### DIFF
--- a/package.json
+++ b/package.json
@@ -131,5 +131,7 @@
   },
   "engines": {
     "node": ">=18"
-  }
+  },
+  "overrides": { "nwsapi": "<2.2.10" },
+  "resolutions": { "nwsapi": "<2.2.10" }
 }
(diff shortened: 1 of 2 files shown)


mootari:
As far as I'm aware npm only respects `overrides` in the root package.json. The alternative would be to add an [npm-shrinkwrap.json](https://docs.npmjs.com/cli/v10/configuring-npm/npm-shrinkwrap-json), or to resort to packages like [npm-force-resolutions](https://www.npmjs.com/package/npm-force-resolutions) that use a preinstall hook to modify package-lock.json.

mbostock:
Can you explain this?

> the problem is that someone who upgrades an existing project will not get the patch…

How much does this PR help or not help?

mbostock:
> As far as I'm aware npm only respects overrides in the root package.json. 

Right. So we could fix this in the `package.json` in the templates, but adding it here won’t actually help our users? And we still don’t have a solution that helps existing projects?

mbostock:
The `yarn.lock` doesn’t help Yarn users of Framework; it’s only for developing on Framework itself. You’re saying `npm-shrinkwrap.json` helps npm users of Framework, but we have no answer for Yarn users?

mbostock:
I’m guessing the fix here will be to change the version of jsdom that we use, since they seem to be having trouble with their selector implementation too…

https://github.com/jsdom/jsdom/releases

Fil:
Unfortunately, adding this constraint in framework’s package.json does not seem to change what the user gets. What else can we do? Is it possible to pin nwsapi to 2.2.7 explicitly somehow?

mbostock:
#1332

PR #1332, opened 2024-05-13
downgrade jsdom to avoid nwsapi

### LINKED ISSUE #1330: crash due to a regression in nwsapi (“Unexpected error: 'h1:not(:first-of-type),…' is not a valid selector”)

mbostock:
Fixes #1330. Supersedes #1331.

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,7 +73,7 @@
     "highlight.js": "^11.8.0",
     "is-docker": "^3.0.0",
     "is-wsl": "^3.1.0",
-    "jsdom": "^23.0.0",
+    "jsdom": "^23.2.0",
     "jszip": "^3.10.1",
     "markdown-it": "^14.0.0",
     "markdown-it-anchor": "^8.6.7",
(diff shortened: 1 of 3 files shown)


mbostock:
It’d probably be useful for us to run `yarn upgrade` and make sure we’re also testing against the latest dependencies… We were still running against nwsapi 2.2.7 which is why we didn’t notice the problem.

Fil:
LGTM with a caveat: "^23.0.0" accepts both [jsdom@23.1.0](https://github.com/jsdom/jsdom/releases/tag/23.1.0) and [jsdom@23.2.0](https://github.com/jsdom/jsdom/releases/tag/23.2.0), which means we don't guarantee which CSS selector engine is used. To be clear, 23.2.0 is OK, but 23.1.0 is broken (tested broken) — and any previous version is probably broken too because they all use nwsapi.

So I think we'd be better off being explicit about `"jsdom": "=23.2.0"`?

mbostock:
I think we should do `^23.2.0` since that’s pretty standard and if they were to do a patch on 23.x it seems like they would keep the same non-broken selectors implementation.

Fil (approved):

PR #1340, opened 2024-05-14
Generate a build manifest, and include it in deploys to Observable

mythmon:
This PR saves build-time metadata to a file and, if available, uploads it to the platform to power improved UIs and analytics.


src/build.ts
```diff
@@ -276,7 +291,8 @@ export async function build(
       const outputPath = join(dirname(sourceFile), basename(sourceFile, ".md") + ".html");
       const path = join("/", dirname(sourceFile), basename(sourceFile, ".md"));
       const resolveOutput = (name: string) => join(config.output, resolvePath(path, name));
-      const pageSize = (await stat(join(config.output, outputPath))).size;
+      const pageSize = (await maybeStat(join(config.output, outputPath)))?.size;
+      if (!pageSize) continue;
```
mbostock, line 295: Why is this change necessary?

src/build.ts
```diff
@@ -252,15 +254,28 @@ export async function build(
...
     await effects.writeFile(outputPath, html);
+    const url =
+      outputPath === "index.html"
+        ? "/"
+        : basename(outputPath) === "index.html"
+        ? join("/", dirname(outputPath), "/")
+        : extname(outputPath) === ".html"
+        ? join("/", dirname(outputPath), basename(outputPath, ".html"))
+        : outputPath;
+    buildManifest.pages.push({url, title: page.title});
```
mbostock, line 273: Here `url` is actually a path rather than a URL. Can we call this `path` instead? (We use `page.path` elsewhere for a similar purpose.)

src/build.ts
```diff
@@ -252,15 +254,28 @@ export async function build(
...
     const html = await renderPage(page, {...config, path, resolvers});
     await effects.writeFile(outputPath, html);
+    const url =
+      outputPath === "index.html"
+        ? "/"
+        : basename(outputPath) === "index.html"
+        ? join("/", dirname(outputPath), "/")
+        : extname(outputPath) === ".html"
+        ? join("/", dirname(outputPath), basename(outputPath, ".html"))
+        : outputPath;
```
mbostock, line 272: Is there code already in Framework that does this? It looks a lot like `getPathNormalizer` in `config.ts` (though not exactly). Maybe it’s fine to duplicate, just wondering if there’s an opportunity to consolidate.

src/build.ts
```diff
@@ -304,7 +320,7 @@ async function accumulateSize(
     const fileResolution = resolveFile(file);
     if (isAssetPath(fileResolution)) {
       try {
-        size += (await stat(resolveOutput(fileResolution))).size;
+        size += (await maybeStat(resolveOutput(fileResolution)))?.size ?? 0;
```
mbostock, line 323: Same question as above — this seems like an unrelated bug fix/change and would appreciate context?

src/build.ts
```diff
@@ -354,4 +373,13 @@ export class FileBuildEffects implements BuildEffects {
     await prepareOutput(destination);
     await writeFile(destination, contents);
   }
+  async writeBuildManifest(buildManifest: BuildManifest): Promise<void> {
+    const destination = join(this.cacheDir, "_buildManifest.json");
```
mbostock, line 377: Can we call this `_build.json` or `_manifest.json` instead? I’m not a big fan of intercapped and potentially case-sensitive file names.

src/build.ts
mythmon: In a previous version of the tests in this PR, I wasn't writing the files to disk (because it wasn't needed). Since it was using `effect` methods to write the files, but direct filesystem access here, there was a conflict that caused the tests to fail.

I ended up not going down that route though, so I can revert this.

src/build.ts
mythmon: In the future I'd like to also support listing the path of the source file and the generated HTML page, and I'm trying to make clear names for all there. What do you think about `urlPath`? That would suggest something like `sourcePath` and `distPath` in the future.

We don't need to build those feature now, but I don't want to close off the possibility. We likely won't be able to easily change this name in the future.

Or do you have other ideas?

mbostock: Ya, we have a similar need for disambiguation within Framework. I’ve been using (at least in my head) “serve path” for the path that is used to serve the page, so `foo/index.md` would be the “source path” and `foo/` would be the “serve path”. But generally I just refer to this as “path” as this is the public view of the page and thus in my mind more important than the source path…

src/build.ts
```diff
@@ -277,6 +285,7 @@ export async function build(
       const path = join("/", dirname(sourceFile), basename(sourceFile, ".md"));
       const resolveOutput = (name: string) => join(config.output, resolvePath(path, name));
       const pageSize = (await stat(join(config.output, outputPath))).size;
+      if (!pageSize) continue;
```
mbostock, line 288: Revert this change?

mbostock (approved):

src/build.ts
mythmon: `getPathNormalizer` was very nearly what I wanted, including the clean url handling which I didn't have in this version. The only difference was that it produced a path of `"."` for a root index.html. Prefixing `outputPath` with `"/"` fixed that though.

PR #1342, opened 2024-05-14
↗︎ for external sidebar links

mbostock:
Followup to #1306 which allowed external links in the sidebar.


mbostock:
Merging without review since this is relatively minor; let me know if you have comments.

PR #1353, opened 2024-05-15
netcdf example: a bit more explanation about zonal wind

Fil:
belated review of #1351
Capture d’écran 2024-05-15 à 14 08 00


mbostock:
I’ll re-deploy manually after merging.

PR #1361, opened 2024-05-16
typo

Fil:


Fil:
I have not redeployed, but I will once it's merged. (Also, I want us to have a deploy script so we don't have to think about it.)

mbostock:
> I have not redeployed, but I will once it's merged.

Ya, and the problem is the `deploy.json` only lives on my machine so you’ll have to be super careful that you deploy to the right project. ☹️ Maybe just easier for me to do the deploy for now. I’ll do it.

I’m thinking the way we fix this, and the desired for shared configuration across examples, by using new command-line flags to override the config. But, future work.

mbostock (approved):

mbostock:
Deployed to https://observablehq.observablehq.cloud/framework-example-responsive-iframe/.

PR #1363, opened 2024-05-16
review port documentation; add EACCES as an acceptable error 

### LINKED ISSUE #1355: When suggesting `--port` in Getting started, clarify that `npm run` needs `--` to pass additional arguments
When I try to run the dev environment for a project initialized via `npm init "@observablehq"`, it fails with:

```
Unexpected error: listen EACCES: permission denied 127.0.0.1:3000
```

Simply because I have indeed another service running on that port. But there is no apparent way of providing an alternative port.

Official [docs](https://observablehq.com/framework/getting-started#2.-develop) suggest using `--port 3001` but it also fails with:

```
observable: Unexpected argument '3001'. This command does not take positional arguments. See 'observable help preview'.
```

Docs also mention this:

>... the preview server will choose the next available port, so your actual port may vary ...

But I doubt it does such a detection.

The way to do it is also passing an additional `--` so that parameters are passed to the underlying command properly:

```
npm run dev -- --port
```

Fil:
* review port documentation
* add EACCES as an acceptable error to continue port hopping
* avoid out of memory error if no registered port is reachable, and throw an explicit error instead

closes #1355

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -173,7 +173,7 @@ For security, the preview server is by default only accessible on your local mac
 
 Port numbers below 1024 are usually protected and might need admin priviledges. For example, to share your preview server on the default http port:
 
-> `sudo yarn dev --host 0.0.0.0 --port 80`
+> `sudo npm run dev -- --host 0.0.0.0 --port 80`
 
 </div>
 
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -167,7 +167,9 @@ You should see something like this:
 
 If port 3000 is in use, the preview server will choose the next available port, so your actual port may vary. To specify a different port, use <nobr>`--port`</nobr>. For example, to run the preview server with npm on port 4321:
 
-> `npm run dev -- --port 4321`
+```sh
+npm run dev -- --port 4321
+```
 
 For security, the preview server is by default only accessible on your local machine using the [loopback address](https://en.wikipedia.org/wiki/Localhost) 127.0.0.1. To allow remote connections, use <nobr>`--host 0.0.0.0`</nobr>.
 
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -171,7 +171,7 @@ If port 3000 is in use, the preview server will choose the next available port, 
 
 For security, the preview server is by default only accessible on your local machine using the [loopback address](https://en.wikipedia.org/wiki/Localhost) 127.0.0.1. To allow remote connections, use <nobr>`--host 0.0.0.0`</nobr>.
 
-Port numbers below 1024 are usually protected and might need admin priviledges. For example, to share your preview server on the default http port:
+Port numbers below 1024 may need admin privileges. For example, to share your preview server on the default HTTP port 80:
 
 > `sudo npm run dev -- --host 0.0.0.0 --port 80`
 
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -164,8 +164,17 @@ You should see something like this:
 ↳ <u><a href="http://127.0.0.1:3000/" style="color: inherit;">http://127.0.0.1:3000/</a></u></pre>
 
 <div class="note">
-  <p>If port 3000 is in use, the preview server will choose the next available port, so your actual port may vary. To specify port 4321 (and similarly for any other port), use <nobr><code>--port 4321</code></nobr>.</p>
-  <p>For security, the preview server is by default only accessible on your local machine using the <a href="https://en.wikipedia.org/wiki/Localhost">loopback address</a> 127.0.0.1. To allow remote connections, use <nobr><code>--host 0.0.0.0</code></nobr>.</p>
+
+If port 3000 is in use, the preview server will choose the next available port, so your actual port may vary. To specify a different port, use <nobr>`--port`</nobr>. For example, to run the preview server with npm on port 4321:
+
+> `npm run dev -- --port 4321`
+
+For security, the preview server is by default only accessible on your local machine using the [loopback address](https://en.wikipedia.org/wiki/Localhost) 127.0.0.1. To allow remote connections, use <nobr>`--host 0.0.0.0`</nobr>.
+
+Port numbers below 1024 are usually protected and might need admin priviledges. For example, to share your preview server on the default http port:
+
+> `sudo yarn dev --host 0.0.0.0 --port 80`
+
 </div>
 
 Now visit <http://127.0.0.1:3000> in your browser, which should look like:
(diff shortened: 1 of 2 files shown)


mbostock (changes requested):
It’s not clear to me that we want to treat EACCESS like EADDRINUSE. I can’t reproduce the EACCESS error and I don’t suspect that choosing a different port will help.

src/preview.ts
```diff
@@ -77,17 +77,18 @@ export class PreviewServer {
...
+      for (port = 3000; port < 49152; ++port) {
         try {
           await new Promise<void>((resolve, reject) => {
             server.once("error", reject);
             server.listen(port, hostname, resolve);
           });
           break;
         } catch (error) {
-          if (!isSystemError(error) || error.code !== "EADDRINUSE") throw error;
+          if (!isSystemError(error) || (error.code !== "EADDRINUSE" && error.code !== "EACCES")) throw error;
```
Fil, line 88:
```suggestion
          if (!isSystemError(error) || error.code !== "EADDRINUSE") throw error;
```

remove EACCES

src/preview.ts
```diff
@@ -77,7 +77,7 @@ export class PreviewServer {
     Telemetry.record({event: "preview", step: "start"});
     const server = createServer();
     if (port === undefined) {
-      for (port = 3000; true; ++port) {
+      for (port = 3000; port < 49152; ++port) {
```
mbostock, line 80: Can you make a `MAX_PORT` constant for this since it’s in a couple places? And where does it come from — seems like 46,152 attempts might be a bit too many lol?

Fil: I just wanted to avoid the infinite loop and needed a reasonable target. 49152 is the start of the [Dynamic, private or ephemeral ports](https://en.wikipedia.org/wiki/Ephemeral_port) range. (See also https://en.wikipedia.org/wiki/Registered_port).

It's a lot of tests, but it's almost instant even if we need to cover the whole range.

Fil: I added MAX_PORT

src/preview.ts
```diff
@@ -77,7 +77,8 @@ export class PreviewServer {
     Telemetry.record({event: "preview", step: "start"});
     const server = createServer();
     if (port === undefined) {
-      for (port = 3000; true; ++port) {
+      const MAX_PORT = 49152;
```
mbostock, line 80:
```suggestion
      const MAX_PORT = 49152; // https://en.wikipedia.org/wiki/Registered_port
```

docs/getting-started.md
```diff
@@ -164,8 +164,17 @@ You should see something like this:
...
+
+If port 3000 is in use, the preview server will choose the next available port, so your actual port may vary. To specify a different port, use <nobr>`--port`</nobr>. For example, to run the preview server with npm on port 4321:
+
+> `npm run dev -- --port 4321`
+
+For security, the preview server is by default only accessible on your local machine using the [loopback address](https://en.wikipedia.org/wiki/Localhost) 127.0.0.1. To allow remote connections, use <nobr>`--host 0.0.0.0`</nobr>.
+
+Port numbers below 1024 are usually protected and might need admin priviledges. For example, to share your preview server on the default http port:
+
+> `sudo yarn dev --host 0.0.0.0 --port 80`
```
mbostock, line 176:
```suggestion
> `sudo npm run dev -- --host 0.0.0.0 --port 80`
```

I think we should bias towards npm rather than Yarn and be consistent if we’re only showing one.

docs/getting-started.md
```diff
@@ -164,8 +164,17 @@ You should see something like this:
...
-  <p>If port 3000 is in use, the preview server will choose the next available port, so your actual port may vary. To specify port 4321 (and similarly for any other port), use <nobr><code>--port 4321</code></nobr>.</p>
-  <p>For security, the preview server is by default only accessible on your local machine using the <a href="https://en.wikipedia.org/wiki/Localhost">loopback address</a> 127.0.0.1. To allow remote connections, use <nobr><code>--host 0.0.0.0</code></nobr>.</p>
+
+If port 3000 is in use, the preview server will choose the next available port, so your actual port may vary. To specify a different port, use <nobr>`--port`</nobr>. For example, to run the preview server with npm on port 4321:
+
+> `npm run dev -- --port 4321`
+
+For security, the preview server is by default only accessible on your local machine using the [loopback address](https://en.wikipedia.org/wiki/Localhost) 127.0.0.1. To allow remote connections, use <nobr>`--host 0.0.0.0`</nobr>.
+
+Port numbers below 1024 are usually protected and might need admin priviledges. For example, to share your preview server on the default http port:
```
mbostock, line 174:
```suggestion
Port numbers below 1024 may need admin privileges. For example, to share your preview server on the default HTTP port 80:
```

docs/getting-started.md
```diff
@@ -164,8 +164,17 @@ You should see something like this:
 ↳ <u><a href="http://127.0.0.1:3000/" style="color: inherit;">http://127.0.0.1:3000/</a></u></pre>
 
 <div class="note">
-  <p>If port 3000 is in use, the preview server will choose the next available port, so your actual port may vary. To specify port 4321 (and similarly for any other port), use <nobr><code>--port 4321</code></nobr>.</p>
-  <p>For security, the preview server is by default only accessible on your local machine using the <a href="https://en.wikipedia.org/wiki/Localhost">loopback address</a> 127.0.0.1. To allow remote connections, use <nobr><code>--host 0.0.0.0</code></nobr>.</p>
+
+If port 3000 is in use, the preview server will choose the next available port, so your actual port may vary. To specify a different port, use <nobr>`--port`</nobr>. For example, to run the preview server with npm on port 4321:
+
+> `npm run dev -- --port 4321`
```
mbostock, line 170:
````suggestion
```sh
npm run dev -- --port 4321
```
````

**Review by mbostock — APPROVED** [review]

File: src/preview.ts (hunk 2 above)
Comment (mythmon) [reply src/preview.ts]: I'm late here, but how would you feel about letting the OS choose a port when there is a conflict? If you request the port 0 you get a "random" free port.

Comment (Fil) [reply src/preview.ts]: seems interesting (& I didn't know that); we could count up to 3010 or 3100, and then do that trick… not sure how you get back the actual port number though?

Comment (mythmon) [reply src/preview.ts]: After you start listening the `server.address()` method has information about that, though it's a little more complicated than just the port. Here's what we do in project-builder:

```ts
httpServer.listen(config.port, "127.0.0.1", () => {
  const address = httpServer.address();
  if (!address || typeof address === "string")
    throw new Error(`Unexpected server address: ${address}`);
  console.log(`Listening on http://${address.address}:${address.port}/`);
});
```

PR #1365, opened 2024-05-17
netcdf-contours example

mbostock:
https://observablehq.observablehq.cloud/framework-example-netcdf-contours/


Fil:
Two possible alternatives, that need less code (or rather, uses code from different libraries):

1. using Plot.contour (after projection), as in https://observablehq.observablehq.cloud/pangea/party/geotiff
2. using **geoContour** from [d3-geo-voronoi](https://github.com/Fil/d3-geo-voronoi):

```
const polygons = geoContour()
  .x((_, i) => (i % isize) / isize * 360 - 180)
  .y((_, i) => Math.floor(i/isize) / jsize * 180 - 90)
  .value(d => d)
    (reader.getDataVariable("UWND"));
```

There are drawbacks to each of the methods, of course : Plot.contour does not return spherical contours, and geoVoronoi has its own world, and might be slower (but in the case of a data loader, this is not an issue); geoContours have the benefit of being more adapted to spherical data (not making “flames” around the poles).

Fil:
Also if the goal is to minimize the file size, we could apply topojson.simplify & quantize the output. Or maybe this is for a different example?

mbostock:
I did the TopoJSON originally and the difference wasn’t enough to justify the added complexity and dependencies (at least for this explainer).

I’m aware of Plot.contour (of course); the goal of this example is to demonstrate conversion to GeoJSON. I can look at using d3-geo-voronoi if you think it would be simpler.

Fil:
geoContour version https://github.com/observablehq/framework/pull/1368

PR #1369, opened 2024-05-19
ensure case sensitive paths

### LINKED ISSUE #1367: case sensitive file names
Preview is (most often?) case insensitive, but hosting can be case sensitive. A file attachment named AAPL.csv might be correctly resolved in preview and build, but fail in production.

I'd suggest we make preview and build case sensitive, ~~or (maybe better) we lowercase all filenames.~~

Fil:
closes #1367

- [x] fix on windows
- [x] refactor
- [x] test

### DIFF
--- a/src/build.ts
+++ b/src/build.ts
@@ -353,7 +353,9 @@ export class FileBuildEffects implements BuildEffects {
     this.logger.log(destination);
     await prepareOutput(destination);
     await writeFile(destination, contents);
-    const pathcase = (await realpath(destination)).slice((await realpath(this.outputRoot)).length + 1);
+    const pathcase = (await realpath(destination))
+      .slice((await realpath(this.outputRoot)).length + 1)
+      .replaceAll("\\", "/");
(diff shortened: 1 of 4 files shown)


mbostock (changes requested):
Is [`realpath`](https://nodejs.org/api/fs.html#fsrealpathpath-options-callback) the right API to check the case of the file name? `realpath` also resolves symbolic links; if the user is using symbolic links, this could cause a false positive.

Fil:
You're right, this does not work… and it can even create new problems. It comes from a wrong analysis of the root cause.

For example, if I have a single page that contains `` and a horse.jpg file, the build is happy copying the file to `_file/HORSE.xxxx.jpg` and _everything works_. This PR would break that.

The issue in #1367 only arises when I add a _second_ page with ``; in that case, one of the two links (either to _file/horse.xxxx.jpg or to _file/HORSE.xxxx.jpg) will fail in production, because we're only able, on a case-insensitive OS, to produce one of these two files.

So the fix must be completely different: see #1373

PR #1371, opened 2024-05-20
DuckDB loader example

Fil:
This example shows just one of the many ways one would like to use DuckDB as a data loader. It illustrates:

* downloading a raw csv file from Eurostat and keeping a local copy so we can iterate faster on the data loader
* filtering data
* exporting as parquet
* making a chart

Eurostat is a huge beast, but I think this particular dataset is both simple enough to understand, and entertaining enough that you can just look at the chart and learn interesting stuff.


examples/loader-duckdb/src/index.md
```diff
@@ -0,0 +1,157 @@
...
+const languages = FileAttachment("educ_uoe_lang01.parquet").parquet();
+```
+
+We can display this dataset with Inputs.table:
+
+```js echo
+Inputs.table(languages)
+```
+
+We can also make a quick chart of the state of learning of modern foreign languages in European countries using [Observable Plot](https://observablehq/com/plot/); note that the stacked percentages go above 100% in most countries because a lot of pupils learn at least two foreign languages. Ireland (IE) is the only exception (as English is not taught as a foreign language).
```
mbostock, line 53:
```suggestion
We can also make a quick chart of most-frequently taught modern foreign languages in Europe using [Observable Plot](https://observablehq.com/plot/); note that the stacked percentages go above 100% in most countries because many pupils learn at least two foreign languages. Ireland (IE) is the only exception as English is not taught as a foreign language.
```

Did I interpret this right?

examples/loader-duckdb/src/educ_uoe_lang01.parquet.sh
```diff
@@ -0,0 +1,21 @@
...
+  SELECT *
+  FROM read_csv('$TMPDIR/$CODE.csv')
+  WHERE true
+    AND TIME_PERIOD = 2019 -- a good year in terms of data quality
+    AND OBS_VALUE > 0 -- filter out zeros
+    AND isced11 = 'ED2' -- lower secondary education
+    AND unit = 'PC' -- ignore absolute numbers, keep percentages
+    AND language != 'TOTAL' -- ignore total
+    AND length(geo) = 2 -- ignore groupings such as EU_27
+) TO '$TMPDIR/$CODE.parquet' (COMPRESSION gzip);
```
mbostock, line 18:
```suggestion
) TO STDOUT (COMPRESSION gzip);
```

Can we avoid the temporary file by writing directly to stdout?

examples/loader-duckdb/src/educ_uoe_lang01.parquet.sh
```diff
@@ -0,0 +1,21 @@
...
+EOF
+cat $TMPDIR/$CODE.parquet >&1  # Write output to stdout
+rm $TMPDIR/$CODE.csv $TMPDIR/$CODE.parquet  # Clean up
```
mbostock, line 21: Is it necessary to remove the `.csv` from the temporary directory? (We could use the same `-f` test to avoid `gunzip` if it already exists, too…)

examples/loader-duckdb/README.md
```diff
@@ -0,0 +1,9 @@
+[Framework examples →](../)
+
+# DuckDB data loader
+
+View live: <https://observablehq.observablehq.cloud/framework-example-loader-duckdb/>
+
+This Observable Framework example demonstrates how to write a data loader as a bash script that downloads a CSV file from Eurostat (the data portal from the statistical office of the European Union), then calls DuckDB to filter and reshape this data, saving the result as Apache Parquet. The example illustrates the technique by fetching statistics about education to modern foreign languages across Europe.
```
mbostock, line 7:
```suggestion
This Observable Framework example demonstrates how to write a data loader as a shell script that downloads a CSV file, then uses the DuckDB CLI to filter and reshape this data, finally outputting the result as Apache Parquet. The example data is statistics about the education of modern foreign languages from Eurostat, the data portal of the statistical office of the European Union.
```

`sh` and Bash are [not technically the same thing…](https://stackoverflow.com/questions/5725296/difference-between-sh-and-bash) so better to call this a shell script. Also edited for clarity (and we care less about the specific dataset than the demonstrated technique).

examples/loader-duckdb/src/index.md
```diff
@@ -0,0 +1,157 @@
+# DuckDB data loader
+
+Here’s a bash data loader that uses curl to download a CSV file from Eurostat (the data portal from the statistical office of the European Union), then calls [DuckDB](https://duckdb.org/) to filter it and saves the result as [Apache Parquet](https://observablehq.com/framework/lib/arrow#apache-parquet).
```
mbostock, line 3:
```suggestion
Here’s a shell script data loader that uses curl to download a CSV file, then uses the [DuckDB CLI](https://duckdb.org/docs/api/cli/overview.html) to filter it, and finally outputs the result as [Apache Parquet](https://observablehq.com/framework/lib/arrow#apache-parquet).
```

examples/loader-duckdb/src/index.md
```diff
@@ -0,0 +1,157 @@
...
+
+The example illustrates the technique with statistics about education to modern foreign languages across Europe ([educ_uoe_lang01](https://ec.europa.eu/eurostat/databrowser/view/educ_uoe_lang01/default/table?lang=en&category=educ.educ_lang.educ_uoe_lang)).
```
mbostock, line 5:
```suggestion
The example data is statistics about the education of modern foreign languages ([educ_uoe_lang01](https://ec.europa.eu/eurostat/databrowser/view/educ_uoe_lang01/default/table?lang=en&category=educ.educ_lang.educ_uoe_lang)) from Eurostat, the data portal of the statistical office of the European Union.
```

examples/loader-duckdb/src/index.md
```diff
@@ -0,0 +1,157 @@
...
+
+The data loader lives in [`src/educ_uoe_lang01.parquet.sh`](./src/educ_uoe_lang01.parquet.sh).
```
mbostock, line 7:
```suggestion
```

This link won’t work here.

examples/loader-duckdb/src/index.md
```diff
@@ -0,0 +1,157 @@
...
+
+Because the Eurostat API does not compress data as part of the http response, but instead offers to send compressed data, we download it outside of DuckDB — here using [curl](https://curl.se/) — then uncompress it with [gunzip](https://en.wikipedia.org/wiki/Gzip). The raw file is saved to a temporary directory, which makes it much faster to iterate on the data loader when we want to tweak the query, for example to change the education level or the time period considered.
```
mbostock, line 9:
```suggestion
Because the Eurostat API does not compress data as part of the http response, but instead offers to send compressed data, we download it outside of DuckDB — here using [curl](https://curl.se/) — then uncompress it with [gunzip](https://en.wikipedia.org/wiki/Gzip). The raw file is saved to a temporary directory, which makes it much faster to iterate on the data loader when we want to tweak the query, for example to change the education level or the time period considered.
```

My impression is that we’re downloading it outside of DuckDB mostly so that we can cache the downloaded file, rather than downloading it each time we run the data loader? Though the risk of that approach is that we must remember to clear a separate cache directory in `$TMPDIR` rather than the normal one in `.observablehq/cache`. So, can we use the latter instead, so that clearing the cache will re-download the file? I would think that using the existing cache would be a best practice that we should recommend.

examples/loader-duckdb/src/index.md
```diff
@@ -0,0 +1,157 @@
...
+    AND length(geo) = 2 -- ignore groupings such as EU_27
+) TO '$TMPDIR/$CODE.parquet' (COMPRESSION gzip);
+EOF
+cat $TMPDIR/$CODE.parquet >&1  # Write output to stdout
+rm $TMPDIR/$CODE.csv $TMPDIR/$CODE.parquet  # Clean up
+```
+
+<div class="note">
+
+To run this data loader, you’ll need to install the `duckdb` binary, as well as `curl` and `gunzip` if they are not already available on your system.
```
mbostock, line 37:
```suggestion
To run this data loader, you’ll need to install the DuckDB CLI (`duckdb`), as well as `curl` and `gunzip` if they are not already installed on your system.
```

mbostock (changes requested):
A bunch o’ nits, but the most important of which is my recommendation to use `.observablehq/cache` as the cache directory instead of `$TMPDIR`.

examples/loader-duckdb/src/index.md
Fil, line 7:
```suggestion
The data loader lives in [`src/educ_uoe_lang01.parquet.sh`](https://github.com/observablehq/framework/blob/main/examples/loader-duckdb/src/educ_uoe_lang01.parquet.sh).
```

examples/loader-duckdb/src/index.md
Fil: Unfortunately the Eurostat API does not send "csv on an http gziped connection", but instead sends a gzip file ("compressed csv"). I haven't managed to access it directly with DuckDB.

examples/loader-duckdb/src/educ_uoe_lang01.parquet.sh
Fil: No, this errors with `FATAL Error: fsync failed!` (even if you add `FORMAT 'parquet'` in the options). It works if you specify a csv output, but I believe the parquet writer needs random access which STDOUT doesn't provide.

examples/loader-duckdb/src/index.md
Fil: I've adopted `.observablehq/cache`; I'm using a subfolder though, to separate the data loader output from its framework copy (even though it works locally if we cat from a file to stdout, then let framework overwrite the file… it felt dangerous).

examples/loader-duckdb/src/educ_uoe_lang01.parquet.sh
Fil: I've simplified this, thanks for the suggestion

mbostock (approved):
I dropped the `gunzip` to simplify. (Feels like it’s not necessary to include in this example — the dataset is relatively small, and it’s a bit of a Eurostat quirk that I don’t think we need to demonstrate here.)

I also changed the data loader to use the same cache folder as its output. This avoids the need for `mkdir -p` and shouldn’t conflict because the two files have different names (`.csv` and `.parquet`).

I adopted `TO STDOUT` per the recent fix #1386.

I tightened up the wording a bit.

I made the graph a bit prettier, and in particular adopted the use of **limit** to avoid needing `d3.groupSort`. This does mean that the gray “others” isn’t explicitly labeled, but I think that’s fine… maybe we fix that in Plot. 😅 

I also put the tables in a grid which makes it denser. I don’t like hiding things behind details because it entices me to click them just to know what’s inside.

src/dataloader.ts
```diff
@@ -248,10 +248,10 @@ export abstract class Loader {
           else await unlink(errorPath).catch(() => {});
         }
         await prepareOutput(tempPath);
+        await prepareOutput(cachePath);
```
mbostock, line 251: This ensures that the cache folder exists before the data loader runs (so no `mkdir -p` needed).

Fil:
Issue to display "unknown" in the legend:
https://github.com/observablehq/plot/issues/651

PR #1372, opened 2024-05-20
Support remote and local assets in custom CSS

### LINKED ISSUE #786: esbuild crashes when style.css contains @font-face definitions

Fil:
This includes images and fonts using url().

Using onResolve (as suggested in https://github.com/observablehq/framework/issues/786#issuecomment-2070227596)

closes #786
supersedes #788

Capture d’écran 2024-05-20 à 19 58 41

The test above uses a remote @font-face and a local background asset (horse.jpg); but any file type is supported (so you can finally have the custom cursors you've been dreaming of).


src/rollup.ts
```diff
@@ -36,16 +37,33 @@ function rewriteInputsNamespace(code: string) {
...
+  files?: Set<string>;
+  aliases?: Map<string, string>;
 }): Promise<string> {
+  const assets = {
+    name: "resolve CSS assets",
+    setup(build: PluginBuild) {
+      build.onResolve({filter: /^\w+:\/\//}, (args) => ({path: args.path, external: true}));
+      build.onResolve({filter: /./}, (args) => {
+        if (args.path.endsWith(".css") || args.path.match(/^[#.]/)) return;
+        if (files) files.add(args.path); // /!\ modifies files as a side effect
```
mbostock, line 56: Modifying passed-in arguments again. It would be better to return another result with the assets.

src/build.ts
```diff
@@ -92,6 +92,7 @@ export async function build(
 
   // For cache-breaking we rename most assets to include content hashes.
   const aliases = new Map<string, string>();
+  const plainaliases = new Map<string, string>();
```
Fil, line 95: Another style question: I don't really like this `plainaliases` Map, but I didn't like either the idea of mixing the keys for aliases (with a ?hash) and plain aliases (without the ?hash) in the same `aliases` Map.

src/rollup.ts
```diff
@@ -36,21 +37,38 @@ function rewriteInputsNamespace(code: string) {
...
+}): Promise<{contents: string; files: Set<string>}> {
+  const files = new Set<string>();
+  const assets = {
+    name: "resolve CSS assets",
+    setup(build: PluginBuild) {
+      build.onResolve({filter: /^\w+:\/\//}, (args) => ({path: args.path, external: true}));
+      build.onResolve({filter: /./}, (args) => {
+        if (args.path.endsWith(".css") || args.path.match(/^[#.]/)) return;
+        files.add(args.path);
+        const path = join("..", aliases?.get(args.path) ?? join("_file", args.path));
```
mbostock, line 56: Another big advantage of the `resolve` hook is that `bundleStyles` won’t have to know about the `_file` directory — that logic can be supplied by the build command.

And for the preview command, ideally we’d supply a `resolve` hook that does `?sha=…` so that module replacement works when you update a file referenced by a local stylesheet?

mbostock (changes requested):
Nice, this is a useful addition!

Some nits, plus I think there is some plumbing missing: `getResolvers` should now resolve the local assets that are referenced by local stylesheets. Then I don’t think you’ll need to accumulate the files separately during build, since `getResolvers` will have done it already and. And that would mean that we’d also now be watching these referenced files during preview, so we’d get hot replacement if you edit a local file referenced by a local stylesheet. (Though we’ll also need to add `?sha…` for references to local files from local stylesheets as mentioned in the comments.)

src/build.ts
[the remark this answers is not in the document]
Fil: Ah yes, this is much better.

mbostock (commented):
Let me know if you want help moving the logic into `getResolvers` (or maybe this gets put on hiatus for a bit while we focus on other things).

Fil:
I think it's close to done? (I'm not sure what you have in mind so that it can be finished.)

mbostock:
Have you pushed your changes? I don’t see any changes to `getResolvers` here.

Fil:
I changed `bundleStyles` to accept a `resolve` hook, to which I pass a function that uses `getResolvers` — but I don't understand what needs to be changed in `getResolvers`.

mbostock:
We need to parse the stylesheets and resolve/collect all the transitive dependencies. We can pair on it sometime.

Fil:
I think it works already, see for example https://github.com/observablehq/framework/pull/1372/files#diff-8adbd6b9683bb1d5268aa477247f4c6edf9ea93e5b13397535ff450b4d57ae38
`style.css` loads `atkinson.css` which loads the font. (To demonstrate further I should maybe add a local resource to atkinson.css — but I've tried and it works too.) Unless I'm missing something, the one thing that doesn't work is watching and HMR.

mbostock:
I’m aware it works, but I think we should implement it by moving it into `getResolvers` because that’s where it belongs. This will fix the watching/hot replacement issue.

mbostock:
I think there’s probably a lot more that we want to do here, and I haven’t figured out yet if it makes sense to try to do it all together or if there’s a more piecemeal approach. Specifically, what about inline styles like this?

```html
<style type="text/css">

body {
  background-image: url(./image.png);
}

</style>
```

We could detect the reference to `./image.png` above and rewrite the CSS to reference the file in `_file`. (Note that we couldn’t support `body.style.backgroundImage = "url(./image.png)"` in JavaScript because it’s not statically analyzable, but you could use a FileAttachment.)

Similarly, given this:

```html
<link rel="stylesheet" href="./test.css" type="text/css">
```

And the following stylesheet:

```css
body {
  background-image: url(./image.png);
}
```

Presumably we’d want to rewrite `test.css` to reference `./image.png` in `_file` too.

And then there’s the question of whether `test.css` should be bundled into a single stylesheet, or if we should only rewrite the stylesheets to correct for references to local assets?

I want to think about this some more.

shadoof:
> Been trying to import self-hosted font assets and stumbled upon this PR. It would be great to be able to define `@font-face` indeed!

Allowing @font-face to import font files that are local to a project would be good practice for possible archiving.

mbostock:
I realized we didn’t have a tracking issue for this, only the font-face crash bug #786, so I filed #1573.

jamesking:
I ran into this issue yesterday and I came up with a different solution. I decided to host the web-fonts on a dedicated sub-domain of my main Observable Framework website. As a bonus, I ended up with a place to collect interesting open source typefaces. Explained more here: https://jamesking.io/2025/06/09/font-collection

PR #1375, opened 2024-05-21
adopt and document the new apache-arrow date encoding

Fil:
refs:
- https://talk.observablehq.com/t/date-values/9294
- https://github.com/apache/arrow/pull/40960

### DIFF
--- a/docs/lib/arrow.md
+++ b/docs/lib/arrow.md
@@ -60,12 +60,19 @@ Visualized with [Plot’s difference mark](https://observablehq.com/plot/marks/d
 
 ```js echo
 Plot.plot({
+  x: {type: "utc"},
   marks: [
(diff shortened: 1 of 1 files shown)


mbostock (approved):
This fixes the Arrow page, but it doesn’t fix the DuckDB page which uses SQL cells.

Screenshot 2024-05-21 at 9 11 40 AM

And, it’s not possible to workaround it by passing additional hints to Inputs.table since you can’t configure how tables are displayed in SQL cells (currently). And also, Inputs.table should just work as-is with Apache Arrow tables; users shouldn’t have to apply a workaround for something so common. We embrace the Arrow table representation for tabular data and it makes sense for Inputs.table to be aware of it and present data accordingly. And ideally Plot, too, in the future.

PR #1386, opened 2024-05-22
fix spawn stdout

mbostock:
Fixes https://github.com/duckdb/duckdb/issues/10849.

Test data loader:

```sh
duckdb :memory: <<EOF
copy (select * from range(10) tbl(i)) to '/dev/stdout' (format parquet, codec zstd)
EOF
```


mythmon (approved):
This fixes the problem for me on Linux. Thanks for the quick fix!

Fil:
awesome!

PR #1389, opened 2024-05-23
Allow specifying a different path for deploy.json

mythmon:
Add a --deploy-config option to deploy, to allow specifying alternate paths for `deploy.json`. This can help support our examples, as well as users that want to maintain multiple deploys, such as stage and prod projects.


docs/deploying.md
```diff
@@ -121,6 +121,18 @@ This uses one cache per calendar day (in the `America/Los_Angeles` time zone). I
 
 <div class="note">You’ll need to edit the paths above if you’ve configured a source root other than <code>src</code>.</div>
 
+## Deploying to multiple targets
```
mbostock, line 124: This framing feels like it’s assuming a particular use case, but all we’re talking about here is controlling where the `deploy.json` file lives. It was already possible to deploy to “multiple targets” if you didn’t mind overwriting your `deploy.json` file. I think it’s reasonable to describe the staging use case as an example, but I think this section should be called “Deploy configuration” and we should move it up above the **Automated deploys** section and combine it with the existing note (“The deploy command creates a file at .observablehq/deploy.json under the source root…”).

I’ll push some commits to this branch.

src/observableApiConfig.ts
```diff
@@ -68,9 +68,11 @@ export async function setObservableApiKey(info: null | {id: string; key: string}
 
 export async function getDeployConfig(
   sourceRoot: string,
+  requestedDeployConfigPath: string | null = null,
   effects: ConfigEffects = defaultEffects
 ): Promise<DeployConfig> {
-  const deployConfigPath = op.join(effects.cwd(), sourceRoot, ".observablehq", "deploy.json");
+  const deployConfigPath =
+    requestedDeployConfigPath ?? op.join(effects.cwd(), sourceRoot, ".observablehq", "deploy.json");
```
mbostock, line 75: I have a slight preference for only applying the default when `requestedDeployConfigPath` is undefined, rather than treating null and undefined as the same. A reasonable alternative interpretation of the null config path is that we want to force the deploy not use a config (neither read nor write); we don’t support that now but perhaps we might want to reserve it for future use.

Also, the `requestedDeployConfigPath` feels a bit verbose ([enterprise-y](https://projects.haykranen.nl/java/)). I think it’d be reasonable to just call it `deployConfigPath` and have a default value assigned in-place rather than having separate variables.

src/deploy.ts
```diff
@@ -149,7 +168,7 @@ export async function deploy(
       if (project) {
         deployConfig.projectSlug = project.slug;
         deployConfig.workspaceLogin = workspace.login;
-        effects.setDeployConfig(config.root, deployConfig);
+        effects.setDeployConfig(config.root, deployConfigPath, deployConfig, effects);
```
mbostock, line 171: This is missing an `await` (pre-existing bug).

test/deploy-test.ts
```diff
@@ -160,7 +170,8 @@ const TEST_OPTIONS: DeployOptions = {
   config: TEST_CONFIG,
   message: undefined,
   deployPollInterval: 0,
-  force: "deploy" // default to not re-building and just deploying output as-is
+  force: "deploy", // default to not re-building and just deploying output as-is
+  deployConfigPath: null
```
mbostock, line 174:
```suggestion
  deployConfigPath: undefined
```

Per earlier comments. We want the default value here, whereas null implies we want no value.

mbostock (changes requested):
This’ll be great — I’d like to find a way to use this for the examples so we can check-in the `deploy.json` without them needing to live in the example directory.

src/observableApiConfig.ts
```diff
@@ -91,14 +93,15 @@ export async function getDeployConfig(
...
   sourceRoot: string,
+  requestedDeployConfigPath: string | null = null,
   newConfig: DeployConfig,
   effects: ConfigEffects = defaultEffects
 ): Promise<void> {
-  const dir = op.join(effects.cwd(), sourceRoot, ".observablehq");
-  const deployConfigPath = op.join(dir, "deploy.json");
+  const deployConfigPath =
+    requestedDeployConfigPath ?? op.join(effects.cwd(), sourceRoot, ".observablehq", "deploy.json");
   const oldConfig = (await getDeployConfig(sourceRoot)) || {};
```
mbostock, line 102: Given that we already materialized the default `deployConfigPath` at this point, it’s a bit dangerous to not pass it through to `getDeployConfig` — we should pass it in to avoid `getDeployConfig` again computing the default. (A bit paranoid but a good practice.) I’ll make this change.

mbostock, line 102: Also the `|| {}` has no effect? `getDeployConfig` can’t return a falsey value.

mythmon: Yeah, I'm not sure what that was about. Blame says it was added six months ago, maybe that was from the previous version of the deploy configuration? I'll remove it.

mbostock (approved):

mbostock:
If this LGTY it LGTM. 👍

Looks like you’ll have a bit of a chore merging against #1295 though. 😬

src/observableApiConfig.ts
mbostock: Ya I imagine at some point `getDeployConfig` returned `null` if there was no existing config, rather than materializing the default config. Harmless but fixed now though!

mythmon:
I tried to merge this with `git merge`, and like you predicted it was looking pretty difficult to work with. Instead I reverted `deploy.ts` back to it's state on main, and manually applied all the changes from this branch to the right logical places. Much easier.

mythmon:
Thanks for the updates, and especially the doc changes.

PR #1392, opened 2024-05-23
Allow specifying multiple configuration files, use it to simplify example configurations

mythmon:
I implemented the layered config idea I sketched out in https://github.com/observablehq/framework/pull/1390#issuecomment-2127933059. This allows the `--config` option to be passed multiple times. The passed configs are merged together before normalization to produce a final configuration. Configuration files can also be functions that take the current config spec as input.

With that done, I used it to simplify the examples. I removed all the common boiler plate that was marked as "feel free to remove this" and moved it to a single file at `examples/common/confing.js`. To build the examples for deploy now we would run (from the specific example's directory).

```sh
yarn build --config ./observablehq.config.js --config ../common/config.js
```

This approach would replace #1390 and #1389 if we provide a way to specify deploy info or deploy config paths in the main config.


examples/common/config.js
```diff
@@ -0,0 +1,58 @@
...
+    </svg>
+  </a>
+  <div style="display: flex; flex-grow: 1; justify-content: space-between; align-items: baseline;">
+    <a href="https://observablehq.com/framework/" target="_self" rel="">
+      <span class="hide-if-small">Observable</span> Framework
+    </a>
+    <span style="display: flex; align-items: baseline; gap: 0.5rem; font-size: 14px;">
+      <a target="_blank" href="https://github.com/observablehq/framework/tree/main/examples/api"><span>View source</span></a>
```
mbostock, line 54: This will need to be tailored to each example. (Here it references `api`.)

Fil: I think I fixed this.

examples/api/observablehq.config.js
```diff
@@ -1,61 +1,3 @@
...
-  <div style="display: flex; flex-grow: 1; justify-content: space-between; align-items: baseline;">
-    <a href="https://observablehq.com/framework/" target="_self" rel="">
-      <span class="hide-if-small">Observable</span> Framework
-    </a>
-    <span style="display: flex; align-items: baseline; gap: 0.5rem; font-size: 14px;">
-      <a target="_blank" href="https://github.com/observablehq/framework/tree/main/examples/api"><span>View source</span></a>
-    </span>
-  </div>
-</div>`
+  root: "src"
```
Fil, line 2: Here and in the other files I'd like to add back all the configuration options as commented-out values, similar to https://github.com/observablehq/framework/blob/main/templates/default/observablehq.config.js.tmpl

(I can do it if everyone is OK.)

examples/common/config.js
mbostock, line 1: Can we put this in `examples/observablehq.config.js`? (I don’t want people clicking on `examples/common` and thinking that it’s an example.)

mbostock (commented):
Here is my alternative (which doesn’t require any changes to Framework, and instead imports and augments the config in the current working directory): #1396.

PR #1393, opened 2024-05-24
suggestions for layered config

Fil:
Merging the result of the function with its argument feels surprising to me. I'm suggesting this alternative, where we only take the result of the function (the function can do `{...config, title: "new title"}` if that is what the user wants).

Also, I'm suggesting to await, so the exported function could be async? (Or, we wait for a real-world use case before we do that?)


Fil:
superseded by #1396

PR #1396, opened 2024-05-24
layered config

mbostock:
Alternative to #1392 which doesn’t require any Framework changes.


mythmon:
I see! There is a bit of nuance here I wasn't appreciating before. To build the examples this way you'd do something this, right?

```sh
$ cd examples/responsive-iframe
$ npm run build -- --config ../observablehq.config.js
```

The one small gap I can see in this is that a fully general version would have to duplicate the JS/TS resolution logic, but that's not too bad.

mbostock:
Yep, that’s right. Though we typically use Yarn.

```sh
yarn build --config ../observablehq.config.js
```

We don’t use a TypeScript config in any of the examples (yet), but even so, I don’t think we need to worry about the TS/JS resolution because TSX would automatically find the `observablehq.config.ts` if you try to import `observablehq.config.js` (assuming that you registered TSX as a loader, which doesn’t happen by default). 

The automatic config watching also only watches the main config file, not the imported one. That’s fixable using [`node --watch`](https://nodejs.org/api/cli.html#--watch), though.

I think both of these limitations are acceptable.

mbostock:
This is ready. 👍

Fil (approved):

PR #1398, opened 2024-05-27
Use clack.log, and a smaller message

Fil:
#1340 logs the contents of the new build manifest

Capture d’écran 2024-05-27 à 08 57 40

### DIFF
--- a/src/deploy.ts
+++ b/src/deploy.ts
@@ -441,6 +441,7 @@ export async function deploy(
       clack.log.warn(`Could not read build manifest: ${error}`);
     }
   }
+  clack.log.info("Uploading build manifest...");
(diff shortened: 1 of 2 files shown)


mythmon (approved):
We can probably skip the log entirely, it isn't all that interesting to the user.

Fil:
okay, removed

src/deploy.ts
```diff
@@ -441,6 +441,7 @@ export async function deploy(
       clack.log.warn(`Could not read build manifest: ${error}`);
     }
   }
+  clack.log.info("Uploading build manifest...");
```
Fil, line 444:
```suggestion
```

skip log altogether

PR #1399, opened 2024-05-29
no poll delay for deploy tests

mbostock:
The `deployPollInterval` was set to zero, but the `deploy` command wasn’t respecting this due to a logical error (`||`). Also, the `deployPollInterval` needs to be passed through to the `loginInner` method’s `pollTime` option to avoid polling there, too. If desired, we could make this latter polling interval a separate option for the deploy command, but it seems reasonable to have the same option apply to both. (And maybe in the future we could avoid polling entirely with server-sent events or some such?)

On my computer, this makes the tests take 4.9 seconds, down from 15.5 seconds.

### DIFF
--- a/src/deploy.ts
+++ b/src/deploy.ts
@@ -145,7 +145,7 @@ class Deployer {
       if (this.effects.clack.isCancel(choice) || !choice)
         throw new CliError("User canceled deploy", {print: false, exitCode: 0});
 
-      ({currentUser, apiKey} = await loginInner(this.effects));
+      ({currentUser, apiKey} = await loginInner(this.effects, {pollTime: this.deployOptions.deployPollInterval}));
       apiClient.setApiKey(apiKey);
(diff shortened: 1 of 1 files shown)


src/deploy.ts
```diff
@@ -145,7 +145,7 @@ class Deployer {
       if (this.effects.clack.isCancel(choice) || !choice)
         throw new CliError("User canceled deploy", {print: false, exitCode: 0});
 
-      ({currentUser, apiKey} = await loginInner(this.effects));
+      ({currentUser, apiKey} = await loginInner(this.effects, {pollTime: this.deployOptions.deployPollInterval}));
```
mythmon, line 148: Happy to take this change, but I'm pretty sure it is an error that the test is even hitting this line. I'll look into it more.

mythmon (approved):

PR #1401, opened 2024-05-29
index examples

mbostock:
Alternative to #1323.

Screenshot 2024-05-29 at 11 37 27 AM


Fil (approved):
LGTM.

To note:
* this makes the MiniSearch index part of our API—if we switch to some other system it will be a breaking change
* the onus is on the user to properly normalize their texts

mbostock:
I think this needs some more tweaking? See this search for [data loaders].

Screenshot 2024-05-29 at 12 00 35 PM

Should we demote examples relative to the docs pages? The `lib/` and `inputs/` pages feel like they should be the same priority as examples, but the core pages should probably be higher.

Also needs documentation for the new **search** config.

Fil:
I think Minisearch doesn't have per-document boosting, only per field. But in the client we could sort based on the score times an additional "boost" field (that defaults to 1).

In the future a user could add this boost value to a page's front-matter, maybe.

(Half-related, we've also mentioned that maybe keywords should be boosted less than titles, e.g. 2 instead of 4, but I didn't experiment with that).

mbostock:
I found you can do per-document boosting with the **boostDocument** option, so I’ve added that.

Fil:
Yes I prefer this approach to exposing MiniSearch. You can replace an existing document by providing the same id (and if you want to delete it, replace it with an empty text—or better yet use the search: false option).

There is a bug in that the usual results now have a double slash  (e.g., `https://observablehq.com/framework//loaders`).

PR #1403, opened 2024-05-29
fix search result url

Fil:
I think the difference is that we indexed an id that didn't start with a slash, and now a path that does start with a slash.

the buglet

An alternative would be to go back to indexing a path without the leading slash?

fixes https://github.com/observablehq/framework/pull/1401#issuecomment-2138405008


mbostock:
Yes, I cowboy’d the same change in f80bbab83f258fc621a2d4689ec01d3e040069d7. 🤠

I wanted to keep the leading slash since we use that elsewhere to represent paths.

PR #1404, opened 2024-05-29
mod-enter to open in new tab

mbostock:
This allows the Mod-Enter keyboard shortcut to open search results in a new tab, making it much easier browse through the results (without needing to restore the search query as in #897 — though we could still do that) as long as you don’t mind having more than one tab open.


Fil (approved):
this brings keyboard navigation on par with cmd-click, so :+1: 

(I still see the need for #897 ;-) )

Issue #636, opened 2024-01-31
Backslashes aren’t unescaped within PRE elements


mbostock:
Consider the following HTML:

```html
<pre>${test}</pre>
```

The `${test}` is interpreted as a live expression, so if `test` is not defined, you’ll see an error:

```
RuntimeError: test is not defined
```

You can prevent the live expression using a backslash:

```html
<pre>$\{test}</pre>
```

However, the resulting backslash is still rendered in the output, producing:

```
$\{test}
```

A workaround is to use `&dollar;` instead:

```html
<pre>&dollar;{test}</pre>
```

This is probably something we should fix in #597.

Fil:
There is also `<pre>$&zwj;{test}</pre>` (I know, not a very helpful comment!)

PR #1416, opened 2024-06-01
no wrapper span

### LINKED ISSUE #11: Inline expressions shouldn’t use a wrapper span
In a case like this…

```md
<div class="grid grid-cols-4" style="grid-auto-rows: 86px;">
  ${modes.map(mode => BigNumber(events[0][mode], {title: mode}))}
</div>
```

we really want to avoid the wrapper SPAN element so that each interpolated big number is a direct descending of the grid container. Maybe we can do this using document fragments? And comment nodes to denote where the fragment starts and ends?

### LINKED ISSUE #1414: Incremental update fails when the body contains top-level text nodes
The preview server assumes that Markdown will always generate top-level elements (_e.g._, wrapping text in paragraph elements). Unfortunately, that’s not true in some cases like this:

```md
<h2>hello</h2> world
```

In this case, ` world ` becomes a top-level next node. Which means that changing the contents won’t update during live preview:

```md
<h2>hello</h2> other
```

There are two ways we could fix this. One way would be to require that all top-level children are elements by wrapping any top-level text in SPAN elements:

```html
<h2>hello</h2><span> other </span>
```

The other way would be to diff `document.body.childNodes` instead of `document.body.children`, which would mean being able to diff text and comment nodes, not just elements. 🤔

### LINKED ISSUE #1415: Placeholders do not render correctly into SVG
For example, this

```html
<svg width=640 height=120>
  <text x=20 y=20>${"hello"}</text>
</svg>
```

is interpreted as

```html
<svg width=640 height=120>
  <text x=20 y=20></text>
</svg><span>hello</span>
```

because a SPAN element is not allowed to live within an SVG element.

mbostock:
Rather than producing an inline span,

```html
<span id="cell-6212702c"></span>
```

inline expressions now produce a special comment:

```html
<!--:6212702c:-->
```

This comment does not serve as a _container_ for any rendered content but rather serves as a _marker_: rendered content (calls to `display`) are inserted before the comment. The rendered contents are tracked in the `_nodes` array and removed as needed (rather than clearing the container via `innerHTML`).

This eliminates the need for a wrapper span for inline expressions, which means things like this now work:

```md
<div class="grid grid-cols-4">
  ${d3.range(10).map((i) => html`<div class="card">${i}</div>`)}
</div>
```

This also fixes rendering inline expressions into SVG and some edge cases with incremental update during preview.

Fixes #11.
Fixes #1414.
Fixes #1415.

Adding in #1425:
Fixes https://github.com/observablehq/framework/issues/375.
Fixes https://github.com/observablehq/framework/issues/396.
Fixes https://github.com/observablehq/framework/pull/597.
Fixes https://github.com/observablehq/framework/issues/636.


mbostock:
I don’t have a great way of automating tests for this functionality, but here are some of the new things you can do:

```html
<div class="flexy">
  ${d3.range(10).map((i) => html`<div>${i}</div>`)}
</div>

<style type="text/css">

.flexy {
  display: flex;
}

.flexy > * {
  flex-grow: 1;
}

</style>
```

```html
<table>
  <thead>
    <tr>
      <td>i</td>
      <td>one</td>
      <td>two</td>
    </tr>
  </thead>
  <tbody>
    ${d3.range(10).map((i) => html`<tr>
      <td>${i}</td>
      <td>one</td>
      <td>two</td>
    </tr>`)}
  </tbody>
</table>
```

```html
<h2>hello</h2> editme
```

```html
<svg width=640 height=120 fill=red>
  <text x=20 y=20>${now}</text>
</svg>
```

```html
<svg width=640 height=120>
  ${d3.range(10).map((i) => svg`<text x=${20 + i * 10} y=20 fill=${d3.interpolateTurbo(i / 10)}>${i}</text>`)}
</svg>
```

Fil:
Nice! This also _almost_ fixes #375:

* a commented out js fenced code block is ignored.
* a comment with inline code `<!-- ${1 + 2} -->` shows `-->` (instead of crashing)

mbostock:
Fixing requires fundamental changes to our parser as described in #597. I don’t think we should tie these two things together. Interpolating within comments is broken and will remain broken in this PR.

PR #1423, opened 2024-06-03
Technique example: data loader, R to ZIP

allisonhorst:
Deployed: https://observablehq.observablehq.cloud/framework-example-loader-r-to-zip/


Fil:
The error in build comes from using this two-step strategy to load the file:

```
const modelZip = FileAttachment("data/penguin-mlr.zip").zip();
const modelEstimates = modelZip.file("estimates.csv").csv({typed: true});
```

Framework is not smart enough here to know that it will need d3-dsv, and does not include it by default. (This is arguably a bug in Framework, in the sense that it builds an incomplete stdlib—maybe we should preventively build all the methods that stdlib might need?)

The most natural way to solve this for the example page is to rewrite that paragraph as:

~~~~diff
-You can then access individual files (estimates.csv, or predictions.csv) from the ZIP archive:
+Or access individual files (`estimates.csv`, or `predictions.csv`):

 ```js echo
-const modelEstimates = modelZip.file("estimates.csv").csv({typed: true});
+const modelEstimates = FileAttachment("data/penguin-mlr/estimates.csv").csv({typed: true});
 ```
~~~~

which I think conveys the intended use of zip files better.



Another way to fix is to make an explicit import of d3-dsv, even an empty one:

~~~~
```js
import {} from "npm:d3-dsv"
```
~~~~

but it does feel weird.

mbostock:
You can shorten that to

```js
import "npm:d3-dsv";
```

but you should use the automatic unpacking of the .csv and not load the .zip in the client anyway. You should only load a .zip in the client if you want to do something “dynamic” with the contents (in which case you can’t benefit from static analysis).

allisonhorst:
@Fil @mbostock thank you! I've updated this to only load in the individual files (no longer loading the ZIP in the client).

examples/loader-r-to-zip/src/data/penguin-mlr.zip.R
```diff
@@ -0,0 +1,27 @@
...
+        residual = penguins_mlr$residuals
+    )
+
+# Write the data frames as CSVs to a temporary directory
+setwd(tempdir())
+write_csv(mlr_est, "estimates.csv")
+write_csv(mlr_fit, "predictions.csv")
+
+# Zip the contents of the temporary directory
+system("zip - -r .")
```
Fil, line 27:
```suggestion
system("zip - -r .")

```

examples/loader-r-to-zip/src/index.md
```diff
@@ -0,0 +1,66 @@
+# R data loader to generate a ZIP archive
+
+The data loader below reads in the [penguins data](https://journal.r-project.org/articles/RJ-2022-020/) from a local file, performs multiple linear regression, then outputs multiple files (with model estimates and predictions) as a ZIP archive.
```
Fil, line 3:
```suggestion
The data loader below reads in the [penguins data](https://journal.r-project.org/articles/RJ-2022-020/) from a local file, performs multiple linear regressions, then outputs multiple files (with model estimates and predictions) as a ZIP archive.
```

Fil (approved):
It feels a bit harsh to give a model without a chart, but that's up to you :)

I added two nits

Issue #32, opened 2023-10-18
Support inline expressions for attributes (and in other places)


mbostock:
Currently inline expression `${…}` are only supported for node content. So you can’t do this:

````md
```js
const link = "https://example.com";
```
This is a <a href=${link}>link</a>.
````

You can workaround it using [Hypertext Literal](https://github.com/observablehq/htl) like so:

````md
```js
const link = "https://example.com";
```
This is a ${htl.html`<a href=${link}>link</a>`}.
````

I’m not immediately sure how we would support this. It might involve writing our own Markdown parser? Or at least pre-processing the content to remove the live code expressions before passing it to markdown-it? Presumably we could repurpose parts of Hypertext Literal to parse the HTML.

PR #1425, opened 2024-06-04
robust placeholder

### LINKED ISSUE #375: uncaught TypeError: HTML comments with template literals and string interpolation

### LINKED ISSUE #396: Complex inline expressions aren’t parsed correctly

### LINKED ISSUE #636: Backslashes aren’t unescaped within PRE elements

mbostock:
This implements more robust parsing for inline expressions `${…}`.

Inline expressions are now parsed prior to Markdown tokenization using a similar strategy to [Hypertext Literal](https://github.com/observablehq/htl), based on the [HTML5 tokenization](https://html.spec.whatwg.org/multipage/parsing.html#tokenization) specification, and replaced with comments (_e.g._, `<--:803886cc:-->`). By removing the inline expression’s code from the Markdown before it is tokenized, we ensure that the code does not corrupt the meaning of the surrounding Markdown, and that the code itself is not corrupted.

The expressions themselves are parsed with Acorn’s JavaScript tokenizer, allowing correct parsing of comments and template literals within inline expressions. The expressions need not be valid syntax, but if they are valid tokens, we can correctly terminate the inline expression on the closing right brace token. If the expressions are invalid tokens (for example, invalid Unicode escape sequences of unterminated template literals), then we fall back to simply counting curly braces.

The parser also now correctly removes backslashes that are used to escape inline expressions, such as `\${…}` or `$\{…}`.

Here are some examples of inline expressions that now work correctly:

```md
${1 + /*}*/ 2}
```

```md
${1
+ 2}
```

```md
${html`<div>
  ${[1, 2, 3].map((i) => i)}
</div>`}
```

```md
${`

hello world

`}
```

Here’s an example of an inline expression now being ignored in a RAWTEXT context:

```md
<textarea>${1 + 2}</textarea>
```

Here are two examples of backslash removal:

```md
<pre>\${1}</pre>
```

```md
<pre>$\{1}</pre>
```

Fixes #375.
Fixes #396.
Fixes #597.
Fixes #636.


src/markdown.ts
```diff
@@ -36,10 +35,8 @@ export interface MarkdownPage {
   code: MarkdownCode[];
 }
 
-interface ParseContext {
+export interface ParseContext {
   code: MarkdownCode[];
-  startLine: number;
-  currentLine: number;
```
mbostock, line 42: This code has been dead for a while, so I’m removing it for now. I assume at some point we’ll get back to implementing better line numbers for errors, but better to delete the code if it’s not currently used.

src/placeholder.ts
```diff
@@ -0,0 +1,422 @@
...
+  STATE_COMMENT_END = 23,
+  STATE_COMMENT_END_BANG = 24,
+  STATE_MARKUP_DECLARATION_OPEN = 25,
+  STATE_RAWTEXT = 26,
+  STATE_RAWTEXT_LESS_THAN_SIGN = 27,
+  STATE_RAWTEXT_END_TAG_OPEN = 28,
+  STATE_RAWTEXT_END_TAG_NAME = 29;
+
+export interface PlaceholderToken {
+  type: "content" | "code";
```
mbostock, line 57: I expect it would now be possible to interpolate into attributes (like Hypertext Literal does). However, this will require a specialized version of the display function (that coerces to a string). I’m leaving this as future work but we are now much closer to fixing #32.

mbostock:
Oh ya, this is interesting:

```md
<code class="language-js">${2}</span>
```

It’s because of the syntax highlighting we do statically; it’s dropping the `<!--:id:-->` placeholder. Presumably this is a preexisting bug in the parent branch.

mbostock:
I need to take another crack at this. The problem with fenced code blocks demonstrates that we can’t simply give inline expressions the highest precedence (by processing them prior to Markdown tokenization); at a minimum, we need to process them after tokenizing fenced code blocks.

But also, we can’t simply process fenced code blocks before inline expressions because you could have a degenerate case like this:

````md
${/*

```
commented-out fenced code block within an inline expression?
```

*/}
````

But maybe we simply ignore this case — it doesn’t have any practical value, and it’s a bit like how script elements don’t allow you to have `</script>` anywhere? Maybe we can process inline expressions after fenced code blocks… I haven’t found yet where markdown-it is breaking up tokens.

**Comment by mbostock** [comment]
I have to run, but I don’t think it’s going to be practical to do this within markdown-it. The problem is that it uses its other block rules to determine when a block ends; so, to detect when a table block ends, it looks for its other “blockquote” terminator rules, and if any of them matches a line, that’s where the table block ends. But if we’re within an inline expression, then we don’t want it testing the lines of code.

So I think the simplest option (which is not so simple) is to replicate just the fenced code block parsing logic:

https://github.com/markdown-it/markdown-it/blob/master/lib/rules_block/fence.mjs

Our inline expression parser will need to detect fenced code blocks, and ignore inline expressions within those code blocks.

**Comment by mbostock** [comment]
Okay, I think everything is good now! I’ve tried to cover everything with tests, but here’s what I was also testing manually:

`````md
## hello ${1 + 2}

```js
const fortytwo = 42.4242;
```

${1/*

```
commented-out fenced code block within an inline expression?
```

*/}

${fortytwo}

<code class="language-js">${fortytwo.toFixed(2)}</code>

`te${1}st`

<!--

```

testing

```

-->

another test

<h2>hello</h2> ${"world"}

<!-- ${'world'} -->

<!--```

testing

```-->

<pre>$\\{1}</pre>

<pre>$\{1}</pre>

<pre>\\${1}</pre>

<pre>\${1}</pre>

<pre>\$</pre>

<pre>\{</pre>

<pre>$\{</pre>

<textarea>${1 + 2}</textarea>

${`

hello world

`}

${1+/*

```
ignore me
```

*/+2}

${html`<div>
  ${[1, 2, 3].map((i) => i)}
</div>`}

${1
+ 2}

${1 + /*}*/ 2}

````js echo
`${Math.random()}`
````

````js echo
${Math.random()}
````

foo${1 + 2}bar

<pre>${Math.random()},${Math.random()}</pre>

${1}
`````

File: src/placeholder.ts [hunk 3]
```diff
@@ -0,0 +1,536 @@
...
+                break;
+              }
+            } while (parser.type !== tokTypes.eof);
+          } catch (error) {
+            if (!(error instanceof SyntaxError)) throw error;
+            // on invalid token (e.g., unterminated template, invalid unicode escape),
+            // read until the braces are closed
+            let j = parser.pos;
+            for (; j < n; ++j) {
+              if (input.charCodeAt(j) === CODE_RBRACE && !--braces) {
```
mbostock, line 201: Maybe we should just stop at the first close brace here? Because we aren’t incrementing in left brace anymore so it feels weird to still be counting. And it’s a degenerate case anyway so probably we want to do the simplest thing.

mbostock: Simplified by now just ignoring inline expressions if the tokens are invalid.

Fil:
Almost! on http://127.0.0.1:3000/reactivity the console now says:

```
TypeError: Failed to execute 'observe' on 'IntersectionObserver': parameter 1 is not of type 'Element'.
    at variable_intersector (runtime.js:1276:14)
    at define (runtime.js:1309:23)
reject @ client.js:182
rejected @ client.js:139
variable_rejected @ runtime.js:963
(anonymous) @ runtime.js:1336
```

mbostock:
You’re finding all the bugs in the parent PR you already approved. 😅 

The problem here is that the `visibility` promise assumes that the variable’s `_root` is an element that can be observed with IntersectionObserver, but now the `_root` is a comment. So either we need to change the definition of `visibility` (which somewhat painfully is within the Observable Runtime), or `visibility` needs to watch the parent element. Probably the latter but that’s maybe not quite right for inline expressions? 🤔

Fil (approved):
I didn't find any more issues (I even tried it on pangea).

PR #1429, opened 2024-06-06
JSX + React

### LINKED ISSUE #971: Support for JSX format as a Markdown alternative
JSX is a format/language that allows JavaScript to be embedded consistently in an HTML-like syntax and to output HTML primitives. Originally written for React, it has been incorporated into a variety of other projects. JSX allows an intuitive mixing of markup and JavaScript elements. 

As such, it seems like a natural fit for Framework. 

While Framework uses markdown with islands of JavaScript, in fact it seems to rely heavily on HTML markup rather than just straight markdown. That's what JSX is meant for. It might be worth looking to see how easily a JSX parser could be dropped in as a markdown alternative format.

An alternative would be to use MDX, which allows JSX in markdown: https://mdxjs.com/table-of-components/

mbostock:
Extracted from #1216, this implements client-side support for JSX with React. It looks like this:

````md
```jsx
import {Counter} from "./components/Counter.js";

display(<Counter title="Hello, JSX" />);
```
````

In JSX blocks, you can only display one thing at a time; subsequent calls to `display` will replace whatever was displayed previously. Also, you can only display React elements.

TODO:

- [x] JSX code blocks
- [x] JSX modules
- [x] `React` built-in
- [x] `ReactDOM` built-in
- [x] Handle syntax errors during JSX transpilation
- [x] Implicit (singular) `createRoot` within `display`
- [x] Documentation
- [x] Tests

Future work:

- Allow JSX in inline expressions `${<Component />}` (requires acorn-jsx)?
- Allow JSX in Markdown directly (à la MDX)?

Fixes #971.

### CODE REVIEW

File: package.json [hunk 1]
```diff
@@ -92,6 +92,18 @@
...
+    "react-dom": "^18.2.0"
+  },
+  "peerDependenciesMeta": {
+    "react": {
+      "optional": true
+    },
+    "react-dom": {
+      "optional": true
+    }
+  },
```
mbostock, line 106: I don’t think this is needed now that we use react from `npm:react` instead of node_modules. 🤔

docs/jsx.md
```diff
@@ -0,0 +1,101 @@
...
+```
+
+With a JSX fenced code block, the [display function](./javascript#explicit-display) behaves a bit differently:
+
+- It replaces the previously-displayed content, if any
+- It never uses the inspector
+
+In addition, JSX fenced code blocks should always display explicitly; JSX fenced code blocks do not support implicit display of expressions.
+
+You don’t need to import `React` when using JSX; `React` is implicitly imported. `React` is also available by default in Markdown. If you need to, you can import it explicitly:
```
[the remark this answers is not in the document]
Fil: (or maybe you mean that it's available silently, but if you want to use its properties in javascript you need to import explicitly?)

docs/jsx.md
```diff
@@ -0,0 +1,101 @@
...
+import * as React from "npm:react";
+```
+
+You can also import specific symbols such as hooks:
+
+```js run=false
+import {useState} from "npm:react";
+```
+
+Always use the `.js` extension to import JSX modules.
```
Fil, line 62:
```suggestion
Always use the `.js` file extension to import JSX modules.
```

(suggestion to slightly reinforce that extension is the file extension not the fenced block language)

docs/jsx.md
```diff
@@ -0,0 +1,101 @@
...
+  return (
+    <Card title={title}>
+      <p>
+        <button onClick={() => setCounter(counter + 1)}>Click me</button>
+      </p>
+      <p>The current count is {counter}.</p>
+      <div
+        style={{
+          transition: "background 250ms ease",
+          backgroundColor: counter & 1 ? "brown" : "steelblue",
```
Fil, line 79: this background is too dark in light mode, making the sentence difficult to read

docs/jsx.md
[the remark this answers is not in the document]
mbostock: Yes, `React` is now a standard library built-in like `Plot` and `d3`. I’ll switch this to use the phrase we use elsewhere _e.g._ “D3 is available by default as d3 in Markdown, but you can import it explicitly like so.”

docs/jsx.md
mbostock: I’m referring to the file extension here. You import `Card.js` even though the source is `Card.jsx`. That’s because the JSX gets transpiled to JS and served as `Card.js`; at runtime it doesn’t matter that it was originally authored in JSX; at runtime it’s JS. (And if you’re a masochist you could hand-author components in JavaScript if you wanted.) The same convention is used to import TypeScript; you import `.js` even though the file is `.ts`.

docs/jsx.md
mbostock: I’ll make it prettier.

Fil (approved):
`<OK />!`

mbostock:
Hmm, I found a bug where if you move a JSX block to a new location during preview, it doesn’t get relocated (because we do the special treatment for `root._nodes` but that doesn’t apply to `root._root`). I think/hope I can fix that pretty easily though.

PR #1431, opened 2024-06-06
clear old content on undefine

### LINKED ISSUE #1430: Editing an imported module causes duplicate display
#1416 introduced a bug where old displays are not cleaned up when imported modules change. To reproduce, create a JavaScript fenced code block like so:

````md
```js echo
import {test} from "./test.js";

display(test);
```
````

Then create the `test.js` like so:

```js
export const test = 2;
```

Edit `test.js` and you will see the bug:

Screenshot 2024-06-06 at 8 26 19 AM

mbostock:
Fixes #1430. We were forgetting to clear old content when a cell is undefined during preview.

### DIFF
--- a/src/client/main.js
+++ b/src/client/main.js
@@ -25,7 +25,6 @@ const rootsById = findRoots(document.body);
 export function define(cell) {
   const {id, inline, inputs = [], outputs = [], body} = cell;
   const variables = [];
-  cellsById.get(id)?.variables.forEach((v) => v.delete());
   cellsById.set(id, {cell, variables});
   const root = rootsById.get(id);
   const loading = findLoading(root);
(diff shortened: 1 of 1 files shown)


src/client/main.js
```diff
@@ -119,7 +118,8 @@ function displayBlock(root, value) {
 }
 
 export function undefine(id) {
-  cellsById.get(id)?.variables.forEach((v) => v.delete());
+  clear(rootsById.get(id));
+  cellsById.get(id).variables.forEach((v) => v.delete());
```
mbostock, line 122: I removed the optional chaining here because we should only be undefining cells that were defined (and we want to notice if our logic is broken).

Fil (approved):

PR #1440, opened 2024-06-06
ignore inline expressions in HTML rawtext and comment blocks

### LINKED ISSUE #375: uncaught TypeError: HTML comments with template literals and string interpolation

mbostock:
Fixes #375. We were ignoring script elements already, but we need to ignore all rawtext elements and we also need to ignore comments. Technically we might also want to ignore cdata and other weird things, but those are extremely rare so I didn’t bother.

Ref. https://spec.commonmark.org/0.31.2/#html-blocks
Ref. https://github.com/observablehq/htl/blob/e3e9777d09f7b5aea3ebc52e73b86a1246dcab7e/src/index.js#L631-L637

### DIFF
--- a/src/markdown.ts
+++ b/src/markdown.ts
@@ -184,7 +184,6 @@ function parsePlaceholder(content: string, replacer: (i: number, j: number) => v
 
 function transformPlaceholderBlock(token) {
   const input = token.content;
-  if (/^\s*<script[\s>]/.test(input)) return [token]; // ignore <script> elements
   const output: any[] = [];
   let i = 0;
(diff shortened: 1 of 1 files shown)


mbostock:
This doesn’t correctly handle this case:

```html
<textarea>${1 + 2}</textarea> ${2}
```

Maybe we want to bring back the tokenizing parser from #1425, even if we’re only going to parse inline expressions within Markdown tokens.

mbostock:
Superseded by #1441.

PR #1441, opened 2024-06-07
inline tokenizer, take two

### LINKED ISSUE #375: uncaught TypeError: HTML comments with template literals and string interpolation

### LINKED ISSUE #396: Complex inline expressions aren’t parsed correctly

### LINKED ISSUE #636: Backslashes aren’t unescaped within PRE elements

mbostock:
Re-applies #1425, but without trying to change Markdown tokenization. Inline expressions must still be within blocks, but we at least now parse them correctly within the block.

Fixes #375.
Fixes #396 (by considering it intended).
Fixes #636.


mbostock:
So this still doesn’t work (by design, because Markdown):

```md
${html`<div>
  ${[1, 2, 3].map((i) => i)}
</div>`}
```

But you can easily fix it by putting it inside an HTML block:

```md
<div>
  ${html`<div>
    ${[1, 2, 3].map((i) => i)}
  </div>`}
</div>
```

Fil (approved):

Issue #1437, opened 2024-06-06
Inline expressions within lists with backticks are not parsed correctly


mbostock:
Blocks take precedence over inlines, so this isn’t parsed correctly:

````md
- `one${1 + 2}
- two`
````

mbostock:
Fixed in #1438.

Issue #1436, opened 2024-06-06
Inline expressions within setext headings with backticks


mbostock:
This is a setext heading, and hence the backtick should be interpreted literally rather than as a code span.

````md
`Hello ${1 + 2}
----
`
````

mbostock:
Fixed in #1438.

Issue #1435, opened 2024-06-06
Inline expressions within triple-backticks within HTML blocks are not parsed correctly


mbostock:
Here, the backticks are interpreted literally because they are inside an HTML block, and hence the inline expression should be allowed.

````md
<div>
  ```
  ${1 + 2}
  ```
</div>
````

mbostock:
Fixed in #1438.

Issue #1434, opened 2024-06-06
Inline expressions with on indented lines are not parsed correctly


mbostock:
E.g., 

```html
<div>
  <div>
    ${"hello"}
  </div>
</div>
```

mbostock:
Fixed in #1438.

PR #1442, opened 2024-06-07
a few parsing tests

Fil:
This adds tests for:

- #1433
- #1434
- #1435
- #1436
- #1437
- #1415

Regarding the SVG placeholder test, I wonder if it's fully working: I'd expect the `<observablehq-loading></observablehq-loading>` element to be removed once `hello` is inserted in the DOM. It does not seem to be the case?

```md
<svg width=640 height=120>
  <text x=20 y=20>${"hello"}</text>
</svg>
```

Capture d’écran 2024-06-07 à 11 31 00


mbostock:
The reason that `<observablehq-loading>` doesn’t get removed is because in SVG, element names are case-sensitive: the client is looking for the tag name `OBSERVABLEHQ-LOADING`, but in SVG the tag name is `observablehq-loading`. But more importantly, the `<observablehq-loading>` doesn’t do anything within an SVG element. (I don’t think custom elements are supported within SVG? But certainly the CSS rules we’re using for the loading indicator don’t work in this context.)

So probably we should remove the `<observablehq-loading>` during the HTML sanitization step where we currently check for reparenting.

mbostock (approved):
I added a test and a fix for the SVG loading issue you raised. (The bug had no visual consequences but still no reason to have a loading indicator if it isn’t doing anything in this context.) Thanks for that and the tests!

PR #1443, opened 2024-06-07
Don't prompt for interactive login on non-interactive terminals

### LINKED ISSUE #1432: `observable deploy` prompts for message in non-interactive shell, throws ERR_TTY_INIT_FAILED
Running `observable deploy` without the `--message` param prompts the user for a deploy message. The prompt is created even when `effects.isTty` is false, causing the process to error with
```
■  Error: TTY initialization failed: uv_tty_init returned EINVAL (invalid argument)
│
└  The full error follows

node:internal/errors:563
    ErrorCaptureStackTrace(err);
    ^

SystemError [ERR_TTY_INIT_FAILED]: TTY initialization failed: uv_tty_init returned EINVAL (invalid argument)
    at new SystemError (node:internal/errors:259:5)
    at new NodeError (node:internal/errors:370:7)
    at new WriteStream (node:tty:93:11)
    at TD.prompt (file:/// ... /node_modules/@clack/core/dist/index.mjs:9:818)
    at Module.te (file:/// ... /node_modules/@clack/prompts/dist/index.mjs:9:7)
    at Module.deploy (file:/// ... /node_modules/@observablehq/framework/dist/deploy.js:238:32)
    at async file:/// ... /node_modules/@observablehq/framework/dist/bin/observable.js:112:7 {
  code: 'ERR_TTY_INIT_FAILED',
  info: {
    errno: -22,
    code: 'EINVAL',
    message: 'invalid argument',
    syscall: 'uv_tty_init'
  },
  errno: [Getter/Setter],
  syscall: [Getter/Setter]
}
```

To reproduce, run the command with `stdin` closed:
```
observable deploy --debug <&-
```

mythmon:
Fixes #1432

### DIFF
--- a/src/deploy.ts
+++ b/src/deploy.ts
@@ -146,12 +146,12 @@ class Deployer {
           throw new CliError("No authentication provided");
         } else {
           const source =
-            apiKey?.source == "file"
+            apiKey.source == "file"
               ? ` from ${apiKey.filePath}`
               : apiKey?.source === "env"
               ? ` from $${apiKey.envVar}`
               : "";
-          throw new CliError(`Authentication${source} was rejected by the server`);
+          throw new CliError(`Authentication${source} was rejected by the server: ${authError ?? "unknown error"}`);
         }
       }
       const message =
--- a/src/deploy.ts
+++ b/src/deploy.ts
@@ -414,12 +414,16 @@ class Deployer {
 
     let message = this.deployOptions.message;
     if (message === undefined) {
-      const input = await this.effects.clack.text({
-        message: "What changed in this deploy?",
-        placeholder: "Enter a deploy message (optional)"
-      });
-      if (this.effects.clack.isCancel(input)) throw new CliError("User canceled deploy", {print: false, exitCode: 0});
-      message = input;
+      if (this.effects.isTty) {
+        const input = await this.effects.clack.text({
+          message: "What changed in this deploy?",
+          placeholder: "Enter a deploy message (optional)"
+        });
+        if (this.effects.clack.isCancel(input)) throw new CliError("User canceled deploy", {print: false, exitCode: 0});
+        message = input;
+      } else {
+        message = "";
+      }
     }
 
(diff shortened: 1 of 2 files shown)


src/deploy.ts
```diff
@@ -141,6 +141,19 @@ class Deployer {
     }
 
     if (!currentUser) {
+      if (!this.effects.isTty) {
+        if (authError === "unauthenticated" || !apiKey) {
```
mbostock, line 145: Do we need to handle the `"forbidden"` or `null` cases of `authError` here too?

mbostock: Looks like if `authError` is `null` that implies that `apiKey` is also `null`, so that is already covered. And if `authError` is `"forbidden"`, then the `else` block displays the rejection here. It’s a big hard to follow but it looks correct. 😅

src/deploy.ts
```diff
@@ -141,6 +141,19 @@ class Deployer {
...
+          throw new CliError("No authentication provided");
+        } else {
+          const source =
+            apiKey?.source == "file"
```
mbostock, line 149: `apiKey` can’t be nullish here because we checked `!apiKey` above.

mootari:
Please note that in the case described in #1432 the deploy fails after a successful upload, because the user is prompted for a message (which should likely default to an empty string in that case).

src/deploy.ts
```diff
@@ -141,6 +141,19 @@ class Deployer {
...
 
     if (!currentUser) {
+      if (!this.effects.isTty) {
+        if (authError === "unauthenticated" || !apiKey) {
+          throw new CliError("No authentication provided");
+        } else {
+          const source =
+            apiKey.source == "file"
+              ? ` from ${apiKey.filePath}`
+              : apiKey?.source === "env"
```
mbostock, line 151:
```suggestion
              : apiKey.source === "env"
```

mbostock (approved):

Issue #904, opened 2024-02-23
Descenders in H1 may be cut-off in the default template’s index page


mootari:
The default ~~theme~~ template index page specifies a line height of ~~1em~~ 1 for the h1 title, causing descenders on the last line in some fonts like Arial Black to get cut off.

<img width="1056" alt="image" src="https://github.com/observablehq/framework/assets/322014/df2b09fc-9e7b-4a7f-ad21-c603beeed7be">

mbostock:
Is this caused by the `background-clip: text`? It shouldn’t normally matter that the `line-height: 1` doesn’t fit the descenders. Also, this isn’t the default theme; it’s simply the index page on the default template.

mbostock:
Not planning on fixing this.

PR #1448, opened 2024-06-10
fix descender clip

mbostock:
Applies a bit of padding (in addition to margin) so the background extends below the descenders. Also moves the styles to the end of the `index.md` so that it’s a bit friendlier to edit the example `index.md` (without getting distracted by some elaborate CSS).

Before:
Screenshot 2024-06-10 at 4 49 29 PM

After:
Screenshot 2024-06-10 at 4 49 12 PM

### DIFF
--- a/templates/default/src/index.md.tmpl
+++ b/templates/default/src/index.md.tmpl
@@ -2,48 +2,6 @@
 toc: false
 ---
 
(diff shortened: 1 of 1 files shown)


Fil (approved):
#904

PR #1449, opened 2024-06-11
self-host sample datasets

### LINKED ISSUE #1427: Self-host sample datasets
E.g., `penguins`

mbostock:
Fixes #1427 by adopting the new [@observablehq/sample-datasets](https://github.com/observablehq/sample-datasets) package. Also fixes an unfiled bug where `import.meta.resolve(…)` would generate a pointless `modulepreload` for non-JavaScript files (such as CSV files).


Fil (approved):
Nice!

I still find it odd that the information about implicit symbols is spread out in two separate files—but I haven't found a way to centralize this.

PR #1451, opened 2024-06-11
tweak format of project identifiers in deploy

mythmon:
Cobus pointed out that the `@workspace/slug` format we were using looks a lot like our Notebook URLs, and can lead people to 404s if they try to naively create a URL from it. Breaking up the information like this helps make it not look like a URL. I'm open to other formats though, including not making the change if you want to debate it with Cobus.


Fil:
@mythmon please merge?

mythmon:
Ah, sorry. I missed that Matt approved this.

PR #1452, opened 2024-06-11
rust data loader example

mythmon:
For our recent data loaders, I had prepared two examples, but only had time to demo one of them in the presentation. This is the other one, a data loader example that uses Rust to simulate 10 million poker hands and reports the probability of getting each hand.

### DIFF
--- a/examples/loader-rust-to-json/src/data/poker.json.rs
+++ b/examples/loader-rust-to-json/src/data/poker.json.rs
@@ -1,0 +1,163 @@
+#!/usr/bin/env rust-script
+//! Since Framework uses rust-script, we can define dependencies here.
+//!
+//! ```cargo
+//! [dependencies]
+//! serde = { version = "1.0.203", features = ["derive"] }
+//! serde_json = "1.0.117"
+//! rand = "0.8.5"
+//! rayon = "1.10.0"
+//! ```
+
+use rand::Rng;
+use rayon::prelude::*;
+use serde::Serialize;
+use serde_json::json;
+use std::collections::HashMap;
+
+fn main() {
+    const COUNT: u32 = 10_000_000;
+    let start = std::time::Instant::now();
+
+    let counts = (0..COUNT)
+        // This line breaks the work up into multiple parallel jobs.
+        .into_par_iter()
+        // Calculate the category of random hands
+        .map(|_| {
+            let hand = Hand::random();
+            // Convert the category into a one-element hashmap, so the reducer
+            // can sum up all the counts for each category.
+            let mut map = HashMap::new();
+            map.insert(hand.categorize(), 1);
+            map
+        })
+        // count up each category
+        .reduce(
+            || HashMap::with_capacity(10),
+            |mut acc, map| {
+                for (category, count) in map {
+                    *acc.entry(category).or_insert(0) += count;
+                }
+                acc
+            },
+        );
+
+    let mut tidy_data = counts
+        .into_iter()
+        .map(|(category, count)| SummaryRow { category, count })
+        .collect::<Vec<_>>();
+    tidy_data.sort_by_key(|data| data.category);
+
+    serde_json::to_writer(std::io::stdout(), &json!({
+        "summary": tidy_data,
+        "meta": { "count": COUNT, "duration_s": start.elapsed().as_secs_f64() },
+    })).unwrap();
+}
+
+// Here, we create types for the domain model of a poker hand. Working with
+// specific types helps makes the rest of the code simpler.
+
+#[derive(Debug, Clone, Serialize)]
+struct SummaryRow {
+    category: HandCategory,
+    count: u32,
+}
+
+#[derive(Debug, PartialEq, Clone, Serialize)]
+struct Hand(Vec<Card>);
+
+#[derive(Debug, PartialEq, Clone, Copy, Serialize)]
+struct Card {
+    /// 1 is an Ace, 2-10 are the numbered cards, 11 is Jack, 12 is Queen, 13 is King.
+    rank: u8,
+    suit: Suit,
+}
+
+#[derive(Debug, PartialEq, Eq, Clone, Copy, Serialize, Hash)]
+enum Suit {
+    Clubs,
+    Diamonds,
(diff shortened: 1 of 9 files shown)


examples/loader-rust-to-json/src/data/poker.json.rs
```diff
@@ -0,0 +1,163 @@
+#!/usr/bin/env rust-script
```
Fil, line 1: I think this line is not strictly necessary (it's ignored); one could consider it informative ("this script is supposed to run with rust-script"), but in a sense it's also confusing as it means "this script is an executable that will call this particular interpreter", when in fact it is called from framework with a (possibly) different interpreter.

Similarly, it is not necessary to have the `+x` bit set on this file. I would argue that it is (slightly) more secure in general to turn off the executable bit when not strictly required.

In other words, only the `.exe` loaders should have a shebang and be `+x` (and even then, possibly only `+x` for the user or group that runs the script, if possible).

examples/loader-rust-to-json/src/data/poker.json.rs
```diff
@@ -0,0 +1,163 @@
...
+//! Since Framework uses rust-script, we can define dependencies here.
```
Fil, line 2: This is great!

examples/loader-rust-to-json/src/index.md
```diff
@@ -0,0 +1,232 @@
...
+})
+```
+
+```js echo
+function formatPercent(x) {
+  // Displays both large and small percentages by adding decimal places
+  // until there is something besides zeroes, or until the limit is hit.
+  for (let decimals = 0; decimals <= 5; decimals++) {
+    const f = d3.format(`.${decimals}%`);
+    let s = f(x);
```
Fil, line 225:
```suggestion
    const s = f(x);
```

examples/loader-rust-to-json/src/index.md
```diff
@@ -0,0 +1,232 @@
...
+      dx: 2,
+    }),
+  ]
+})
+```
+
+```js echo
+function formatPercent(x) {
+  // Displays both large and small percentages by adding decimal places
+  // until there is something besides zeroes, or until the limit is hit.
```
Fil, line 222: Neat! Another “trick” I like with percentages is to not round 0.99999 to 100% but display it as `99.9%` (not applicable here, but I can see this function being reused in other contexts).

examples/loader-rust-to-json/src/index.md
```diff
@@ -0,0 +1,232 @@
...
+
+<div style="max-width: 300px;">
+
+```js
+Inputs.table(hands.summary)
+```
+
+</div>
+
+Taking advantage of Rust’s performance and simple parallelism, the data loader above was able to summarize ${hands.meta.count / 1e6} million hands in ${hands.meta.duration_s.toFixed(1)} seconds.
```
Fil, line 191: This shows as “the data loader above was able to summarize 10 million hands in 1.3 seconds” and might show as “1 seconds” for someone with a slightly faster machine. You could use milliseconds to avoid having to do a singular “1 second”.

Fil:
Just a few nitpicking comments :) It's a great example!

examples/loader-rust-to-json/src/index.md
```diff
@@ -188,7 +188,7 @@ Inputs.table(hands.summary)
 
 </div>
 
-Taking advantage of Rust’s performance and simple parallelism, the data loader above was able to summarize ${hands.meta.count / 1e6} million hands in ${hands.meta.duration_s.toFixed(1)} seconds.
+Taking advantage of Rust’s performance and simple parallelism, the data loader above was able to summarize ${hands.meta.count / 1e6} million hands in ${hands.meta.duration_ms.toLocaleString()} ms.
```
mbostock, line 191:
```suggestion
Taking advantage of Rust’s performance and simple parallelism, the data loader above was able to summarize ${hands.meta.count / 1e6} million hands in ${hands.meta.duration_ms.toLocaleString("en-US")} ms.
```

When interpolating localized values into English text, we should specify the locale explicitly. It’ll look weird _e.g._ to interpolate French localization into English.

examples/loader-rust-to-json/src/index.md
mythmon: Good call!

mythmon:
As well as the review feedback, I made a fairly mechanical conversion of the Rust data loader into TypeScript, to get a speed comparison. Unsurprisingly, the Rust example is much faster, about 65x faster on my system. 20x of that can be theoretically attributed to parallelism, leaving a ~3x language speed up. I'm not sure if this should be included in the example though. What do you all think?

Fil:
I like the idea!

examples/loader-rust-to-json/src/index.md
```diff
@@ -0,0 +1,247 @@
...
+
+```js
+const extrapolatedTsTime = handsTs.meta.duration_ms * (hands.meta.count / handsTs.meta.count);
+```
+
+</div>
+
+Taking advantage of Rust’s performance and simple parallelism, the data loader above was able to summarize ${hands.meta.count / 1e6} million hands in ${hands.meta.duration_ms.toLocaleString("en-US")} ms. For comparison, an equivalent data loader written in TypeScript (which you can see in `data/poker-ts.json.ts`) would take ${Math.floor(extrapolatedTsTime / 1000).toLocaleString("en-US")} seconds to do the same simulation.
+
+We can make a quick chart of the poker hands with Plot
```
allisonhorst, line 201: Tiny nit, add a colon at the end of "We can make a quick chart of the poker hands with Plot"?

examples/loader-rust-to-json/src/data/poker.json.rs
allisonhorst: Yeah this is really cool

examples/loader-rust-to-json/src/index.md
mythmon:
```suggestion
We can make a quick chart of the poker hands with Plot:
```

allisonhorst (approved):

PR #1461, opened 2024-06-13
Update penguin classification to technique example (Py data loader)

allisonhorst:
- Renames directory to `/loader-python-to-csv`
- Updates index.md to match structure/length/style of other technique examples
- Updates README (for this example, and examples overview page)
- Tests new pattern for including virtual environment setup in copyable chunks
- Live version: https://observablehq.observablehq.cloud/framework-example-loader-python-to-csv/


Fil:
Is it a requirement to keep the URL? I think I'd prefer we use a new predictable URL https://observablehq.observablehq.cloud/framework-example-loader-python-to-csv/ and possibly deploy in the old URL a single file that redirects.

Other than that LGTM

mbostock:
I think platform supports redirects natively so you can just change the slug in the UI.

Fil (approved):

PR #1467, opened 2024-06-15
npm:react-dom

mbostock:
Tiny optimization to React: the `npm:react-dom/client` endpoint imports `npm:react-dom`, so there’s no reason to import the former. It’s slightly faster to just import `npm:react-dom` directly: one less dependency.

Also deletes an unlisted and now-defunct lib/react page.


Fil (approved):
:-)

https://github.com/reactjs/react.dev/issues/3896

PR #1468, opened 2024-06-15
consolidate venv docs

mbostock:
This consolidates the recommendation on using a Python virtual environment to the [Data loaders: Execution](https://observablehq.com/framework/loaders#execution) page. In addition, it avoids using `$` when demonstrating commands so that the code can be copied using the copy button and pasted into the terminal without editing.

The virtual environment documentation is now also slightly expanded to cover both venv and uv:
Screenshot 2024-06-14 at 6 39 44 PM

Before:
Screenshot 2024-06-14 at 6 40 43 PM

After:
Screenshot 2024-06-14 at 6 39 29 PM


mbostock:
Note: Once this is approved, I will manually re-deploy the affected examples.

allisonhorst:
So much better! Thank you!

allisonhorst (approved):
Looks great

PR #1473, opened 2024-06-20
if stdout is not a tty, replace clack.spinner by console.log

### LINKED ISSUE #1446: Add option to disable spinner during deploys
When an application logs the output of the `deploy` command, the spinner output "Server processing deploy" will appear in logs as displayed in the examples below. Without a progress bar (see #687) there is arguably not much value to it, and even with a progress status it may not be desirable to log the output in such detail.

A solution could be to add a `--no-progress` option to the `deploy` command (or globally) that would suppress any spinner output except for the `.stop()` message.

#### Output examples

Docker container logs:
~~~
◒  Server processing deploy
◐  Server processing deploy
<SNIP>
◑  Server processing deploy.
◒  Server processing deploy.
<SNIP>
◒  Server processing deploy..
◐  Server processing deploy...
◓  Server processing deploy...
◑  Server processing deploy...
◇  Deploy complete\n
~~~
GitHub action logs:
~~~
◒  Server processing deploy◐  Server processing deploy◓  Server processing deploy◑  Server processing deploy◒  Server processing deploy◐  Server processing deploy◓  Server processing deploy◑  Server processing deploy◒  Server processing deploy.◐  Server processing deploy.◓  <SNIP> Server processing deploy..◐  Server processing deploy..◇  Deploy complete
~~~

### LINKED ISSUE #1447: `deploy` spinner prints ANSI escape sequences in non-TTY environments, produces output without line breaks
The `deploy` command uses clack's spinner to print the upload progress for files. When running the command in a non-interactive environment the output still contains ANSI escape sequences used to clear the line, producing output like the following where all spinner updates are joined into a single line:
~~~
[?25l│
[999D[J◒  3 / 3 uploading example-report.html[999D[J◐  3 / 3 uploading example-report.html[999D[J◓  3 / 3 uploading example-report.html[999D[J◑  3 / 3 uploading example-report.html[999D[J◒  3 / 3 uploading example-report.html[999D[J◐  3 / 3 uploading example-report.html[999D[J◓  3 / 3 uploading example-report.html[999D[J◑  3 / 3 uploading example-report.html[999D[J◇  3 uploaded, 49 unchanged, 52 total.
[?25hUploading build manifest... {
~~~
In most cases the output will also print color codes like `[35m◒[39m`.

Related: #1446

Fil:
closes #1447
closes #1446

~~I'm not happy that I had to make c8 specifically ignore the non-coverage of the non-tty output tests, but I did not find a better way.~~ (solved in the new approach, see below)

### DIFF
--- a/src/convert.ts
+++ b/src/convert.ts
@@ -1,14 +1,13 @@
 import {existsSync} from "node:fs";
 import {utimes, writeFile} from "node:fs/promises";
 import {join} from "node:path/posix";
-import * as clack from "@clack/prompts";
 import wrapAnsi from "wrap-ansi";
 import type {ClackEffects} from "./clack.js";
 import {CliError} from "./error.js";
 import {prepareOutput} from "./files.js";
 import {getObservableUiOrigin} from "./observableApiClient.js";
 import type {TtyEffects} from "./tty.js";
-import {bold, cyan, faint, inverse, link, reset, defaultEffects as ttyEffects} from "./tty.js";
+import {bold, cyan, defaultEffects as defaultTtyEffects, faint, inverse, link, reset} from "./tty.js";
 
 export interface ConvertEffects extends TtyEffects {
   clack: ClackEffects;
@@ -19,8 +18,7 @@ export interface ConvertEffects extends TtyEffects {
 }
 
 const defaultEffects: ConvertEffects = {
-  ...ttyEffects,
-  clack,
+  ...defaultTtyEffects,
   async prepareOutput(outputPath: string): Promise<void> {
     await prepareOutput(outputPath);
   },
--- a/src/create.ts
+++ b/src/create.ts
@@ -7,7 +7,6 @@ import {basename, dirname, join, normalize} from "node:path/posix";
 import {setTimeout as sleep} from "node:timers/promises";
 import {fileURLToPath} from "node:url";
 import {promisify} from "node:util";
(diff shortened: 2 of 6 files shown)


mythmon:
Generally we've been pushing these tty checks higher in the call stack, so that we can have different messages on non-TTYs. In the deploy effects we have a `isTty` variable that we can use in tests to test both kinds of messages. I'd prefer if we keep that pattern instead of making Clack sometimes behave differently.

Fil:
`isTty` tests whether **stdin** is a tty, which will orient the questions one way or the other.

This by contrast is checking if **stdout** is a tty, so that we know not to output control characters; this is similar to what we do with color() in 
https://github.com/observablehq/framework/blob/f0e45f1e4490592dff4aa476f7d553fca77e8b39/src/tty.ts#L21

(In a sense, `isTty` is poorly named.)

mythmon:
I suppose we should add `isStdinTty` and `isStdoutTty` (I'd keep a `isTty` that is equal to `isStdinTty && isStdoutTty` too).

My point was more that we should push this up the stack, and leverage the effect system instead of doing it at import time.

Fil:
@mythmon I've tried a different approach, hopefully closer to what you had in mind?

mythmon (approved):
I think that's better. Good enough to merge at least so our CI logs look better.

PR #1475, opened 2024-06-20
expand the inspector if its previous version was expanded

### LINKED ISSUE #1458: keep the inspector open on update
Related to #1457, it would be great if we could keep the inspector open in the same[1] position when the inspected object is updated, instead of reverting to the initial (closed) state.

[1] or same-ish… as much as the differences between data structure make possible.

Fil:
We save the current inspector's expansion state in root.inspectorExpanded, and restore it when we create a new inspector, by passing a div that contains a span with the expected "observablehq--expanded" class.

for now inspectorExpanded is true if defined — this should be revised more thoroughly when we fix https://github.com/observablehq/inspector/issues/12

closes #1458


src/client/main.js
```diff
@@ -67,7 +67,12 @@ export function define(cell) {
 function noop() {}
 
 function clear(root) {
-  for (const v of root._nodes) v.remove();
+  delete root._inspectorState;
```
mbostock, line 70: Is this supposed to be `_inspectorExpanded`?

src/client/main.js
```diff
@@ -67,7 +67,12 @@ export function define(cell) {
...
+  for (const v of root._nodes) {
+    // If this was an inspector, note whether it is expanded.
+    if (v.firstChild?.classList.contains("observablehq--expanded")) root._inspectorExpanded = true;
```
mbostock, line 73: So, if any of the displayed items was expanded, all of them will be expanded on the next display? (Unlike in notebooks, code blocks in Framework can call `display` multiple times.)

I prefer that either we track the index of the expanded items and restore them consistently, or we special-case the expanded state of the first item, or we special-case the expanded state for the display of a singular item. I don’t think we really need to handle restoring the expanded state of multiple displayed items, but I don’t want to accidentally expand a lot of displayed items, either.

src/client/inspect.js
```diff
@@ -1,7 +1,15 @@
 import {Inspector} from "observablehq:runtime";
 
-export function inspect(value) {
+export function inspect(value, expand) {
   const node = document.createElement("div");
+  // If the previous inspector was expanded, make the new inspector expand. This
+  // relies on https://github.com/observablehq/inspector/blob/main/src/index.js
```
mbostock, line 6:
```suggestion
  // relies on https://github.com/observablehq/inspector/blob/63535de10ffb802959324e73553b9825311bfe48/src/index.js#L21-L23
```

Seems a bit hacky to create an empty SPAN element with the appropriate class to trick the inspector into restoring the expanded state. Maybe we should expose the `inspect` method within the inspector so there is more control over how the inspector renders?

mbostock (changes requested):
We should avoid using `delete` and instead represent inspector state consistently by initializing it when the root node is attached (_i.e._, in `define`).

Also, we should handle the multiple displayed items case better — we should err towards collapsing rather than expanding.

(I think technically we avoid avoid storing any additional state on the `root` because it could be part of the `display` closure, but that’s just an optimization I guess.)

src/client/main.js
Fil: oops, yes

src/client/main.js
Fil: I've now changed the logic to store a boolean for each node; however I don't think there is any way to absolutely guarantee constancy if the code block is playing tricks like this:

~~~
```js
if (Math.random() > 0.5) display(foo);
display(bar);
```
~~~

similarly:

~~~
```js
display(Math.random() > 0.5 ? foo : bar);
```
~~~

you will expand foo, refresh, and see bar expanded. I think it's OK, though.

src/client/inspect.js
Fil: Ideally as part of https://github.com/observablehq/inspector/issues/12 ?

I'm all for it, but also think this PR has a good Pareto ranking (solving > 90% of the use cases for  < 10% of the work).

src/client/main.js
Fil: > represent inspector state consistently by initializing it when the root node is attached

I'm not sure I understand this comment, since the state varies with user interaction. Either we have to listen to changes, or we read the state when we "unmount" the node to make room for the new versions.

Fil:
my `test.md` bench:

~~~

```js
const value = view(Inputs.button("reset"));
```

```js
const foo = {name: "foo", value};
const bar = {name: "bar", more: {value}};
```

```js
foo
```

```js
if (Math.random() > 0.5) display(foo);
display(bar);
```

```js
Math.random() > 0.5 ? foo : bar
```
~~~

src/client/main.js
mbostock: All I meant here is that we should do something like

```js
root._inspectorExpanded = [];
```

around here when the `root` is first bound:

https://github.com/observablehq/framework/blob/e123e7a85d61c83557a555fd005e0246037cbdd3/src/client/main.js#L31

mbostock (approved):
🚀

PR #1492, opened 2024-06-28
framework example: loader-github-react

Fil:
an example of querying GitHub, git, npm… to get a lot of information about a repo.

built for the [facebook/react](https://github.com/facebook/react) repo and deployed at https://observablehq.observablehq.cloud/framework-example-loader-github-react/

**facebook/react** is very interesting as it's a large, popular, collaborative project with quite a bit of history. I'm not sure that the charts are all self-explaining, and maybe it would be interesting to add some editorial content? (At the very least we could hard-code the repo name in the title before deploying.)

Also, maybe this is a showcase example… or a template as we initially thought…?

I've tested this with other repos, such as torvalds/linux and its 1.2+ million commits — and was also thinking that the deployed version could have several pages showing different source repos.


Fil:
superseded by [oss-analytics](https://github.com/observablehq/oss-analytics/)

PR #1493, opened 2024-06-28
Don't use a select widget to select from >10 projects

mythmon:
This should help when deploying projects to workspaces with many projects.

### DIFF
--- a/src/deploy.ts
+++ b/src/deploy.ts
@@ -88,7 +88,11 @@ type DeployTargetInfo =
 /** Deploy a project to ObservableHQ */
 export async function deploy(deployOptions: DeployOptions, effects = defaultEffects): Promise<void> {
   Telemetry.record({event: "deploy", step: "start", force: deployOptions.force});
-  effects.clack.intro(`${inverse(" observable deploy ")} ${faint(`v${process.env.npm_package_version}`)}`);
+  effects.clack.intro(
+    `${inverse(" observable deploy ")}${
+      process.env.npm_pacakge_version ? faint(` v${process.env.npm_package_version}`) : ""
+    }`
+  );
 
   const deployInfo = await new Deployer(deployOptions, effects).deploy();
 
@@ -754,7 +758,7 @@ export async function promptDeployTarget(
     throw error;
   }
 
-  if (existingProjects.length > 0) {
+  if (existingProjects.length > 0 && existingProjects.length < 10) {
     const chosenProject = await effects.clack.select<{value: string | null; label: string}[], string | null>({
       message: "Which project do you want to use?",
       options: [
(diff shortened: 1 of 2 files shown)


mbostock:
Can I see a screenshot of what this looks like?

Fil:
| first step | second step |
| -- | -- |
| first-step | screenshot |

I'm not sure we need to list all the names like this, as the list might expand into the hundreds or thousands.

Fil:
This PR allowed me to redeploy a project, so fixes the issue I had. 👍

mythmon:
The UI that I'd really like to have (for both cases) would show a partial list, and allow filtering it by typing. Then once you've narrowed it down you could use arrow keys to select one. Sort of like [`fzf`](https://junegunn.github.io/fzf/).

Given that we require them to type a valid slug, I wanted some way to convey the available options. I agree that this version of the UI definitely isn't very elegant in that regard. Maybe we could print the URL to the workspace's project listing instead?

Sorting them would be a good improvement. If we want to keep the list in the CLI, I'll add that.

src/deploy.ts
```diff
@@ -88,7 +88,11 @@ type DeployTargetInfo =
 /** Deploy a project to ObservableHQ */
 export async function deploy(deployOptions: DeployOptions, effects = defaultEffects): Promise<void> {
   Telemetry.record({event: "deploy", step: "start", force: deployOptions.force});
-  effects.clack.intro(`${inverse(" observable deploy ")} ${faint(`v${process.env.npm_package_version}`)}`);
+  effects.clack.intro(
+    `${inverse(" observable deploy ")}${
+      process.env.npm_pacakge_version ? faint(` v${process.env.npm_package_version}`) : ""
```
mbostock, line 93: This gets baked into the code during build, so it shouldn’t be necessary to check this.

https://github.com/observablehq/framework/blob/2bba99bfe0ff58a56790230b6bd61f0b5682de3c/build.js#L17

src/deploy.ts
```diff
@@ -772,6 +776,38 @@ export async function promptDeployTarget(
     } else if (chosenProject !== null) {
       return {create: false, workspace, project: existingProjects.find((p) => p.slug === chosenProject)!};
     }
+  } else if (existingProjects.length >= 10) {
```
mbostock, line 779: Any harm in nesting `if (existingProjects.length < 10) {…} else {…}` here so that we don’t have to repeat the 10 twice and risk a chance that one of the conditions won’t be handled?

src/deploy.ts
```diff
@@ -772,6 +776,38 @@ export async function promptDeployTarget(
...
+        validate: (slug) => {
+          if (existingProjects.some((p) => p.slug === slug)) return undefined;
+          return "Project not found";
+        }
+      });
+      if (effects.clack.isCancel(chosenProject)) {
+        throw new CliError("User canceled deploy.", {print: false, exitCode: 0});
+      } else if (chosenProject !== null) {
+        const project = existingProjects.find((p) => p.slug === chosenProject);
+        if (!project) throw new Error("selected a non-existent project");
```
mbostock, line 807: Is this possible given the check earlier on L799? Just for safety?

mbostock (changes requested):
Seems good, just some cosmetic requests.

mbostock:
I just noticed this appears to be fixed in the top-of-trunk version of clack, but that version has not yet been published to npm?

https://github.com/bombshell-dev/clack/pull/174

But despite not happening automatically, it looks like in the latest released version we are using, it can be set manually, so maybe we should just set it to `Math.max(process.stdout.rows - 4, 0)` as above?

https://github.com/bombshell-dev/clack/releases/tag/%40clack%2Fprompts%400.7.0

mythmon:
Thanks for the review! I didn't realize that was a built-in Clack option. It might take me a little bit to get back to this though.

mbostock:
All good @mythmon. I put up #1618 and if you don’t have time, hopefully @Fil can review.

mythmon:
Closing in favor of #1618

PR #1504, opened 2024-07-08
support view(jsx)

### LINKED ISSUE #1503: react/jsx should work with view
~~~~
```jsx echo
const a = view(<input type="range"></input>);
```

${a}
~~~~

Fil:
probably not the correct solution (?). Creates a "fake" `onMount` event and runs it after the jsx component is rendered into the DOM.

it works for the basic example:

~~~~
```jsx echo
const a = view(<input type="range"></input>);
```

${a}
~~~~

The main difficulty is that `jsx` cannot return its DOM node because it's creating it asynchronously, and `view` expects it synchronously. The second difficulty is that I don't know if we should `useRef`, `onMount`, or something else instead. 

closes #1503

### DIFF
--- a/src/client/main.js
+++ b/src/client/main.js
@@ -101,6 +101,7 @@ function displayJsx(root, value) {
       root.parentNode.insertBefore(node, root);
     }
     client.render(value);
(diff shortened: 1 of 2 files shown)


mbostock (commented):
Per other issue I think we shouldn’t do this.

PR #1506, opened 2024-07-09
Document how to use aq.loadArrow in combination with a file attachment.

Fil:
This question came as a support request.

### DIFF
--- a/docs/lib/arquero.md
+++ b/docs/lib/arquero.md
@@ -78,4 +78,10 @@ dt.fold(aq.all(), {as: ["city", "sun"]})
   .objects()
 ```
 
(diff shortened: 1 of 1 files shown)


mbostock (commented):
Does loadArrow return a promise to a table? It would probably better to include the const declaration so that it’s more obvious how to use the loaded table.

Fil:
I've updated the copy with a more complete example.

mbostock (commented):
Should we add a `FileAttachment.arquero` method? Presumably it’d be trivial and it makes sense given that `aq` is already part of the standard library.

mbostock:
Superseded by #1509.

PR #1509, opened 2024-07-09
FileAttachment.arquero

mbostock:
Ref. https://github.com/observablehq/framework/pull/1506#pullrequestreview-2167443702


Fil:
A question would be, should the `.arquero` method also work on a JSON or CSV file (or even a "fixed" file, for [aq.loadFixed](https://idl.uw.edu/arquero/api/#loadFixed))?

mbostock:
Okay, I added support for CSV, TSV, and JSON.

Fil:
I've added a bit of documentation.

I this this raises another question: if we want to generalize this approach, shouldn't we have a duckdb method as well, _e.g._ `FileAttachment("data.csv").duckdb()`?

mbostock:
We certainly could add a DuckDB method, but I don’t think that should block this PR, and I think it makes slightly less sense in the DuckDB case because typically a database has multiple tables and so you don’t want to give the file attachment primacy (_i.e._, it’s better to use `DuckDBClient.of` to accommodate multiple tables than `FileAttachment.duckdb` which is limited to a single table).

mbostock:
Added support for Apache Parquet files (and preloading which is a bit tricky because we only need to additionally preload parquet-wasm when using an Apache Parquet file with Arquero). Made the Arquero loader detection a bit more robust considering the file extension when the MIME type isn’t sufficient. Also tweaked the documentation a bit.

PTAL @Fil thank you!

Fil (approved):

PR #1510, opened 2024-07-10
Add example of canvas to png data loader

palewire:
I propose adding this example, which shows how you can use Node.JS in a data loader to serve a canvas map as a PNG. I currently use this pattern to render large amounts of data in lightweight static files.

I used [the official d3 example](https://observablehq.com/@d3/u-s-map-canvas) of how to make a canvas map to put together this demo.

Screenshot from 2024-07-10 10-49-05
Screenshot from 2024-07-10 10-49-16

PR #1514, opened 2024-07-11
d3/canvas to png

Fil:
local copy of #1510  by @palewire

the example is deployed at <https://observablehq.observablehq.cloud/framework-example-loader-canvas-to-png/>

### DIFF
--- a/examples/loader-canvas-to-png/.gitignore
+++ b/examples/loader-canvas-to-png/.gitignore
@@ -1,0 +1,5 @@
+.DS_Store
+/dist/
+node_modules/
+yarn-error.log
+.venv
--- a/examples/loader-canvas-to-png/README.md
+++ b/examples/loader-canvas-to-png/README.md
@@ -1,6 +1,6 @@
 [Framework examples →](../)
 
-# Python data loader to generate PNG
+# Data loader to generate PNG from canvas
 
 View live: <https://observablehq.observablehq.cloud/framework-example-loader-canvas-to-png/>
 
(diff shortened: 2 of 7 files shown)


mbostock (changes requested):
Example looks good. Please add it to the list in the examples README.

examples/README.md
```diff
@@ -54,6 +54,7 @@
 
 - [`loader-airtable`](https://observablehq.observablehq.cloud/framework-example-loader-airtable/) - Loading data from Airtable
 - [`loader-arrow`](https://observablehq.observablehq.cloud/framework-example-loader-arrow/) - Generating Apache Arrow IPC files
+- [`loader-canvas-to-png`](https://observablehq.observablehq.cloud/framework-example-loader-canvas-to-png/) - Generating a map server-side as a PNG file with node, canvas and D3
```
mbostock, line 57: This description can be shortened and better match the other descriptions. Some alternatives:

1. Generating PNG from JavaScript
2. Generating PNG from Node.js
3. Generating PNG using node-canvas
4. Generating PNG

I think my preference is (3) because most of our other data loader examples assume Node.js + JavaScript in the description, so it feels anomalous to call this one out, and it’s nice to reference the supporting library (node-canvas) explicitly.

(Also, it’s “Node.js” not “node”. Maybe “Node” is okay in some contexts, but I wouldn’t use “node” unless you are using `node` to refer specifically to the command you run from the terminal.)

mbostock (changes requested):

mbostock (approved):

PR #1522, opened 2024-07-17
chaining data loaders

### LINKED ISSUE #332: Chaining data loaders
Suppose we want to 1. download a dataset from an API and 2. analyze it. Currently a data loader must do both at the same time, and will run the download part again if we update the analysis code.

Ideally we'd want to separate this into two chained data loaders, that would still be live, _i.e._ if a page relies on 2 that relies on 1, editing 1 would tell the page to require the analysis again, which would trigger a new download. But editing 2 would only run the analysis again, not the download.

This would also make it easier to generate several files from a common (and slow) download.

Fil:
We must use the same file server for browser preview and machine calls (i.e. chained data loaders), because concurrent requests on a same data loader must be joined.

But the paths are different. The data loader for `caller.csv` receives a `$SERVER` environment variable equal to
`http://127.0.0.1:3000/_chain/caller.csv::`, and might retrieve a dependency by concatenating that variable and the file path it needs, _e.g._ calling `http://127.0.0.1:3000/_chain/caller.csv::/dependency.zip`.

This should make it possible to derive the dependency graph, at least after we run the data loaders (not sure how to maintain state when we restart a server and we have a cache). Also, if the file is not found, we send an empty 404 instead of the decorated page intedned for the browser; this makes it a bit more foolproof.

The current server information is saved as a global (in `process.env`—should it be `globalThis`?) for now. It feels a bit wrong, but at the same time it really is a global state.

TBC…

closes #332

### DIFF
--- a/src/dataloader.ts
+++ b/src/dataloader.ts
@@ -317,7 +317,10 @@ class CommandLoader extends Loader {
   }
 
   async exec(output: WriteStream): Promise<void> {
-    const subprocess = spawn(this.command, this.args, {windowsHide: true, stdio: ["ignore", output, "inherit"]});
+    const address = process.env.OBSERVABLEHQ_ADDRESS;
+    if (address == null) throw new Error("chained data loaders are not implemented in this context");
+    const env = {...process.env, SERVER: `${address}_chain${this.targetPath}::`};
+    const subprocess = spawn(this.command, this.args, {windowsHide: true, stdio: ["ignore", output, "inherit"], env});
     const code = await new Promise((resolve, reject) => {
       subprocess.on("error", reject);
       subprocess.on("close", resolve);
--- a/src/preview.ts
+++ b/src/preview.ts
@@ -95,13 +95,14 @@ export class PreviewServer {
     } else {
       await new Promise<void>((resolve) => server.listen(port, hostname, resolve));
     }
-    const url = `http://${hostname}:${port}/`;
+    const address = `http://${hostname}:${port}/`;
     if (verbose) {
       console.log(`${green(bold("Observable Framework"))} ${faint(`v${process.env.npm_package_version}`)}`);
-      console.log(`${faint("↳")} ${link(url)}`);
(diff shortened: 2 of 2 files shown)


mbostock:
This seems like the right direction. 👍 

I’m thinking of going a step further and spinning up a temporary “asset server” each time we invoke a data loader, so that each data loader gets a separate port for making requests to load assets/dependent files. I think that’s cleaner than using a prefixed path on the preview server, means build and preview follow the same code path, and anyway it should be cheap to expose a little asset server port temporarily while running a data loader (since there’s no initialization other than opening the port). And the asset server can further restrict what each data loader has access too (only files, not pages). And we can give it clean paths.

I think we could also get away with not tracking the dependencies explicitly and just let circular data loaders deadlock for now…

mythmon:
Could we make the "protocol" of how to load data loaders flexible enough to provide other capabilities? For example, it would be nice to be able to explicitly record dependencies that Framework can't detect, and one way we could do that is by making an HTTP call to the injected environment variable to record a file that we don't need to request but should be watched for changes.

Two ways I could imagine doing this is either to make that environment variable less generic, like `$DATA_SERVER` instead of `$SERVER`, or by adding a path prefix like `$SERVER/load/path/to/my/file.csv`. That would allow something like `$COMMAND_SERVER` or `$SERVER/track/path/to/library.js`. (all very bike-sheddable names)

Fil:
I have rewritten this based on the suggestions in the review. There is a new TODO list.

Fil:
Some of our examples will benefit from chained data loaders (in lieu of an intermediate archive). This will allow to split the analysis into independent small scripts. (thinking about this in the context of #1667).

ae3e:
Hi,

Is there any plan to merge this PR soon?
It would be a game changer for our use cases! 🚀 (multi-steps data analysis reusing first generated files)
Anyway, thx for your great job with this project! 🙏

PR #1523, opened 2024-07-17
parameterized routing & page loaders

### LINKED ISSUE #245: Parameterized data loaders & pages
What if we supported dynamic routes for data loaders, e.g. `customer/@[login].csv.ts` would match `customer/@mbostock.csv`, and would pass `--login=mbostock` as a command-line argument (parameter) to the data loader. This way you could use the same data loader to generate multiple files (though all files would need to be statically referenced in order to be built).

### LINKED ISSUE #931: Server-Side Rendering (SSR)

mbostock:
This introduces two concepts: **page loaders** allow server-side rendering of dynamic Markdown pages at build time (like data loaders, but for pages), while **parameterized routes** allow both page and data loaders to generate multiple outputs with varying parameter values.

Note: after this merges, or if you check out this branch locally, you may need to run `git clean -i` to remove the previous copies of theme pages.

TODO

- [x] Parameterized data loader routing
- [x] Parameter in file name (`[param].csv.js`)
- [x] Parameter in folder name (`[param]/data.csv.js`)
- [x] Partially-parameterized names? (`prefix-[param].csv.js`)
- [x] Names with multiple parameters? (`[param1]-[param2].csv.js`)
- [x] Extract parameters from route (`--param=foo`)
- [x] Parameterized page routing
- [x] Parameterized JavaScript components (rewriting `observable.params.param`)
- [x] Parameterized JavaScript code blocks (rewriting `observable.params.param`)
- [x] Parameterized JavaScript code expressions (rewriting `observable.params.param`)
- [x] Respect shadowing when rewriting `observable.params.param`
- [x] Allow static string concatenation within `FileAttachment` calls
- [x] Parameterized stylesheets (no rewriting; not recommended but should work)
- [x] Parameterized static files (no rewriting; not recommended but should work)
- [x] Parameterized archives
- [x] Don’t allow empty parameter values to match (`path/to/.csv`)
- [x] Don’t allow empty parameter names (`[].csv.js`)
- [x] Parameter value enumeration from config during build
- [x] Page loaders
- [x] Fix sidebar and pager links
- [x] Fix search indexing
- [x] ~~Default paths should include page loaders, too (just not _parameterized_ routes)~~ Future work
- [x] Rename the **paths** config option to ~~**params**~~ **dynamicPaths**
- [x] Documentation for parameterized routing
- [x] Documentation for page loaders
- [x] Tests for parameterized routing
- [x] Tests for page loaders

Fixes #245.
Fixes #931.


mbostock:
Precedence:

```
# literal request
/data/foo.csv

# first look for exact matches for each interpreter
/data/foo.csv.js
/data/foo.csv.py
/data/foo.csv.go

# parameterized data loader
/data/[user].csv.js --user=foo

# look for parameterized parents (1)
/[resource]/foo.csv.js --resource=data

# look for parameterized parents (2)
/[resource]/[user].csv.js --resource=data --user=foo

# then look for parent extractors
/data.zip
/data.zip.js
/data.zip.py
/data.zip.go
/data.tar.gz.js
/data.tar.gz.py
/data.tar.gz.go

# then look for parameterized extractors (3)
/[resource].zip.js --resource=data
```

Valid vs. invalid parameters:

```
# valid parameters
/[name]/foo.csv.js
/data/[name].csv.js

# invalid parameters
/foo-[name]/foo.csv.js
/[name]-bar/foo.csv.js
/data/[name]-bar.csv.js
```

mbostock:
Starting to head down the rabbit hole… 🎩  🐇 🤔 

So far, this PR implements dynamic routing for data loaders: you can say _e.g._ `FileAttachment("foo.json")` and it will be served by `[param].json.js` with `--param=foo`. So far so good, but this isn’t especially useful as the parameter value (`foo`) must be declared statically (since `FileAttachment` requires a static string as argument, but moreover because we require all files to be statically declared so we know what to build).

It’ll be more useful when we also have dynamic routing ofr pages, so you can have _e.g._ `/users/mbostock` as a page, which is backed by `/users/[user].md`. But there’s no way to statically reference `/users/[user]/data.json` that I can think of. So we could instead have a page `/users/mbostock/` backed by `/users/[user]/index.md`, and then a relative (static) reference to `./data.json` resolves to `/users/mbostock/data.json` backed by `/users/[user]/data.json.js` which works.

But now imagine that we also have a JavaScript component `/users/[user]/widget.js`. This too could be imported as `./widget.js` from `/users/[user]/index.md`. Should this be considered a “parameterized” component? Should there be a way to reference the parameter value within the code, such as `process.env.USER`, that gets baked-into the component when it is generated? Or should this be served as `/users/[user]/widget.js` and we effectively treat it as a static route within the dynamic route, serving the same content across all parameter values?

For now, I’m tempted to do the latter, since I think adding parameterized aliasing for JavaScript components (and likewise stylesheets) would introduce quite a bit of complexity since we need to track both the serving path (`/users/mbostock/widget.js`) and the source path (`/users/[user]/widget.js`) to traverse imports and compute content hashes correctly.

mbostock:
The purpose of parameterized pages is essentially to enable parameterized data loaders (through relative paths as described in my previous comment). If we enumerate the parameter values in the config, we can build a page for each parameter value; the parameterized page can then reference a parameterized data loader as a relative path.

I don’t have any use case imagined (yet) for parameterized JavaScript components or stylesheets. It’s just something that “falls out” of having parameterized pages. I’m not trying to actively support parameterized JavaScript components, but we’ll have to do something with path rewriting to avoid it.

mbostock:
As another example, say you have the parameterized page `/[user]/index.md` which references the file `./foo.txt`. If this is a static file, _i.e._ there exists the file `/[user]/foo.txt` in the source root, then the contents of this file are the same regardless of the parameter value. So if we have both `/a/index.md` and `/b/index.md`, we don’t want to generate copies `/a/foo.txt` and `/b/foo.txt`; we want to use a single static `/[user]/foo.txt` that is shared by both pages. The same goes for a JavaScript component (that can’t be generated by a data loader, at least not currently).

So when we resolve references from parameterized pages, we only want to preserve the parameterized-ness for resources that are generated by data loaders. For other resources we’ll need to rewrite the paths so that they are treated statically.

mbostock:
I’ve thought of another challenge of trying to share static resources across parameter values.

If a parameterized page imports a JavaScript component, we want that JavaScript component to be considered static (since the code will be the same across all parameter values and we’re not trying to support parameterized components). However, that JavaScript component could reference a file, and that file could be generated by a data loader which is parameterized. Since JavaScript components live in `_import` while files live in `_file`, we need to know the full path in the source root to the file… which means the built JavaScript component _can’t_ be considered static, because it will need to bake-in the parameterized path, _e.g._, `FileAttachment("../../foo/data.txt", import.meta.url)` where `foo` is the parameter value. 🤔

So I think either we should duplicate shared resources across parameterized pages (de-duplicating is just a performance optimization anyway, and can be done manually by moving the resources to a non-parameterized path) — or at least duplicate shared components. Or maybe we avoid the concept of parameterized pages entirely.

But if we don’t have parameterized pages, then how would you practically take advantage of parameterized data loaders?  You can’t pass a dynamic argument to `FileAttachment`. And I think we really want parameterized pages anyway… without them we’d force users to parse query parameters to create the equivalent linkable view, and that feels awkward.

mbostock:
I want to try leaning into parameterization.

I think the answer here is that we don’t try to share static assets across parameters; if you have `/[user]/foo.png`, it’s duplicated to `/a/foo.png`, `/b/foo.png`, _etc._ Maybe in the future when we have server-side rendering, only some of those assets end up being present for some parameter values. And it’s desirable to not leave any evidence as to whether `foo.png` is generated by a data loader or not.

Then, for JavaScript components with parameterized paths, we likewise duplicate the components as referenced. So `/[user]/widget.js` becomes `/a/widget.js`, `/b/widget.js`, _etc._ And therefore any relative `FileAttachment` reference in `widget.js` will “just work”.

In addition, we can allow parameterized components and JavaScript code blocks in parameterized pages to refer to parameter values, say as `observable.params.user`. Referenced parameter values are baked-in to the code when built (in `transpileModule`, perhaps using esbuild’s [**define** option](https://esbuild.github.io/api/#define), in a similar fashion to `process.env` references in Next.js).

With a bit more magic, this approach could even allow JavaScript components to refer to parameter values in arguments to `FileAttachment`, since the parameter values are known statically. This is nice because now we can have `/[user].md` load the corresponding `/[user].json`, rather than being forced to use `/[user]/index.md`. It also means you could have `/[property]/[user].md` load `/[user].json` which is not (otherwise) statically expressible as a relative path.

docs/[dir]/index.md
```diff
@@ -0,0 +1,31 @@
...
+
+display(foo);
+```
+
+```js
+display(`${observable.params.dir}.json`);
+```
+
+```js
+display(await FileAttachment("./" + observable.params.dir + ".json").json());
```
Comment (mbostock) [inline docs/[dir]/index.md:28]: This is the cool part: you can refer to parameters when constructing `FileAttachment` paths because they’re statically baked into the code!

src/dataloader.ts
```diff
@@ -80,31 +82,35 @@ export class LoaderResolver {
...
-          preload: async () => archive,
-          inflatePath: targetPath.slice(archive.length - ext.length + 1),
-          path: join(this.root, archive),
-          root: this.root,
-          targetPath,
-          useStale
-        });
-      }
-      const archiveLoader = this.findExact(archive, {useStale});
+      const archiveLoader = this.findExact(archive, {useStale}) ?? this.findDynamic(archive, {useStale});
```
mbostock, line 85: I think it would be more correct to compute the Cartesian cross-product of archive extensions (.zip, .tar, etc.) and interpreter extensions (.js, .py, etc.) here so that a more-specific parameterized match takes precedence over a less-specific parameterized match with a different archive extension. For example, you’d want /foo/bar.tar to take precedence over /[param]/bar.zip.

Although I think this is actually already broken because crawling the parent folders currently requires an exact match in order to look for an archive/extractor; so parameterized archives probably don’t work as intended. We need use glob to match `[param]` directory names. This will require further investigation. There are a lot of combinations of dynamic routing to consider! 😓

src/dataloader.ts
```diff
@@ -68,9 +70,9 @@ export class LoaderResolver {
    * abort if we find a matching folder or reach the source root; for example,
    * if src/data exists, we won’t look for a src/data.zip.
    */
-  find(targetPath: string, {useStale = false} = {}): Loader | undefined {
-    const exact = this.findExact(targetPath, {useStale});
-    if (exact) return exact;
+  find(targetPath: string, {useStale = false} = {}): Asset | Loader | undefined {
+    const result = this.findExact(targetPath, {useStale}) ?? this.findDynamic(targetPath, {useStale});
```
mbostock, line 74: I expect with the new `route` we could combine these branches, too…

Fil (commented):
Fantastic. I also tested a parametrized stylesheet — using `<link rel>`.

I'm not sure if we can parametrize the front-matter yet—presumably this requires page loaders (#1216)?

mbostock:
I’m interested in feedback on the name of the **paths** config option.

I’m reasonably happy with the design. I think it’s both flexible and intuitive (as opposed to, say, explicitly enumerating each parameter’s values and using that to extrapolate the set of paths). It lets you express the pages you want and Framework figures out how these map to routes and parameter values. It lets you use the same parameter name for different routes with different values (_e.g._, `/products/[id].md` and `/customers/[id].md`), and have routes with multiple (nested) parameters without generating the Cartesian product of all parameter values (_e.g._, `/[customer]/[product].md`).

But it’s a little weird in that it lets you specify paths that _aren’t_ parameterized. And I think currently it will generate a page twice if it’s listed multiple times? And it’s a little weird that the name “paths” doesn’t reference “parameters”.

Originally I wanted to call it “pages”, but we already use that for the sidebar and pager. And in the embedded analytics future, I want to be able to extend it to enumerate parameterized routes that aren’t pages, for example for embeddable, parameterized JavaScript modules or data files.

Fil:
**routes** would feel more precise—less generic. We already use the word “path[s]“ in 17 places, whereas “route” is used exclusively in https://observablehq.com/framework/project-structure and for this same concept.

src/preview.ts
```diff
@@ -213,6 +210,11 @@ export class PreviewServer {
         console.error(error);
       }
       if (req.method === "GET" && res.statusCode === 404) {
+        if (req.url?.startsWith("/_file/") || req.url?.startsWith("/_import/")) {
```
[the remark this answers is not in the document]
Fil: We can generalize, but then, we should probably ignore files and directories in src/ that start with an underscore? Currently you can have a page at `/_npm` if you have a file in `src/_npm.md`.

src/javascript/params.ts
```diff
@@ -0,0 +1,51 @@
...
+
+export function checkParams(node: Node, input: string, params: Params): void {
+  for (const [name, param] of findParams(node, params, input)) {
+    param.value = params[name];
+  }
+}
+
+export function findParams(body: Node, params: Params, input: string): [name: string, node: ParamReference][] {
+  const matches: [string, ParamReference][] = [];
+  const references = findReferences(body, {filterReference: ({name}) => name === "observable"});
```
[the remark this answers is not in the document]
mbostock: I prefer it as a global.

observablehq.config.ts
```diff
@@ -2,6 +2,7 @@ import {existsSync} from "node:fs";
 import {readFile, readdir, stat} from "node:fs/promises";
 import {join} from "node:path/posix";
 import {formatPrefix} from "d3-format";
+import {themes} from "./docs/themes.md.ts";
```
mbostock, line 5: Fun that you can import from a page loader in your config file!

mythmon:
I agree with Fil about **paths** feeling very generic. I don't feel strongly either way about **routes**. What if the name somehow referenced that these were parameterized, dynamic, or generated? Some examples might be **dynamicRoutes**, **parameterizedPaths**, or **generatedPages**.

As far as the general approach of listing strings that then map back to the parameterized file paths, I have a couple thoughts thoughts: First, what would happen if one of the strings pointed to a real concrete file that doesn't have any parameters? Is that just a no-op? an error?

Second, and I could probably figure this out by reading code, is how non-parameter and parameter path segments interact. If I had both `data/[foo]/a.json.js` and `[bar]/static/a.json.js`, I'm not sure I could predict which one would be used. Does it go one path-segment at a time, so a fixed match to `data/` would ignore any files in other directories?

mbostock:
> If I had both `data/[foo]/a.json.js` and `[bar]/static/a.json.js`, I'm not sure I could predict which one would be used.

It’s hinted at above (see [Precedence](https://github.com/observablehq/framework/pull/1523#issuecomment-2233915522)) but the short answer is that the most specific match is given preference. So for `data/static/a.json` that matches both, `data/[foo]/a.json.js` would be given precedence over `[bar]/static/a.json.js` because for the first part of the path `data` is more specific than `[bar]`.

mbostock:
I’m leaning towards **dynamicPaths**…

yurivish:
This looks amazing! Out of curiosity (and because I think there might be a use case for it), do page loaders work together with parameterized routes to allow generating a unique page for each route?

mbostock:
Yes @yurivish! You can see an example here:

https://github.com/observablehq/framework/blob/mbostock/parameterized/docs/theme/%5Btheme%5D.md.ts

mbostock:
For future work, I think we should also consider automatically building pages that are discoverable through static links, including those specified in the **pages** config option.

docs/[dir]/dynamic.md.js
```diff
@@ -0,0 +1,23 @@
+import {parseArgs} from "node:util";
+
+const {
+  values: {dir}
+} = parseArgs({
+  options: {dir: {type: "string"}}
+});
+
+console.log(`# Hello ${dir}
```
Comment (Fil) [inline docs/[dir]/dynamic.md.js:9]: Are we adopting `console.log` (I thought we wanted to use `process.stdout.write`). I don't have a strong preference, but somehow a lot of users struggle with this because they're so used to `console.log` for debugging.

src/javascript/module.ts
```diff
@@ -204,36 +208,37 @@ export function getFileInfo(root: string, path: string): FileInfo | undefined {
   return entry;
 }
 
-function resolveJsx(path: string): string | null {
-  return !existsSync(path) && path.endsWith(".js") && existsSync((path += "x")) ? path : null;
+export function findModule(root: string, path: string): RouteResult | undefined {
+  const ext = extname(path) || ".js";
+  const exts = [ext];
+  if (ext === ".js") exts.push(".jsx");
+  return route(root, path.slice(0, -ext.length), exts);
```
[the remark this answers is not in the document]
Fil: I'm not sure if this is resolved?

Fil:
This is wonderful to use. Parametrized pages with SSR can be as simple as:

```
import { parseArgs } from "node:util";
import * as d3 from "d3";

const {
  values: { projection: projName },
} = parseArgs({
  options: { projection: { type: "string" } },
});

const path = d3.geoPath(projection === "orthographic" ? d3.geoOrthographic() : d3.geoMercator());

const quakes = await fetch(
  "https://earthquake.usgs.gov/earthquakes/feed/v1.0/summary/all_week.geojson"
).then((response) => response.json());

console.log(`# Quakes (${projection})

<svg width="960" height="500" xmlns="http://www.w3.org/2000/svg">
  <path stroke="currentColor" d="${path({ type: "Sphere" })}"></path>
  <path stroke="currentColor" d="${path(d3.geoGraticule10())}"></path>
  <path fill="red" fill-opacity="0.8" d="${path(quakes)}"></path>
</svg>

`);
```

src/config.ts
```diff
@@ -274,6 +298,18 @@ export function normalizeConfig(spec: ConfigSpec = {}, defaultRoot?: string, wat
...
+function getDefaultPaths(root: string): string[] {
+  return Array.from(visitMarkdownFiles(root))
+    .filter((path) => !isParameterizedPath(path))
+    .map((path) => join("/", dirname(path), basename(path, ".md")));
+}
+
+function normalizePaths(spec: unknown): Config["paths"] {
+  if (typeof spec === "function") return spec as () => AsyncIterable<string>;
+  if (spec == null || spec === false) return async function* () {};
+  throw new Error(`invalid paths: ${spec}`);
```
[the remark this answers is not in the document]
mbostock: I was implementing @mythmon’s suggestion. The problem with also support iterables is that it leads to an explosion of possible types: iterable, async iterable, function that returns iterable, async function that returns iterable, function that returns async iterable, async function that returns async iterable, maybe some other that I didn’t think of. I don’t want to handle all those cases for convenience and instead am only supporting a single patttern of a function that returns an async iterable.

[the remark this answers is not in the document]
mbostock: I’ve added support for **dynamicPaths** as an iterable — but not those other variations. So your choice is now either a (synchronous) iterable, possibly in conjunction with top-level await, or a function that returns an async iterable if you want to load the list of dynamic paths asynchronously. That feels like enough. 😅

Fil:
What is our story with params values being something other than "a subset of ascii without slashes"? I've tried dynamic routing with `["/😂", "été", "Привіт", …]` and get various types of errors:

* `URI malformed` for http://127.0.0.1:3000/%E9t%E9
* Invalid character in header content ["Location"] for http://127.0.0.1:3000/%F0%9F%98%82 and http://127.0.0.1:3000/%D0%9F%D1%80%D0%B8%D0%B2%D1%96%D1%82

We can restrict these (the idea being that the params should be more "ids" than "content", but I'm not sure where this restriction currently happens.

mbostock:
@Fil That’s just a bug with the Preview server trying to redirect to the canonical path; fixed in e6e3600.

Fil:
I tried it on pangea and got the following error:

`SyntaxError: The requested module 'glob' does not provide an export named 'globSync'`

The version of glob that I have in yarn.lock is ancient:

```
glob@^8.0.3:
  resolved "https://registry.yarnpkg.com/glob/-/glob-8.1.0.tgz#d388f656593ef708ee3e34640fdfb99a9fd1c33e"
```

as required by `@rollup/plugin-commonjs@^25.0.7`

Shouldn't glob be a dependency, not a devDependency?

Fil:
When a page loader crashes, the preview server _sometimes_ crashes. 

I think the reproduction is the following:

create crash.md.js, with contents 

`process.stdout.write(`# Hello, world`);`

open the preview server to /crash, the page shows _hello world_. Now edit the contents to be:

`process.stdout.write(`# Hello ${name}`);`

the server crashes.


<details><summary>trace</summary>

```
load /crash.md → [stale] file:///Users/fil/Source/pangea/src/crash.md.js:1
process.stdout.write(`# Hello ${name}`);
                                ^

ReferenceError: name is not defined
    at file:///Users/fil/Source/pangea/src/crash.md.js:1:33
    at ModuleJob.run (node:internal/modules/esm/module_job:218:25)
    at async ModuleLoader.import (node:internal/modules/esm/loader:329:24)
    at async loadESM (node:internal/process/esm_loader:28:7)
    at async handleMainPromise (node:internal/modules/run_main:113:12)

Node.js v20.11.1
error in 56ms: loader exited with code 1
/Users/fil/Source/pangea/node_modules/@observablehq/framework/src/loader.ts:471
      throw new Error(`loader exited with code ${code}`);
            ^


Error: loader exited with code 1
    at CommandLoader.exec (/Users/fil/Source/pangea/node_modules/@observablehq/framework/src/loader.ts:471:13)
    at <anonymous> (/Users/fil/Source/pangea/node_modules/@observablehq/framework/src/loader.ts:404:11)

Node.js v20.11.1
error Command failed with exit code 1.
```
</details>


Note: it does _not_ crash if you directly load the erroneous page, in that case it fails gracefully as expected.

mbostock:
Yes, glob should be a dependency, not a devDependency. We are using glob@10; glob@11 drops support for older versions of Node but we haven’t done the same (yet) in Framework so we shouldn’t use it. I’ve fixed it in the latest commit.

I’ll investigate the Preview server crash when a page loader crashes next. I’ve seen that before, too.

mbostock:
Okay, fixed the page loader crash during preview. I also fixed the logic so that the exponential backoff on reopening the socket works as intended when the socket opens and then immediately crashes because of the page loader; we should only reset the socket reopen delay when the page loader is successful.

Fil:
I don't think it's an issue of any consequence, but it's fun to see that if you have a dynamic route `[x][y].md.js` and call `/framework`, the letters are distributed as `x=framewor, y=k`.

I've tried to pass an option value that starts with a dash (e.g., `rotate,[x].md.js` with url `rotate,-45`). It's not clear to me if this is safe to do, or if we're unwittingly passing the raw value when we should be passing a quoted value?

Anyhow my hint is that I could not read the negative number from parseArgs, because it errors with 'ERR_PARSE_ARGS_INVALID_OPTION_VALUE' (it's still possible to look at argv directly).

```js
import {parseArgs} from "node:util";
const {
  values: {x}
} = parseArgs({
  options: {x: {type: "string"}}
});
//const x = +process.argv[1 + [...process.argv].findIndex((d) => d === "--x")]; // this works but…

process.stdout.write(`# Hello ${JSON.stringify(process.argv)}

${JSON.stringify({x})}`);
```

Fil:
I've tried to break it some more, but I can't find how

docs/page-loaders.md.js
```diff
@@ -0,0 +1,73 @@
+process.stdout.write(`---
+keywords: server-side rendering, ssr
+---
+
+# Page loaders <a href="https://github.com/observablehq/framework/pull/1523" class="observablehq-version-badge" data-version="prerelease" title="Added in #1523"></a>
+
+Page loaders are like [data loaders](./data-loaders), but for generating (or “server-side rendering”) dynamic pages. Page loaders are programs that emit [Markdown](./markdown) to standard out, and have a double extension starting with \`.md\`, such as \`.md.js\` for a JavaScript page loader or \`.md.py\` for a Python page loader.
```
Fil, line 7:
```suggestion
Page loaders are like [data loaders](./data-loaders), but for generating (or “server-side rendering”) pages dynamically. Page loaders are programs that emit [Markdown](./markdown) to standard out, and have a double extension starting with \`.md\`, such as \`.md.js\` for a JavaScript page loader or \`.md.py\` for a Python page loader.
```

(the idea is that once the page is generated, it's just the same as any other page, static or not — the viewer has no way to tell the difference)

docs/page-loaders.md.js
```diff
@@ -0,0 +1,73 @@
...
+
+By “baking” dynamically-generated content into static Markdown, you can further improve the performance of pages since the content exists on page load rather than waiting for JavaScript to run. You may even be able to avoid loading additional assets and JavaScript libraries.
+
+For example, to render a map of recent earthquakes into static SVG using D3:
+
+~~~js run=false
+import * as d3 from "d3-geo";
+import * as topojson from "topojson-client";
```
Fil, line 15: this example does not work out of the box. You get
`Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'topojson-client' imported from`.

and you can't fix it (apparently?) by using `"npm:topojson-client"`, you have to `yarn add topojson-client`

mbostock: You need to `yarn add --dev d3-geo topojson-client`. WAI.

docs/page-loaders.md.js
```diff
@@ -0,0 +1,73 @@
...
+
+<div class="tip">
+
+When using page loaders, keep an eye on the generated page size, particularly with complex maps and data visualizations in SVG. To keep the page size small, consider server-side rendering a low-fidelity placeholder and then replacing it with the full graphic using JavaScript on the client.
+
+</div>
+
+## Parameterized pages
+
+Page loaders often use [parameterized routes](./params) to generate multiple pages from a single program. Parameter values are passed as command-line arguments. In a JavaScript page loader, you can use [\`parseArgs\`](https://nodejs.org/api/util.html#utilparseargsconfig) from \`node:util\` to parse command-line arguments. You can then bake parameter values into the resulting page, or reference them dynamically in JavaScript using \`observable.params\`.
```
Fil, line 43:
```suggestion
Page loaders often use [parameterized routes](./params) to generate multiple pages from a single program. Parameter values are passed as command-line arguments. In a JavaScript page loader, you can use [\`parseArgs\`](https://nodejs.org/api/util.html#utilparseargsconfig) from \`node:util\` to parse command-line arguments. You can then bake parameter values into the resulting page, or reference them dynamically in client-side JavaScript using \`observable.params\`.
```

reinforces this part, which I think will be a FAQ: `observable` is only available in the client-side JS. Server-side you have to read the script’s argv.

docs/page-loaders.md.js
```diff
@@ -0,0 +1,73 @@
...
+  options: {product: {type: "string"}}
+});
+
+process.stdout.write(\`# Hello $\{product}
+
+Or alternatively, $\\{observable.params.product}.
+\`);
+~~~~
+
+Framework’s [theme previews](./themes) are implemented as parameterized page loaders; see [their source](https://github.com/observablehq/framework/blob/main/docs/theme/%5Btheme%5D.md.ts) for a practical example.
```
Fil, line 60:
```suggestion
Within this document, the [theme previews](./themes) are implemented as parameterized page loaders; see [their source](https://github.com/observablehq/framework/blob/main/docs/theme/%5Btheme%5D.md.ts) for a practical example.
```

Fil:
That's all I have!

Fil (approved):
I've disabled auto-merge to give you a chance to see the remaining suggestion.

This is a HUGE step forward.

mbostock:
I’m going to move the parameterized page loader docs into the parameterized routes docs so that it isn’t duplicated in both places. That makes the page loader docs pretty short, but that’s not too surprising since they’re very closely related to data loaders. I also want to expand the examples a bit and make them slightly more realistic.

docs/page-loaders.md.js
mbostock: I think it’s clear that “theme previews” are part of documentation, no?

PR #1526, opened 2024-07-20
close floating sidebar on escape

mbostock:
This small enhancement makes it so that when the sidebar is opened in floating mode (with a drop shadow), Escape will close the sidebar in the same fashion as clicking on the backdrop. (Note that this sidebar can either be opened persistently with Command-B, or temporarily opened with Command-K.)


Fil (approved):
Of note, on a small screen with the search field focused, this means we now need to hit Esc twice: to empty the field, then to close the sidebar. I'm fine with this new behavior (which feels less surprising, in a way).

PR #1527, opened 2024-07-22
document sample datasets

### LINKED ISSUE #817: document example data sets
the [example data sets](https://github.com/observablehq/stdlib/blob/6ad71d98c4362d71e086fd5a0b5f97128cb9d906/src/library.js#L61-L72) defined in [stdlib](https://github.com/observablehq/stdlib) are available in `framework` but not documented, as pointed out in [this post](https://talk.observablehq.com/t/observable-framework-where-did-penguins-come-from/8792).  i'm not sure exactly where these would fit in the structure, but they do seem like a good thing to add.

~~i think [this notebook](https://observablehq.com/@observablehq/sample-datasets) is the most canonical documentation for the list.~~

https://github.com/observablehq/sample-datasets

Fil:
The page is indexed but not listed — the goal is that when someone searches for "penguins" they can find the provenance.

closes #817

### DIFF
--- a/docs/sample-datasets.md
+++ b/docs/sample-datasets.md
@@ -4,68 +4,68 @@ index: true
 
 # Sample datasets
 
-To help you get started, Observable Framework provides a handful of sample datasets by default. If you reference any of these variables (`aapl`, `alphabet`, `cars`, `citywages`, `diamonds`, `flare`, `industries`, `miserables`, `olympians`, `penguins`, `pizza`, or `weather`) in your code, their definition defaults to a Promise loading the corresponding dataset with the appropriate method (csv, tsv, json), and returning the data. This makes it easier to get started, with simpler examples of charts and data transformations.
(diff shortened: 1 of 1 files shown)


docs/sample-datasets.md
```diff
@@ -0,0 +1,72 @@
...
+
+# Sample datasets
+
+To help you get started, Observable Framework provides a handful of sample datasets by default. If you reference any of these variables (`aapl`, `alphabet`, `cars`, `citywages`, `diamonds`, `flare`, `industries`, `miserables`, `olympians`, `penguins`, `pizza`, or `weather`) in your code, their definition defaults to a Promise loading the corresponding dataset with the appropriate method (csv, tsv, json), and returning the data. This makes it easier to get started, with simpler examples of charts and data transformations.
+
+The source data files are downloaded on demand from the [@observablehq/sample-datasets](https://www.npmjs.com/package/@observablehq/sample-datasets) npm package, and served from your project’s cache. Note that these names are not “reserved”: you can define `alphabet` or `industries` as you see fit, if you need a variable or a function with that name in your project.
+
+The following lists the provenance of each of the sample datasets:
+
+## aapl.csv
```
mbostock, line 13:
```suggestion
## aapl
```

I think we should list them by the symbol identifier rather than the name of the file (which is invisible to the user).

mbostock (changes requested):

mbostock (approved):

PR #1541, opened 2024-07-29
@observablehq/inputs 0.11

mbostock:
Ref. https://github.com/observablehq/inputs/releases/tag/v0.11.0

We still need to bake-in Inputs for two reasons: we define `Inputs.file` using our implementation of `AbstractFile`, and we modify the stylesheet to use theme variables. Fixes `file.type` and `file.lastModified` when using `Inputs.file`.


Fil:
We should probably document the new select option in inputs/table?

Fil (approved):

PR #1543, opened 2024-07-29
feat: Add Google BigQuery data loader

dirkbosman:
Extended the loaders by adding a Google BigQuery loader with a concrete example on how data can be used. I hope this can also helpful for others to :)


dirkbosman:
Let me know if anything else is missing, then I can include it :)

examples/loader-google-bigquery/README.md
```diff
@@ -0,0 +1,7 @@
+[Framework examples →](../)
```
Fil, line 1: this file needs to be updated

examples/loader-google-bigquery/package.json
```diff
@@ -0,0 +1,24 @@
...
+  "scripts": {
+    "clean": "rimraf src/.observablehq/cache",
+    "build": "rimraf dist && observable build",
+    "dev": "observable preview",
+    "deploy": "observable deploy",
+    "observable": "observable"
+  },
+  "dependencies": {
+    "@google-cloud/bigquery": "^7.8.0",
+    "@observablehq/framework": "^1.8.0",
```
Fil, line 13: this could use the current version of framework, or "latest"

examples/loader-google-bigquery/src/index.md
```diff
@@ -0,0 +1,168 @@
...
+
+export async function runQuery(query) {
+  const [rows] = await bigQueryClient.query({ query });
+  return rows;
+}
+```
+
+<div class="note">
+
+To run this data loader, you’ll need to install `@google-cloud/bigquery`, `d3-dsv`, `d3-time-format`, and `dotenv` using your preferred package manager such as npm or Yarn.
```
Fil, line 96:
```suggestion
This data loader depends on `@google-cloud/bigquery`, `d3-dsv`, `d3-time-format`, and `dotenv`, which we reference in `package.json`.
```

I wonder if this kind of wording might be better? Since there's nothing special to do for the user besides installing the project.

examples/loader-google-bigquery/src/index.md
```diff
@@ -0,0 +1,168 @@
...
+
+```js echo
+Inputs.table(covidStats)
+```
+
+Lastly, we can pass the table to `Plot.plot` to make a simple line chart.
+
+```js echo
+display(
+  Plot.plot({
```
Fil, line 136: That's a bit too much code for a simple line chart, we can trim it down quite a bit (I can help with that).

examples/loader-google-bigquery/src/index.md
```diff
@@ -0,0 +1,168 @@
...
+        curve: "step"
+      }),
+      Plot.ruleY([0])
+    ]
+  })
+);
+```
+
+### The output should look as follows:
+![image info](covid_stats.png)
```
Fil, line 168: I don't think it's necessary to add a static image since the chart is displayed just above?

examples/loader-google-bigquery/src/data/covidstats_it.csv.js
```diff
@@ -0,0 +1,54 @@
...
+    AND date BETWEEN '2020-05-01 00:00:00 UTC' AND '2020-05-15 00:00:00 UTC'
+  GROUP BY 1,2
+  ORDER BY 1 ASC;
+`;
+
+(async () => {
+  try {
+    const rows = await runQuery(query);
+
+    console.log("Query Results:", rows);
```
Fil, line 28: The data loader should only send the formatted csv to stdout, which is then cached in a file; if you want to add "logs", use console.warn so as to send them to stderr instead (they will be reflected in the preview server logs).

examples/loader-google-bigquery/src/data/covidstats_it.csv.js
```diff
@@ -0,0 +1,54 @@
...
+
+    if (rows.length === 0) {
+      console.log("No data returned from the query.");
+      return;
```
Fil, line 32: here it's better to throw an error, so that framework knows that the loader has failed

examples/loader-google-bigquery/src/data/covidstats_it.csv.js
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    const csvData = csvFormat(
```
Fil, line 35: this is the text you can send to stdout (with `process.stdout.write(csvData)` or equivalently `console.log(csvData)`. No need to send it to a file on disk, Framework will do that for you.

examples/loader-google-bigquery/src/data/covidstats_it.csv.js
```diff
@@ -0,0 +1,54 @@
...
+
+    const filePath = join(__dirname, "covidstats_it.csv");
+    await fs.writeFile(filePath, csvData);
+    console.log(`CSV file written successfully to ${filePath}`);
+
+    const fileContent = await fs.readFile(filePath, "utf8");
+    const parsedData = csvParse(fileContent, { typed: true });
+
+    console.log("Parsed Data:", parsedData);
+
```
Fil, line 50:
```suggestion
```

this can be removed

examples/loader-google-bigquery/src/data/covidstats_it.csv.js
```diff
@@ -0,0 +1,54 @@
...
+  } catch (error) {
+    console.error('Error running query or processing file:', error);
```
Fil, line 52: It would be simpler (and, more importantly, better) to not try/catch, but let the loader error naturally; this way Framework would know that it has errored, would not cache the result, and would stop the build (if building), or retry (if in preview).

examples/loader-google-bigquery/src/data/covidstats_it.csv.js
```diff
@@ -0,0 +1,54 @@
...
+import { promises as fs } from "fs";
+import { dirname, join } from "path";
+import { fileURLToPath } from "url";
+
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = dirname(__filename);
+
+const query = `
+  SELECT 
+    FORMAT_TIMESTAMP('%Y-%m-%d', date) as formatted_date,
```
Fil, line 12: I think this should be called `date` throughout.

Fil (changes requested):
Thanks, this is a very useful addition!

I think you can simplify it quite a bit by having only one page (`index.md`), and by taking full advantage of Framework's data loader and Plot's defaults to trim a lot of the code. (I can do it if you prefer.)

examples/loader-google-bigquery/package.json
mbostock: This is fine as-is. I don’t think we shops recommend using latest over a tilde or caret version range which is more common and safer. Even though this doesn’t refer to the latest version of Framework, you will still get the latest version that is compatible with this range when you install, and this is the same pattern that is used by the other examples and we want to be consistent.

dirkbosman:
> Thanks, this is a very useful addition!
> 
> I think you can simplify it quite a bit by having only one page (`index.md`), and by taking full advantage of Framework's data loader and Plot's defaults to trim a lot of the code. (I can do it if you prefer.)

I actually mirrored the Google Analytics Loader file/dir-setup, but if you think it is helpful to keep it simpler, let's go with that route rather 🙏

examples/loader-google-bigquery/README.md
dirkbosman: Something like below could work? Just need to test that the "view live" when final changes is merged :)

# Google BigQuery Data Loader

View live: <https://observablehq.observablehq.cloud/framework-example-loader-google-bigquery/>

This Observable Framework example demonstrates how to write a JavaScript data loader that fetches metrics from the Google Bigquery API  using the [Google BigQuery Data Node.js Client](https://cloud.google.com/nodejs/docs/reference/bigquery-connection/latest). The data loader fetches the `confirmed_cases` metric (based on a publicly available Covid-19 BigQuery dataset) for a specified date range and then displays the data in a line chart. The data loader lives in [`src/data/covidstats_it.csv.js`](./src/data/covidstats_it.csv.js) and uses the helper [`src/data/google-bigquery.js`](./src/data/google-bigquery.js).

examples/loader-google-bigquery/src/index.md
dirkbosman: Agreed :) The reason why I added it, is that it is always night to upfront see what users want to replicate and whether the result what they need to achieve is the same. Not absolutely necessary and might even be better to move this image to the README.

examples/loader-google-bigquery/src/data/covidstats_it.csv.js
dirkbosman: This is a nice to have, can be removed :)

examples/loader-google-bigquery/src/data/covidstats_it.csv.js
dirkbosman: Thanks, very useful 🙏

examples/loader-google-bigquery/src/data/covidstats_it.csv.js
dirkbosman: Can be removed yes 🙏

examples/loader-google-bigquery/src/data/covidstats_it.csv.js
dirkbosman: Yes, can be updated if needed

examples/loader-google-bigquery/src/index.md
dirkbosman: Yes, can you trim please 🙏

examples/loader-google-bigquery/README.md
Fil: looks good!

dirkbosman: Will you directly do all suggested trims/changes and merge after (if possible)? I don't mind, let me know what works best for you / what will be the fastest for you 🙏

Fil: yes I can take it from here

PR #1557, opened 2024-08-07
hash global imports

### LINKED ISSUE #1555: Built files in `_observablehq`, `_import`, and `_npm` can change content without changing their name
Framework tries to maintain the contract (via content hashes and exact versions) that built files in `_observablehq`, `_import`, and `_npm` should change their name if their content changes, allowing immutable caching in the browser for better performance. However, in practice, this contract can be broken for several reasons.

1. Files in `_observablehq` can change between Framework releases, and because they bake-in the resolved versions of npm dependencies (such as via the recommended libraries and `FileAttachment` APIs).
2. Files in `_import` can change because they bake-in the resolved versions of npm dependencies, and these resolved versions are not considered as part of the content hash (which only considers the _original_ path of the import, pre-resolution, such as `npm:@observablehq/plot`).
3. Files in `_npm` can change because they bake-in the resolved versions of (transitive) npm dependencies. For example, `@observablehq/plot` depends on `d3`, and the version of `d3` can change independently of `@observablehq/plot`.

To maintain this contract, Framework will need to apply hashes (or other unique identifiers) to both `_observablehq` and `_npm`, and also fix how the hash is computed in `_import`.

mbostock:
Fixes #1555. TODO:

- [x] Content hashes for `_import` JavaScript incorporate transitive npm and node resolutions
- [x] Content hashes for `_npm` and `_node` JavaScript
- [x] Content hashes for `_observablehq` JavaScript
- [x] Content hashes for `_observablehq` stylesheets
- [x] No content hashes for `_npm` and `_node` non-JavaScript (_e.g._, DuckDB-Wasm’s WebAssembly)
- [x] Fix resolution of `observablehq:search` from within `search-init.ts`


Fil:
Awesome!

I'm seeing a problem when I use this branch to build and deploy pangea:

`yarn observable deploy --build --message "trying 1557"`

```
■  Error: ENOENT: no such file or directory, copyfile
│  'src/.observablehq/cache/_observablehq/minisearch.json' ->
│  'dist/_observablehq/minisearch.json'
```

Similarly (but slightly different), when I run `yarn docs:build` the `_observablehq/search.js` file is missing from the built site (but in fact there is `_observablehq/search.cbdff8eb.js`; and the index is `_observablehq/minisearch.94f07110.json`).

mbostock:
Thanks for testing. I’ll see if I can isolate and reproduce.

Fil:
In pangea the [home page](https://github.com/Fil/pangea/blob/main/src/index.md?plain=1) is calling `import.meta.resolve("observablehq:minisearch.json")`.

mbostock:
Okay, very close now. I’ve fixed the bug you found in Pangea, but the last remaining (known) problem is that we don’t apply content hashing (or more generally the configurable `resolveImport` hook) to `search-init.js` and `sidebar-init.js` inline scripts. The former has an import of `observablehq:search` that is therefore broken. But it shouldn’t be too hard to fix.

mbostock:
Hmm, the built artifacts now look correct, but unfortunately DuckDB-Wasm is broken and I’m not sure why yet. 🤔

mbostock:
Okay, the problem is that non-JavaScript files in `_npm` are getting content hashes applied, but we should only be applying content hashes to JavaScript files. (The only reason we apply content hashes to `_npm` is that we bake-in references to resolved versions of dependencies, but only JavaScript files can have dependencies.)

mbostock:
Fixed, finalllly. 😓

src/client/search.js
```diff
@@ -8,7 +8,7 @@ const resultsContainer = document.querySelector("#observablehq-search-results");
 const activeClass = "observablehq-link-active";
 let currentValue;
 
-const index = await fetch(import.meta.resolve(global.__minisearch))
+const index = await fetch(import.meta.resolve("observablehq:minisearch.json"))
```
mbostock, line 11: Now that we’ve implemented configurable import resolution in `rollupClient`, we don’t need to special-case the `minisearch.json` resolution (which was already content-hashed since it changes often); we can use the normal mechanism for import resolution instead.

src/deploy.ts
```diff
@@ -62,7 +62,7 @@ export interface DeployEffects extends ConfigEffects, TtyEffects, AuthEffects {
   output: NodeJS.WritableStream;
   visitFiles: (root: string) => Generator<string>;
   stat: (path: string) => Promise<Stats>;
-  build: ({config, addPublic}: BuildOptions, effects?: BuildEffects) => Promise<void>;
+  build: ({config}: BuildOptions, effects?: BuildEffects) => Promise<void>;
```
mbostock, line 65: I removed the `addPublic` flag which was only used by tests to reduce the complexity of building (slightly). There’s still some public-specific logic in `build-test.ts`.

mythmon:
I ran all the examples with this new version, and I found at least one problem (some of the errors aren't obvious, and I'm still working through them).

When running a couple examples, including `markdown-it-wikilinks` I get errors like this:

```
markdown-it-wikilinks $ ../../dist/bin/observable.js build --debug
found 3 pages in src
parse src/Wiki.md in 92ms
parse src/Wiki_Links.md in 56ms
parse src/index.md in 42ms
bundle /_observablehq/client.js → src/.observablehq/cache/_observablehq/client.js
bundle /_observablehq/runtime.js → src/.observablehq/cache/_observablehq/runtime.js
bundle /_observablehq/stdlib.js → src/.observablehq/cache/_observablehq/stdlib.js
build /_observablehq/theme-air,near-midnight.css → dist/_observablehq/theme-air,near-midnight.e68849dc.css
alias /_observablehq/client.js → /_observablehq/client.7daf1d7b.js
alias /_observablehq/runtime.js → /_observablehq/runtime.ef648817.js
alias /_observablehq/stdlib.js → /_observablehq/stdlib.3cc0bf9d.js
build /_observablehq/client.js → dist/_observablehq/client.7daf1d7b.js
build /_observablehq/runtime.js → dist/_observablehq/runtime.ef648817.js
build /_observablehq/stdlib.js → dist/_observablehq/stdlib.3cc0bf9d.js
render /Wiki →
Unexpected error: Entry module "../../dist/client/sidebar-init.js" cannot be external.
The full error follows

file:///home/mythmon/src/framework/node_modules/rollup/dist/es/shared/parseAst.js:394
    const errorInstance = Object.assign(new Error(base.message), base);
                                        ^

Error [RollupError]: Entry module "../../dist/client/sidebar-init.js" cannot be external.
    at getRollupError (file:///home/mythmon/src/framework/node_modules/rollup/dist/es/shared/parseAst.js:394:41)
    at error (file:///home/mythmon/src/framework/node_modules/rollup/dist/es/shared/parseAst.js:390:42)
    at ModuleLoader.loadEntryModule (file:///home/mythmon/src/framework/node_modules/rollup/dist/es/shared/node-entry.js:19133:20)
    at async Promise.all (index 0) {
  code: 'UNRESOLVED_ENTRY'
}

Node.js v20.13.1
```

mythmon:
I finished combing through all the other projects. That's the only error I found, though it shows up on multiple examples: `markdown-it-wikilinks`, `us-dams`, and `responsive-iframe`.

mbostock:
Thanks for finding the bug. I have a guess as to what’s wrong and I’ll diagnose later tonight if I’m able.

Fil:
I've been confused by a situation where I tried to add one word to a page, then deployed. The word ("sempervirens") did not appear in search. I removed the word, re-deployed, it still didn't appear. Strangely, the deploy mentioned:

```
◇  0 uploaded, 2109 unchanged, 2109 total.
```

The URL of search.js did not change (https://observablehq.observablehq.cloud/pangea/_observablehq/search.09b1e05e.js), same hash, but its contents changed, linking either to 
https://observablehq.observablehq.cloud/pangea/_observablehq/minisearch.2098a410.json (which has the word)
or to 
https://observablehq.observablehq.cloud/pangea/_observablehq/minisearch.daae6d48.json (which has not)

That resource is returned with the header `max-age=7776000, immutable`.

As a consequence the request for minisearch sometimes (depending on the browser cache?) hits the old minisearch.json URL, and search is not in sync, or even broken if the index has been deleted and 404s.

Here you can see side by side the same situation in Firefox and Chrome: in Firefox (left) I have the up-to-date version and search works; in Chrome (right) search is broken, still linking to a now-deleted https://observablehq.observablehq.cloud/pangea/_observablehq/minisearch.b82bf641.json

minisearch-cache-404


Note: maybe this was already happening before this PR, and it's not caused by it. But I had never witnessed it, and it seems at least conceptually related.

mbostock:
Okay, I will also check whether the hash for search.js considers the contents of minisearch.json. It should!

mbostock:
@Fil I fixed that issue. Now working on the one @mythmon raised.

Fil:
Thanks! A slightly annoying consequence is that if I change one word in one page, deploying means uploading all the pages again (and the new index), because they now reference a different hash of `search.js`. Before, it would just upload the new index and the updated page. But I understand it's unavoidable in the name of consistency.

Fil:
There is an added risk with async loading, since linked files can disappear when a new version is deployed.

However, I tested what happens in the following scenario:
- I load a page.
- I deploy a new version
- I click "search" on the previously loaded page.

Apparently it seems to work… there is maybe a grace period during which the older files are still available?

mbostock:
Yes @Fil that’s an unavoidable consequence if we want files in `_observablehq` to be immutable.

But I’m noting now that they _aren’t_ actually marked immutable currently: they are instead marked as `public, max-age=0, must-revalidate`. So, I think that means we could go in one of two directions:

1. We could make `_observablehq` immutable (like `_npm`, `_node`, and `_import`) as assumed in #1555 and improve performance with better caching.
2. We could leave `_observablehq` as mutable, and remove the content hashes from these assets to reduce the frequency of global changes.
3. We could specifically leave `_observablehq/minisearch.json` as mutable, and serve everything else in `_observablehq` as immutable.

I think my preference is (1) and that we shouldn’t over-index on the Pangea hundreds of pages/documentation use case. Conceivably (3) is better but it adds some complexity and more work, and I think we should try to fix this bug today at the latest. (2) seems to penalize common-case performance to cater to the Pangea case.

Fil:
I don't want to overindex on pangea, we should only use it as a test bench. Consistency has benefits, too, and the inconvenience I noticed is not unbearable (even for pangea on ADSL, re-uploading everything only takes a minute).

Also, the constraints that content-indexing impose are more general, not limited to the search index. To make sure the site doesn't break between updates, all async dependencies of a web page must be accessible for a duration equal to the lifetime of the page (which some users might want to be long, say a week, if the page contains features that are loaded on user demand). A different fix could be to make the build configurable—for example it could be better on some web hosts to build references as `file.ext?hash` instead of `file.hash.ext` (or even to skip the hash altogether), or to have a hashed version with a fallback on the non-hashed version. This feels out of scope for this PR.

In the meantime, for the search index, I'm OK with either choice; strategy 2 corresponds to what we had up to now, but strategy 1 feels stronger (more opinionated, more consistent). Since this PR auto-merges, I assume this means we'll go with strategy 2, and maybe reinforce/rethink this later.

Fil (approved):

mythmon (commented):
Retroactive, but this looks good to me too. I verified that the bug I found before was fixed. I also read through all the code and all of the changes make sense.

test/deploy-test.ts
```diff
@@ -1054,10 +1006,10 @@ describe("deploy", () => {
...
       .handlePostDeploy({projectId: DEPLOY_CONFIG.projectId, deployId})
       .expectFileUpload({deployId, path: "index.html", action: "upload"})
-      .expectFileUpload({deployId, path: "_observablehq/theme-air,near-midnight.css", action: "skip"})
-      .expectFileUpload({deployId, path: "_observablehq/client.js", action: "skip"})
-      .expectFileUpload({deployId, path: "_observablehq/runtime.js", action: "skip"})
-      .expectFileUpload({deployId, path: "_observablehq/stdlib.js", action: "skip"})
+      .expectFileUpload({deployId, path: "_observablehq/theme-air,near-midnight.e68849dc.css", action: "skip"})
+      .expectFileUpload({deployId, path: "_observablehq/client.c35dfd1a.js", action: "skip"})
+      .expectFileUpload({deployId, path: "_observablehq/runtime.c45c72e0.js", action: "skip"})
+      .expectFileUpload({deployId, path: "_observablehq/stdlib.1b8a97c3.js", action: "skip"})
```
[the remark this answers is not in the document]
mbostock: I tried to do that but the `index.html` action is `upload`, not `skip`…

PR #1560, opened 2024-08-08
ignore rel=external links

mbostock:
Allows opting out of Framework’s link resolution by adding `rel="external"` to any element. For example:

```html
<a href="/account" rel="external">test</a>
```


pstuffa (approved):

Fil:
👍 

Add documentation?

PR #1565, opened 2024-08-09
normalize `_observablehq` hashes during testing

mbostock:
This rewrites the build tests so that the hashes for anything under `_observablehq` aren’t dependent on the actual content. This avoids needing to regenerate the test snapshots whenever we change Framework’s client code (_e.g._, #1564).


Fil:
Nice. This doesn't seem to cover some of the esm files? (_e.g._, `test/output/build/files/_npm/d3-dsv@3.0.1/cd372fb8.js`)

mbostock:
> This doesn't seem to cover some of the esm files?

What do you mean by “doesn’t cover”? Those files are generated and covered by the test. The caveat is that we don’t actually talk to jsDelivr; we use undici to mock the request, which returns an empty file. That’s in the `mockJsDelivr` function.

mbostock:
I’m only trying to normalize against Framework’s client code changing, since that causes churn in the test snapshots. The _npm content won’t cause churn because we already mock the jsDelivr API and hence control when it changes.

Fil:
I meant they still had a 8 letter hash. (But since it's the hash of empty, it doesn't matter.)

Fil (approved):

mbostock:
Oh, yeah… though now I’m curious why it isn’t `e3b0c442` which is the empty hash? I’ll investigate before merging.

mbostock:
Ah, okay so it’s the hash of the empty hash. 😂 

```js
createHash("sha256").update("e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855").digest("hex") // cd372fb85148700fa88095e3492d3f9f5beb43e555e5ff26d95f5a6adc36f8e6
```

PR #1568, opened 2024-08-09
clean before building

### LINKED ISSUE #1538: The automatic re-build on deploy should be clean
As @mcglincy demonstrates, when we ask whether you’d like to rebuild as part of deploying, we should really be doing a clean build so that we don’t include old files in the published output.

mbostock:
Fixes #1538. Now `observable build` is always clean.

Leaving as draft because I think we want a `--clean` flag to clean automatically, and without it we prompt you before cleaning the existing output root (if it exists) (and error in CI if the output root is dirty and `--clean` or `--no-clean` wasn’t specified).

And maybe `--clean` isn’t the right name because a “clean” build will still respect the cache so it’s not fully clean? Sigh.


mcglincy:
Code changes LGTM.

If "clean" seems too inaccurate, maybe something "clean-dist" (and then later "clean-cache", "clean-all"?) could be used.

mcglincy (approved):
LGTM

mbostock:
Thinking about this some more, and I think cleaning the output directory is the right default, and probably doesn’t even need a flag. For precedence, Vite’s build command has an `emptyOutDir` option that likewise defaults to true, but only if the output directory is within the project root (so as to avoid accidentally deleting something important, such as /). I’ll add that to the PR and commit.

srenatus:
Sorry for posting to an old issue, but it seems like the best spot:

> Thinking about this some more, and I think cleaning the output directory is the right default, and probably doesn’t even need a flag.

I've come here looking for a flag 😅 The problem is this: I'm using framework in a concourse pipeline, and the output directory is declared as _the task's output_, so that _only that data is moved to the next task_. Concourse seems to do that by mounting something into the output location, and that something cannot be removed:

```
> observable
> observable "build"

(node:2607) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
(Use `node --trace-deprecation ...` to show where the warning was created)
Warning: the config file is missing the root option, which specifies the path to
the source root. The recommended source root is "src"; however, since docs
exists and was previously the default for this option, we will use "docs". You
can suppress this warning by specifying root: "docs" in the config file.

Unexpected error: EBUSY: resource busy or locked, rmdir 'dist'

Tip: To see the full stack trace, run with the --debug flag.


If you think this is a bug, please file an issue at
↳ https://github.com/observablehq/framework/issues
```

I think I can just use another directory within the output directory, so it's probably not a big deal.

PR #1579, opened 2024-08-14
fix: convert output to default to the source root

### LINKED ISSUE #838: convert should read the config, and respect its output option

Fil:
Note that it is diverging a bit from the ideation in #838. Because, in the config file, `output` defaults to `dist` (it is where the built files go); in the case of the `convert` command, I don't think we want that, but rather default to the source root, aka the `root` option (which defaults to `src`).

If the `--output` command-line option is passed explicitly, it must be considered relative to the cwd, not relative to the root. But this means we're comfortable with running this command from anywhere in the project, so if it's not passed explicitly, we make sure to use the source root relative to the root. Otherwise you would end up creating a new "src" folder inside your cwd.

(I was surprised in my tests that the cwd is incorrect when calling `yarn tsx ../src/bin/observable.ts convert`; I ended up testing manually the built script.)

(I'm making this PR against cobus/convert-documentation but it could be merged afterwards.)

closes #838


src/bin/observable.ts
```diff
@@ -1,4 +1,5 @@
 #!/usr/bin/env node
+import {join} from "node:path";
```
mbostock, line 2:
```suggestion
import {join} from "node:path/posix";
```

I don’t think it matters here but stylistically we use posix by default and when we do need the os-specific behavior we use `import op from "node:path";`.

mbostock (changes requested):
Functionally good but one quibble on the import. 👍 

Also I deleted the command-line flags section from the docs because it’s getting too long and was buried and I think it’s more appropriate to use `--help` to get help for the command-line flags. (I did include an inline mention of `--output` though.)

Also, can you hold off on merging this today so that we can document `convert` as it currently exists rather than blocking the docs & `convert` blog post on a new release of Framework? I think the current behavior is acceptable since you often have to move things around anyway.

mbostock:
This needs a re-merge from main to review again. Let me see if I can do it…

mbostock (approved):

PR #1581, opened 2024-08-15
bigquery loader example

Fil:
import of #1543 by @dirkbosman

deployed at https://observablehq.observablehq.cloud/framework-example-loader-google-bigquery/


mbostock (changes requested):
The deploy is missing the [parent config](https://github.com/observablehq/framework/blob/main/examples/observablehq.config.js) (and thus the header shared across all examples). Can you fix and then re-deploy? You have to point the deploy at the parent config using `--config` and then it imports the child config based on the current working directory. We should have instructions somewhere…

examples/loader-google-bigquery/src/data/covidstats_it.csv.js
```diff
@@ -0,0 +1,22 @@
...
+    AND date BETWEEN '2020-05-01 00:00:00 UTC' AND '2020-05-15 00:00:00 UTC'
+  GROUP BY 1,2
+  ORDER BY 1 ASC;
+`;
+
+(async () => {
+  const rows = await runQuery(query);
+  if (rows.length === 0) throw new Error("No data returned from the query.");
+  process.stdout.write(csvFormat(rows));
+})();
```
mbostock, line 22:
```suggestion
const rows = await runQuery(query);
if (rows.length === 0) throw new Error("No data returned from the query.");
process.stdout.write(csvFormat(rows));
```

I believe we can use top-level await here (and already do in other examples).

mbostock: I’d also prefer to inline `query` instead of having a separate variable.

examples/loader-google-bigquery/src/data/google-bigquery.js
```diff
@@ -0,0 +1,20 @@
+import "dotenv/config";
+import {BigQuery} from "@google-cloud/bigquery";
+
+const {BQ_PROJECT_ID, BQ_CLIENT_EMAIL, BQ_PRIVATE_KEY} = process.env;
+
+if (!BQ_PROJECT_ID) throw new Error(".env missing BQ_PROJECT_ID");
+if (!BQ_CLIENT_EMAIL) throw new Error(".env missing BQ_CLIENT_EMAIL");
+if (!BQ_PRIVATE_KEY) throw new Error(".env missing BQ_PRIVATE_KEY");
```
mbostock, line 8:
```suggestion
if (!BQ_PROJECT_ID) throw new Error("missing BQ_PROJECT_ID");
if (!BQ_CLIENT_EMAIL) throw new Error("missing BQ_CLIENT_EMAIL");
if (!BQ_PRIVATE_KEY) throw new Error("missing BQ_PRIVATE_KEY");
```

These environment variables don’t _have_ to be defined in the `.env` (and typically won’t in a CI environment — the `.env` is instead of local development).

examples/loader-google-bigquery/src/data/google-bigquery.js
```diff
@@ -0,0 +1,20 @@
...
+
+const bigQueryClient = new BigQuery({
+  projectId: BQ_PROJECT_ID,
+  credentials: {
+    client_email: BQ_CLIENT_EMAIL,
+    private_key: BQ_PRIVATE_KEY.replace(/\\n/g, "\n")
```
mbostock, line 14:
```suggestion
    private_key: BQ_PRIVATE_KEY
```

We shouldn’t do this here. (See the google-analytics example for precedent.) If you want to use escape sequences in your environment variables in your `.env`, you need to put double quotes around the value.

examples/loader-google-bigquery/src/index.md
```diff
@@ -0,0 +1,113 @@
...
+
+<div class="warning">
+
+The `.env` file should not be committed to your source code repository; keep your credentials secret.
+
+</div>
+
+<div class="note">
+
+See the [Google Bigquery API Quickstart](https://cloud.google.com/bigquery/docs/authentication) for how to create the service account needed to access the Google BigQuery API after it has been enable in the GCP console. The following [Google codelab](https://codelabs.developers.google.com/codelabs/cloud-bigquery-nodejs#0) might be helpful walk-through before getting started. Although the data in this example lives in the public database `bigquery-public-data.covid19_italy`, you still need to set up a service account and authenticate into one of your projects to access it.
```
mbostock, line 78:
```suggestion
See the [Google BigQuery API Quickstart](https://cloud.google.com/bigquery/docs/authentication) for how to create the service account needed to access the Google BigQuery API after it has been enable in the GCP console. The following [Google codelab](https://codelabs.developers.google.com/codelabs/cloud-bigquery-nodejs#0) might be helpful walk-through before getting started. Although the data in this example lives in the public database `bigquery-public-data.covid19_italy`, you still need to set up a service account and authenticate into one of your projects to access it.
```

examples/loader-google-bigquery/src/index.md
```diff
@@ -0,0 +1,113 @@
...
+    private_key: BQ_PRIVATE_KEY.replace(/\\n/g, "\n")
+  }
+});
+
+export async function runQuery(query) {
+  return (await bigQueryClient.query({query})).rows;
+}
+```
+
+For the data loader to authenticate with the Google Bigquery API, you will need to set several environment variables containing secret credentials. If you use GitHub, you can use [secrets in GitHub Actions](https://docs.github.com/en/actions/security-guides/using-secrets-in-github-actions) to set environment variables; other platforms provide similar functionality for continuous deployment. For local development, we use the `dotenv` package, which allows environment variables to be defined in a `.env` file which lives in the project root and looks like this:
```
mbostock, line 62:
```suggestion
For the data loader to authenticate with the Google BigQuery API, you will need to set several environment variables containing secret credentials. If you use GitHub, you can use [secrets in GitHub Actions](https://docs.github.com/en/actions/security-guides/using-secrets-in-github-actions) to set environment variables; other platforms provide similar functionality for continuous deployment. For local development, we use the `dotenv` package, which allows environment variables to be defined in a `.env` file which lives in the project root and looks like this:
```

BigQuery is intercapped.

examples/loader-google-bigquery/src/data/google-bigquery.js
Fil: my mistake

Fil:
> We should have instructions somewhere…

it's just this

`yarn deploy --config ../observablehq.config.js --build`

but I keep forgetting to do it on the first deploy :(

mbostock (approved):
💯

dirkbosman:
Amazing, thanks again @mbostock & @Fil - I hope this opens up a whole lot of new insights in/through Framework & BigQuery private & public datasets <3

PR #1585, opened 2024-08-17
Datawrapper loader example

palewire:
This proposal adds an example that generates and then embeds a chart created by [Datawrapper](https://datawrapper.de).

In addition to the Python dependencies, it requires a Datawrapper API key available in the Python environment as `DATAWRAPPER_ACCESS_TOKEN`.
Screen Shot 2024-08-17 at 6 50 30 AM

Screen Shot 2024-08-17 at 6 50 41 AM


examples/loader-datawrapper/src/index.md
```diff
@@ -0,0 +1,83 @@
...
+dw.publish_chart(chart_id)
+
+# Write the chart's embed code to stdout
+html = dw.get_iframe_code(chart_id, responsive=True)
+sys.stdout.write(html)
+```
+
+<div class="note">
+
+To run this data loader, you will need to create an API token in Datawrapper and set it as an environment variable named `DATAWRAPPER_ACCESS_TOKEN`.
```
Fil, line 57: It would be nice to have detailed instructions to create the venv. In particular since this repo expects the venv to live in .venv (judging from the .gitignore file).

I tried this:

```sh
python -m venv .venv
source .venv/bin/activate
pip3 install -r requirements.txt
```

but it failed with:
```
error: externally-managed-environment

× This environment is externally managed
```

To make this work I had to run the last command with the `--break-system-packages` option 😅:
```sh
python -m venv .venv
source .venv/bin/activate
pip3 install -r requirements.txt --break-system-packages
```

(Maybe it's just me—I appear to be doomed when I need to install python stuff.)

examples/loader-datawrapper/src/index.md
```diff
@@ -0,0 +1,83 @@
...
+
+You’ll also need python3 and the pandas and datawrapper modules installed and available on your `$PATH`.
```
Fil, line 59: this is what the command above ensures. Is there a way to make this easier by conflating this note and the tip below into a single procedure?

examples/loader-datawrapper/src/index.md
```diff
@@ -0,0 +1,83 @@
...
+
+</div>
+
+<div class="tip">
+
+We recommend using a [Python virtual environment](https://observablehq.com/framework/loaders#venv), such as with venv or uv, and managing required packages via `requirements.txt` rather than installing them globally.
+
+</div>
+
+The above data loader lives in `data/chart.html.py`, so we can load the data using `data/chart.html` with `FileAttachment`:
```
Fil, line 69: I hoped we could do something like this instead:

~~~
html`<iframe src="${FileAttachment("data/chart.html").href}">`
~~~

but it doesn't work because we don't have any means to access the iframe's width and height. It will get better when we have SSR (#1216).


In the meantime, I'd recommend doing it all in one cell, something like:

~~~
```js echo
const div = display(document.createElement("DIV"));
div.innerHTML = await FileAttachment("data/chart.html").text();
```
~~~

examples/loader-datawrapper/src/data/chart.html.py
```diff
@@ -0,0 +1,47 @@
...
+
+# Pull out the chart's unique identifier
+chart_id = chart_config["id"]
+
+# Publish the chart
+dw.publish_chart(chart_id)
+
+# Write the chart's embed code to stdout
+html = dw.get_iframe_code(chart_id, responsive=True)
+sys.stdout.write(html)
```
Fil, line 47:
```suggestion
sys.stdout.write(html)

```

examples/loader-datawrapper/requirements.txt
```diff
@@ -0,0 +1,2 @@
+pandas
+datawrapper
```
Fil, line 2:
```suggestion
datawrapper

```

examples/loader-datawrapper/src/index.md
Fil, line 57: For the **DATAWRAPPER_ACCESS_TOKEN** can we specify where to create it (https://app.datawrapper.de/account/api-tokens) and the scopes we have to create? I initially tried with a narrow scope (Chart | read  write) and the data loader created the chart, but couldn't read it back. It worked when I enabled more reading capabilities (Folder | read  write; Theme | read; User | read; Visualization | read), but I'm not sure which of those were strictly necessary.

It could also be useful to explain how to set it inside the venv.

Fil (changes requested):
The example demonstrates the technique, but the resulting chart is not really showing the best of what datawrapper offers—why should a user invest so much effort into creating a basic bar chart? It might be worth looking for a more visually striking example, where going through datawrapper really makes sense (and shines).

A different way of using Datawrapper inside of Framework might be to use the datawrapper UX to create a chart, then export to an iframe. Maybe we just need a comment about this.

Another “tip” that we could add is that this approach creates a new chart in DataWrapper each time the data loader runs. Although it is unavoidable, it can be surprising.

Fil:
To help reviewing I have deployed the project to https://observablehq.observablehq.cloud/framework-example-loader-datawrapper/

examples/loader-datawrapper/src/index.md
palewire: I'm agnostic on Python package managers, so I'm open to handling this however you want.

My routine is to use pipenv. So, in my world, I run something like this:

```bash
pipenv install datawrapper pandas
pipenv run npm run dev
```

I put the requirements file in my proposals because I saw it that way in other examples and presumed it was your preference. It's not the canonical way to run pipenv, [but it's supported](https://docs.pipenv.org/en/latest/advanced.html#importing-from-requirements-txt), so here's how I did things in this case.

```bash
pipenv shell
pipenv install -r requirements.txt
```

I don't have any experience with venv but there must be a way to make it work without resorting to hacks or unusual options. Let me know how you'd like to proceed.

palewire:
I believe I addressed all of your requests. I'm open to whatever approach you want to take regarding the Python environment.

examples/loader-datawrapper/src/index.md
palewire: I'm open to whatever environment management technique you want to recommend. I typically use a `.env` file with the standard pipenv conventions that is excluded from the git repository. In this case, for simplicity's sake I just did something like:

```bash
DATAWRAPPER_ACCESS_TOKEN=xxxyyzzz pipenv run npm run dev
```

Fil:
I'm sorry I edited my main comment just after I sent it, but you might have read it by email?
https://github.com/observablehq/framework/pull/1585#pullrequestreview-2245372013

Also you have removed the "fixed" data/chart.html that I added. We need it so people can see what to expect without actually having to connect to the API (and so we can also rebuild all the examples when necessary).

examples/loader-datawrapper/src/index.md
Fil: wow, I didn't know about `pipenv`; maybe it can save me from my personal python hell!

palewire:
I did miss that, sorry.

I'm open to another example. I was thinking, "Let's keep it super simple" and give the user a bare-bones starting point.

I'm happy to change the chart to something else. What appeals to you? A [dressed-up stacked bar](https://academy.datawrapper.de/article/151-examples-of-datawrapper-stacked-bar-charts), a [range plot](https://academy.datawrapper.de/article/281-examples-of-datawrapper-dot-charts), a [dozen donuts](https://academy.datawrapper.de/article/188-examples-of-datawrapper-pie-donut-charts)?

Fil:
If you ask me, I like "Germany is the third-oldest country in the world". But it's only an opinion :)

palewire:
Alrighty. I've updated it with the dot range plot. 
Screenshot from 2024-08-19 16-30-59

palewire:
Ready for your review again, @Fil

Fil:
I'm a bit reluctant to publish it as is, because there are a few things that ought to be better:

*dark mode*

I've found a way to enable dark="auto" in the chart's preferences (editing the published chart), but I don't know how to pass this "render" option in the API call that creates the chart. Maybe it's not available in the python API?

*responsive*

Datawrapper has this concept of responsive charts, where the height depends on the (reactive) width. This example instead uses a fixed height, which ends up hiding some information (in our example, the footer) in a mobile view.

I've created a helper function to solve both problems, though. I'll push it here in a moment.


*tons of new charts*

It's kind of a bummer that it creates a new chart each time it runs. This makes it quite damaging for <del>the environment</del> your account, and the Datawrapper admins might not like that; more importantly maybe, it makes a lousy user experience as it prevents you from making the chart better by editing it (post-creation) with the DataWrapper UI.

Ideally what I'd like to do is have a unique code in the script; if a chart with that code already exists, we reuse it, otherwise we populate it. If the chart depends on the data, then the code can be computed as a hash of the data state.

I have not implemented that (yet), it's more on the python side.

Capture d’écran 2024-08-21 à 15 59 16

Fil:
I've redeployed to https://observablehq.observablehq.cloud/framework-example-loader-datawrapper/

Note that this helper function should be able to accommodate several charts on the page — maybe that's something we could add to the example too.

Fil:
> what I'd like to do is have a unique code in the script; if a chart with that code already exists, we reuse it, otherwise we populate it

We could search for a chart that has that id set in a given hidden field, but I don't see if it's a possibility with the current API.

palewire:
Thanks for this detailed feedback. I'm going to try to address your concerns piecemeal. The most recent commit is my first effort. I believe I have configured the Python chart configuration to activate Datawrapper's dark mode, which is off by default. 

I'll loop back on your other items soon.

palewire:
@Fil, here's a thought on the responsive iframe issue. Would we be better off switching to the web component, as described here: https://blog.datawrapper.de/web-component-embedding/

Fil:
I think you're right!

Also, I have tried `pipenv`, and it's great. Maybe this is an opportunity to write down how it works in this case.

For me it was:

```sh
# https://docs.pipenv.org/#install-pipenv-today
brew install pipenv 
# https://docs.pipenv.org/basics/#importing-from-requirements-txt
pipenv install -r ./requirements.txt
# https://app.datawrapper.de/account/api-tokens
echo 'DATAWRAPPER_ACCESS_TOKEN="****"' > ./.env
# https://docs.pipenv.org/basics/
pipenv run yarn dev
```

palewire:
Do you have an established pattern for templating a script tag to be injected into the page? I tried passing the chart ID out of the data loader and then printing that into the page with a ${x} style template literal, but that's not working. 

I was thinking something like this, but I'm not sure how that kind of thing fits into your framework.

``````
```js
const chartId = await FileAttachment("data/chart.html").text()
```

<div>
   <script type="text/javascript" defer 
        src="https://datawrapper.dwcdn.net/${chartId}/embed.js?v=14">
   </script>
   <noscript>
      <img src="https://datawrapper.dwcdn.net/${chartId}/full.png" alt="[aria-description]" />
   </noscript>
</div>
``````

**Comment by palewire** [comment]
I think I got it working and solved the responsive bug with:
``````
```js echo
const chartId = await FileAttachment("data/chart.txt").text()
const tag = display(document.createElement('script'));
tag.setAttribute('src', `https://datawrapper.dwcdn.net/${chartId}/embed.js?v=14`);
```
``````

Let me know what you think. If we have this one, I will move onto the final issue: Endless chart creation.

**Comment by Fil** [comment]
Yes it looks much simpler.

One issue that I had solved with my system but has a regression now, is when the page sets a color mode that is not the same as the browser's preference.

If you set this in the front-matter

```
---
theme: dark
---
```

and visit with a browser with light mode preferences, the chart is generated in light mode, and some text becomes unreadable. Same issue if you reverse the situation (theme: light, and a dark mode browser).

I think this is an upstream bug with datawrapper, though? We correctly set `:root {color-scheme: dark;}` and it should respect that, if set, over the browser preference. Or maybe I'm missing something?

palewire:
I'm not a dark mode expert, so I'll defer to you on this. I can contact the datawrapper support team after we finish this ticket, if you think that's the right path. 

On Fri, Aug 23, 2024, at 7:06 AM, Philippe Rivière wrote:
> 
> 
> Yes it looks much simpler.
> 
> One issue that I had solved with my system but has a regression now, is when the page sets a color mode that is not the same as the browser's preference.
> 
> If you set this in the front-matter
> 
> `---
> theme: dark
> ---
`

Fil:
Inspecting the script, I see it only reads `window.matchMedia("(prefers-color-scheme: dark)").matches`. Probably something to report upstream.

palewire:
Good catch. I'll send an email to the Datawrapper support team reporting the bug.

Separately, I've patched the loader to "get or create" the chart, rather than create a new one each time. 

Are you good with that edit? If so, anything else you'd like to see before merging?

Fil:
I think we only need the built chart.txt (so that in the future we can rebuild the project without datawrapper credentials).

What do you think of calling this "datawrapper-api" rather than "loader-datawrapper"? (It's not really "loading data", even though it uses the data loader architecture.)

Also, shouldn't we present the simpler version at the beggining of the page, explaining how a user can create a chart in datawrapper, then embed it in a framework page with a simple `<script>`. And then start the more powerful approach by saying "but here's how you can also create the chart from your framework code…". (Most people will probably only ever need the first approach, and for the others it can be used as a gentle introduction.)

palewire:
I've added the cached chart.txt file and renamed the example as "datawrapper API.

I'm not sure I know what you mean by "the simpler version," but I took at what I think you mean. A web component that does not rely on the data loader is at the top. It is then followed by the Python code, which is then embedded with a variation on the snippet at the top. Is that what you mean?

Fil:
Yes, exactly!

Fil:
Reading https://blog.datawrapper.de/web-component-embedding/ I found out how to solve the dark mode issue, by adding data-dark="true" or "false". I'll fix in a moment (also doing some light copy-editing).

Fil:
hmmm, I was trying to embed another chart and I'm getting errors on nki3d and K2J9t, looks like in these instances the script tries to get the data from my own server…?

palewire:
Huh. Now you're past my knowledge. The chart isn't so old that it predates web components. Maybe we found another bug?

Fil:
I don't know what I was doing wrong, but K2J9t and nki3d now work as expected. I've added a different chart at the top for some diversity, polished the prose a bit, and added a helper function that handles dark mode in sync with Framework (bypassing the bug we've noticed). I think it's close to done!

Fil:
Redeployed at https://observablehq.observablehq.cloud/framework-example-datawrapper-api/

palewire:
Looks good. I'm ready to ship it when you are. Where is that `dark` variable being set?

Fil:
It's a reactive variable available in all pages. See https://observablehq.com/framework/lib/generators#dark (I linked to this page in the "tip")

palewire:
Very cool

Fil:
I was ready to ship but as I was giving it another look I again saw the error where the first chart is broken as it tries to load the dataset from https://observablehq.observablehq.cloud/framework-example-datawrapper-api/dataset.csv instead of from datawrapper. But when I reloaded, it worked. I don't understand if it's a bug in the dw script, or in something we do…

palewire:
Out of curiousity, I cloned the chart. The data appear to be "locally" hosted rather than an external URL or Google Sheet or anything.

Screen Shot 2024-08-26 at 6 30 38 AM

I wonder if that's their bug, which if you screenshot I can report to their support channel.

In the meantime, I wonder if we can get something workable shipped if we use an example with an externally hosted CSV file, where there's a static URL in the configuration. You can see one of those here:

https://academy.datawrapper.de/article/346-how-to-create-a-live-updating-chart-or-table

Fil:
Here's my screenshot. The weirdest part of this is that it's not happening all the time, maybe once out of 10.

Capture d’écran 2024-08-26 à 14 13 13

palewire:
Gotcha. I see the bug too sometimes. I've sent your screenshot and a bug report with links to the Datawrapper support team.

Do you get the bug with the externally hosted data example on COVID I linked?

Fil:
No I don't see it on the covid example — it's loading https://static.dwcdn.net/data-feed/coronavirus/confirmed-per-day-selected-countries-rolling.csv?b&v=1724685720000

palewire:
Gotcha. I think that gives a little weight to my "the bug is with local data uploads" theory. I've got that filed in with the DW support team now.

You okay with release as is with the external fix? I suspect the problem will be fixed upstream shortly.

Fil:
I don't want someone to be excited to test this out and then get disappointed. Either we wait for the upstream fix (my preference), or we add a warning that there is an issue and explain how to go around it.

palewire:
I respect that for sure. Let's see if I hear anything from Datawrapper soon.

palewire:
Almost on cue, I heard this from the Datawrapper team:

> Sorry for the trouble here! Indeed, this is due to a bug on our side, where multiple charts embedded on one page using web components can contain conflicting versions of the embeds; the good news is, a simple republish of the offending chart does resolve this. 
> 
> I've just gone ahead and republished the chart with id OuHrk on our side, so everything should render correctly. Can you take a look and let me know if that helps, and if there are any other questions?

If that's correct and you can verify the republish has fixed things, I'd vote for merging. It would mean that this is only a bug in the edge case where you are trying to publish multiple version of years-old charts, which seems unusual.

Fil:
It's hard to prove a negative, but I've tested a dozen times and it's not coming back—and more importantly we can trust the Datawrapper team :)

So yeah, let's 🥳

palewire:
Great. If you need anything else from me, just shout. I'll promote on social media once it's live.

Fil (approved):

PR #1588, opened 2024-08-19
Documentation font changes

CobusT:
As per design's guidelines.

- Spline Sans Mono for the h1 headers
- Inter 16px for the body text, line-height of 1.5
- Font weight 500 for both
- Dark mode font color changed to #FEFEFE

### DIFF
--- a/docs/style.css
+++ b/docs/style.css
@@ -4,6 +4,8 @@
 @import url("https://fonts.googleapis.com/css2?family=Inter:ital,opsz,wght@0,14..32,100..900;1,14..32,100..900&family=Spline+Sans+Mono:ital,wght@0,300..700;1,300..700&display=swap");
 
 :root {
+  --serif: Inter, sans-serif;
+  --sans-serif: Inter, sans-serif;
   --theme-phosphate: #148576;
 }
 
@@ -54,34 +56,26 @@
 
 #observablehq-main {
   min-height: calc(100vh - 23.8rem);
-}
-
-#observablehq-main {
-  font-family: "Inter", sans-serif;
   font-size: 16px;
-  line-height: 1.5;
-  color: var(--theme-foreground);
-  background: var(--theme-background);
   font-weight: 500;
 }
 
-#observablehq-main p a {
-  color: var(--theme-foreground) !important;
-  text-decoration: underline !important;
+p a {
+  color: inherit;
+  text-decoration: underline;
 }
 
-#observablehq-main h1 {
+h1 {
   font-family: "Spline Sans Mono", monospace !important;
-  font-weight: 500 !important;
+  font-weight: 500;
   margin-bottom: 1rem;
 }
 
-#observablehq-main h2,
-#observablehq-main h3,
-#observablehq-main h4,
-#observablehq-main h5,
-#observablehq-main h6 {
-  font-family: "Inter", sans-serif;
+h2,
+h3,
+h4,
+h5,
+h6 {
   margin-bottom: 0.5rem;
 }
 
--- a/docs/style.css
+++ b/docs/style.css
@@ -54,7 +54,7 @@
   color: var(--theme-foreground-focus);
 }
 
-body {
+#observablehq-main {
   font-family: var(--sans-text);
   font-size: 16px;
   font-weight: 500;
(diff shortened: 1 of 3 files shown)


Fil:
preview https://observablehq.observablehq.cloud/framework-preview-1588/

docs/style.css
```diff
@@ -1,6 +1,20 @@
 @import url("observablehq:default.css");
 @import url("observablehq:theme-air.css") (prefers-color-scheme: light);
 @import url("observablehq:theme-near-midnight.css") (prefers-color-scheme: dark);
+@import url("https://fonts.googleapis.com/css2?family=Inter:ital,opsz,wght@0,14..32,100..900;1,14..32,100..900&family=Spline+Sans+Mono:ital,wght@0,300..700;1,300..700&display=swap");
+
+:root {
+  --serif: Inter, sans-serif;
```
mbostock, line 7: It seems strange to have `--serif` point to a sans-serif font. The only place Framework uses `--serif` internally is for the body font, so we should be able to override that without changing the definition of `--serif`. (And this only affects our own documentation site, so we don’t have to worry about this as a public API change.)

Also, I’m not sure if this matters, but typically we use a much longer set of system fonts for fallback rather than just `sans-serif` (or `serif`):

```css
  --serif: "Source Serif Pro", "Iowan Old Style", "Apple Garamond", "Palatino Linotype", "Times New Roman",
    "Droid Serif", Times, serif, "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol";
```

docs/style.css
```diff
@@ -36,12 +50,33 @@
...
 
-.observablehq-link-active a[target=_blank]:not(:hover, :focus)::after {
+.observablehq-link-active a[target="_blank"]:not(:hover, :focus)::after {
   color: var(--theme-foreground-focus);
 }
 
 #observablehq-main {
   min-height: calc(100vh - 23.8rem);
+  font-size: 16px;
+  font-weight: 500;
```
mbostock, line 60: These should probably be on the body instead of `#observablehq-main`.

docs/style.css
```diff
@@ -36,12 +50,33 @@
...
+}
+
+p a[href] {
+  color: inherit;
+  text-decoration: underline;
+}
```
mbostock, line 66: This should probably be applied to `a[href]` , not `p a[href]`. There’s no guarantee that links only appear within paragraph elements.

docs/style.css
```diff
@@ -36,12 +50,33 @@
...
+
+h1 {
+  font-family: "Spline Sans Mono", var(--monospace);
```
mbostock, line 69: I’d prefer a variable (`--sans-mono`) so we can keep the font definitions together up in `:root`.

docs/style.css
```diff
@@ -36,12 +50,33 @@
...
+  font-weight: 500;
+  margin-bottom: 1rem;
```
mbostock, line 71: What are these global margin changes to heading elements for? This will require extensively checking all the pages to make sure it works as intended.

Fil:
Related https://github.com/observablehq/plot/pull/2139

docs/style.css
Fil: I'm deferring to @CobusT on this one.

docs/style.css
Fil: makes sense, but then then it should be 

```
#observablehq-main a[href]:not(.observablehq-header-anchor)
```

we don't want to underline all the anchors in the copy text, nor the links in the site chrome (sidebar, toc…)

docs/style.css
CobusT: I started out like this, but body also affects the other elements on the page, like the left nav and the ToC. Those are overwritten by more specific css rules to get them back to only `var(--sans-serif)` so it probably doesn't matter, but I think `#observable-main` was the intended scope of the font change.

docs/style.css
CobusT: They are not needed here. I removed them.

Fil:
There are margins added to h2, h3, h4… just below. Remove them too?

docs/style.css
CobusT: Changing it back to only be for `#observablehq-main`

Fil (approved):

Issue #423, opened 2023-12-22
Preload and self-host web fonts based on custom styles


mbostock:
#414 adds support for custom styles, ~~but we still hard-code the use of Google Fonts~~. We should ~~instead~~ transform the CSS bundle somehow to flatten external imports so that preloading is accurate.

mbostock:
We likely want to introduce a `google-fonts:` protocol here so that users can control what’s self-hosted. It would be analogous to the self-hosting you get with `npm:`, versus when you link to `https://cdn.jsdelivr.net/` to load the resource at runtime from an external website. So maybe this is self-hosted:

```html
<link rel="stylesheet" href="google-fonts:/css2?family=Source+Serif+Pro:ital,wght@0,400;0,600;0,700;1,400;1,600;1,700&display=swap">
```

But this is hotlinked:

```html
<link rel="stylesheet" href="https://fonts.googleapis.com/css2?family=Source+Serif+Pro:ital,wght@0,400;0,600;0,700;1,400;1,600;1,700&display=swap">
```

Fil:
Does this need to be tied to the vendor? Or can we imagine calling this `fonts:`, with the vendor indicated elsewhere (in the configuration or in the URL rather than in the "protocol")?

e.g.
`<link rel="stylesheet" href="fonts:https://fonts.googleapis.com/css2?family=Source+Serif+Pro:ital,wght@0,400;0,600;0,700;1,400;1,600;1,700&display=swap">`

or if we want shorthand:

`<link rel="stylesheet" href="fonts:google/css2?family=Source+Serif+Pro:ital,wght@0,400;0,600;0,700;1,400;1,600;1,700&display=swap">`

Fil:
This needs to be rethought after #1597 solved the main issue (_i.e._, that google fonts was hard-coded and prevented complete self-hosting); maybe it's just the same as #1573 now?

PR #1589, opened 2024-08-20
An option to control the google font links

Fil:
The **extraHead** option allows to define an HTML fragment to add to the top of the page’s head. It defaults to the 3 links that allow to use the Source Serif Pro font from Google Fonts. (Prefer the **head** option, unless your goal is to remove or replace the default fonts.)

Note that its defaults will evolve once we self-host fonts (#423).


mbostock (commented):
Regarding the temporary nature of this feature:

If we don’t intend to support this as a public feature, we don’t need to write documentation for it (even commented-out). That’s just something we’ll need to delete when we remove the feature. Also, we should have the private config option(s) start with an underscore so that it’s more obviously signaled as as private (even if you don’t notice that it’s undocumented).

Regarding the feature itself:

There are two places where the hard-coded dependency on Google Fonts is expressed. One is the hard-coded stylesheet:

https://github.com/observablehq/framework/blob/d298eac9fd569ae14402a021791d8c75b46d3418/src/resolvers.ts#L110

The other is the hard-coded preconnect hint:

https://github.com/observablehq/framework/blob/d298eac9fd569ae14402a021791d8c75b46d3418/src/render.ts#L220

I was reacting to the fact that this (private) feature is called `googlefonts` but there’s nothing about it other than the preconnect hint that presupposes it points to Google Fonts. Maybe there could be a private option for `_extraStylesheets` and `_extraHead` that controls what is rendered in these two respective locations?

Fil:
Yes, for example here is the approach that vitepress supports:
https://vitepress.dev/reference/site-config#example-adding-google-fonts

which I implemented in the companion PR for Plot here
https://github.com/observablehq/plot/pull/2139/commits/071e9a19df5a8b2b08588cbaf397aa27595e0819

mbostock:
Yes so maybe we just need an `extraHead` option that does both the preconnect and the stylesheet, and then we can remove both hardcoded references to Google Fonts?

Fil:
I've implemented extraHead but now I'm thinking that the only thing we need is a flag to cancel the three links, since we can implement the new fonts in the **head** option?

mbostock:
Can you please re-merge main and resolve the conflicts? I can’t review this because it now includes a copy of all the other changes.

test/output/build/archives.posix/tar.html
```diff
@@ -2,10 +2,10 @@
 <meta charset="utf-8">
 <meta name="viewport" content="width=device-width, initial-scale=1, maximum-scale=1">
 <title>Tar</title>
-<link rel="preconnect" href="https://fonts.gstatic.com" crossorigin>
-<link rel="preload" as="style" href="https://fonts.googleapis.com/css2?family=Source+Serif+Pro:ital,wght@0,400;0,600;0,700;1,400;1,600;1,700&amp;display=swap" crossorigin>
+<link rel="preconnect" href="https://fonts.gstatic.com" crossorigin="">
+<link rel="preload" as="style" href="https://fonts.googleapis.com/css2?family=Source+Serif+Pro:ital,wght@0,400;0,600;0,700;1,400;1,600;1,700&amp;display=swap" crossorigin="">
```
Fil, line 6: note how this changes `crossorigin` to `crossorigin=""`

Fil:
I don't like the extra complexity, and suggest #1592 instead.

mbostock (commented):
Would this feature go away when we fix #423? Should we keep it as a private feature in the interim?

Fil:
I believe this is temporary, since #423 wants to detect and preload the default google font (if it's used).

(We'll probably have to move font definitions from `global.css` to `font.css` to make it possible to create a custom `style.css` that does not include the now unused import.)

My vote is to keep this private (and simple), #1595.

mbostock (approved):

mbostock (commented):
After more thought, I prefer #1597. 🙏

PR #1592, opened 2024-08-21
Adopt system font stacks

Fil:
Default to browser font stacks, and let the developer add external vendor fonts if they want, whether in `style.css` or in the **head** option —as we do in our docs— for better performance.

I think the "transitional" stack is the closest to the spirit that we had before by loading Source Serif Pro
https://github.com/system-fonts/modern-font-stacks?tab=readme-ov-file#transitional 

supersedes #1589 
removes one hurdle to #423

---

The following screenshots show what the default style looks like, in the themes, or when applied to the documentation (that is, if you remove our the custom css).

Air
Capture d’écran 2024-08-21 à 10 21 40

Near midnight
Capture d’écran 2024-08-21 à 10 21 54

Docs 1
Capture d’écran 2024-08-21 à 10 19 19

Docs 2
Capture d’écran 2024-08-21 à 10 19 28

Docs 3
Capture d’écran 2024-08-21 à 10 20 06

### DIFF
--- a/observablehq.config.ts
+++ b/observablehq.config.ts
@@ -86,7 +86,8 @@ export default {
     {name: "Contributing", path: "/contributing", pager: false}
   ],
   base: "/framework",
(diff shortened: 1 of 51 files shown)


mbostock (changes requested):
I want to keep Source Serif Pro as the default. It looks a lot better than the system stack, especially on non-macOS.

PR #1595, opened 2024-08-21
_ignoreDefaultFontLinks boolean private config option

Fil:
This seems more reasonable than introducing a new extraHead option; it doesn't change the default font stack that loads Source Serif (née Source Serif Pro, now [Source Serif 4](https://fonts.google.com/specimen/Source+Serif+4)) from Google Fonts.

supersedes #1589
supersedes #1592


mbostock (commented):
This doesn’t actually feel simpler to me; it feels like the same complexity but less powerful. (And I also quibble with the name `ignoreDefaultFontLinks` which I would probably invert to `includeDefaultFonts`, but that’s easily fixed.) I’m going to make some small tweaks to the `extraHead` branch which I think is still the way to go.

PR #1596, opened 2024-08-21
fix orphaned “code.”

mbostock:


mbostock:
Pretty trivial so just going to land this one. 😁

PR #1597, opened 2024-08-22
globalStylesheets config option

mbostock:
Alternative to #1589 and #1595. Adds a **globalStylesheets** option to the config which specifies a set of stylesheets to add to all pages (in addition to the **style** option which can be specified either globally or on a specific page). By default it’s Source Serif ~~Pro~~ 4 from Google Fonts. Also adds a little heuristic to add the preconnect for Google Fonts if a Google Fonts stylesheet is detected. Related #423.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -86,6 +86,7 @@ export interface Config {
   footer: PageFragmentFunction | string | null; // defaults to “Built with Observable on [date].”
   toc: TableOfContents;
   style: null | Style; // defaults to {theme: ["light", "dark"]}
+  defaultStylesheets: string[]; // defaults to Source Serif from Google Fonts
   search: SearchConfig | null; // default to null
   md: MarkdownIt;
   normalizePath: (path: string) => string;
@@ -99,6 +100,7 @@ export interface ConfigSpec {
   base?: unknown;
   sidebar?: unknown;
   style?: unknown;
+  defaultStylesheets?: unknown;
   theme?: unknown;
   search?: unknown;
   scripts?: unknown;
@@ -224,6 +226,10 @@ export function normalizeConfig(spec: ConfigSpec = {}, defaultRoot?: string, wat
       : spec.style !== undefined
       ? {path: String(spec.style)}
       : {theme: normalizeTheme(spec.theme === undefined ? "default" : spec.theme)};
+  const defaultStylesheets =
+    spec.defaultStylesheets === undefined
+      ? defaultDefaultStylesheets()
+      : Array.from(spec.defaultStylesheets as any, String);
   const md = createMarkdownIt({
     linkify: spec.linkify === undefined ? undefined : Boolean(spec.linkify),
     typographer: spec.typographer === undefined ? undefined : Boolean(spec.typographer),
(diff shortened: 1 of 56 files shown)


mbostock:
Also tacked on an upgrade to Source Serif 4 that @Fil noticed…

src/style/global.css
```diff
@@ -1,8 +1,7 @@
 :root {
   --monospace: Menlo, Consolas, monospace;
   --monospace-font: 14px/1.5 var(--monospace);
-  --serif: "Source Serif Pro", "Iowan Old Style", "Apple Garamond", "Palatino Linotype", "Times New Roman",
-    "Droid Serif", Times, serif, "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol";
+  --serif: "Source Serif 4", serif;
```
Fil, line 4:
```suggestion
  --serif: "Source Serif 4", Charter, "Bitstream Charter", "Sitka Text", Cambria, serif;
```

With this, when you set `globalStylesheets: []` or **when the user has no internet** (working offline, or can't connect to google fonts for some reason like corp. policy/firewall), you still get a decent font by default.

*source serif*
source serif

*charter*
charter

*times new roman*
times

src/config.ts
```diff
@@ -224,6 +226,10 @@ export function normalizeConfig(spec: ConfigSpec = {}, defaultRoot?: string, wat
       : spec.style !== undefined
       ? {path: String(spec.style)}
       : {theme: normalizeTheme(spec.theme === undefined ? "default" : spec.theme)};
+  const globalStylesheets =
+    spec.globalStylesheets === undefined
+      ? defaultGlobalStylesheets()
+      : Array.from(spec.globalStylesheets as any, String);
```
Fil, line 232:
```suggestion
  const globalStylesheets =
    spec.globalStylesheets === undefined
      ? defaultGlobalStylesheets()
      : !spec.globalStylesheets
      ? []
      : Array.from(spec.globalStylesheets as any, String);
```

Accept null and false rather than crashing? Or maybe we want to ignore the option in that case, rather than support falsy values?

Fil (changes requested):
Yeah I think this is the right call 😅

Fil:
I've added a bit of documentation

src/style/global.css
mbostock: I was just following Google’s recommendation but I guess they don’t try very hard for when the font doesn’t load. I’m not sure we need to address this case really either, but if you feel we should, I’m inclined to just revert to what we had before and address it separately.

src/config.ts
mbostock: Yeah my thinking was that null isn’t a valid specification, so why allow it? We don’t need to allow everything. But in that case perhaps we should throw a nice error instead of crashing arbitrarily.

Fil: I've added a nicer error.

Fil (approved):

src/config.ts
mbostock: I found two more places where we call Array.from unsafely, too (pages and scripts).

shadoof:
Apologies for possibly naive question, but when creating a new framework with latest version, there is no indication in `observablehq.config.js` for how the 'default' version of this option should be setup (i.e. so as to load Source Serif 4) which is exactly what I would like. Visual Studio Code offers ["src/styles.css"] as a value but since this file does not exist after `npx @observablehq/framework@latest create` is run, I'm guessing that's a problem?

And does the above issue discussion imply that `@import url('[google font]')` should work globally in this context in any  my array of 'globalStylesheets' ?

PR #1608, opened 2024-08-27
file.size

mbostock:
~~Marking as draft until https://github.com/observablehq/inputs/pull/266 is merged and Inputs 0.12.0 is released.~~ Done. ~~Also, I think I can clean up the implementation slightly (like maybe consolidating `getSourceLastModified` and `getSourceSize` into `getSourceInfo`).~~ Done. Also fixes an unfiled bug where `file.href` didn’t work for `Inputs.file`.

### DIFF
--- a/src/client/stdlib/inputs.js
+++ b/src/client/stdlib/inputs.js
@@ -1,5 +1,60 @@
-import {fileOf} from "@observablehq/inputs";
+import {file as _file} from "@observablehq/inputs";
 import {AbstractFile} from "npm:@observablehq/stdlib";
 
-export * from "@observablehq/inputs";
-export const file = fileOf(AbstractFile);
+export {
+  button,
+  checkbox,
+  radio,
+  toggle,
+  color,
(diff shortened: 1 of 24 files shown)


mbostock:
This might enable some smarter heuristics for DuckDBClient, such as using `CREATE TABLE` instead of `CREATE VIEW` for small files to avoid the overhead of repeated range requests.

src/client/stdlib/inputs.js
```diff
@@ -1,5 +1,60 @@
...
+  input,
+  bind,
+  disposal,
+  formatDate,
+  formatLocaleAuto,
+  formatLocaleNumber,
+  formatTrim,
+  formatAuto,
+  formatNumber
+} from "@observablehq/inputs";
```
mbostock, line 34: Repeating this is unfortunate but ESLint doesn’t like us exporting `file` twice. 😞

Fil (approved):

PR #1609, opened 2024-08-27
add meta generator tag

mbostock:
Intended primarily to assist in debugging so that we can tell the version of Framework used to build the site. Also [standard practice](https://html.spec.whatwg.org/multipage/semantics.html#meta-generator) for site generators.


Fil (approved):
hoping the tests pass on windows with cross-env

mbostock:
Thanks for the fixes!

PR #1616, opened 2024-08-27
restore Source Serif 4 for theme previews

### LINKED ISSUE #1615: Theme previews aren’t loading Source Serif 4
E.g., https://observablehq.com/framework/theme/air

mbostock:
Fixes #1615. Maybe not the best fix since this still loads Inter and Spline Sans Mono even though they are unused. That’s because the **globalStylesheets** option applies to the theme previews even though they don’t use our custom stylesheet. Maybe we should have a **stylesheets** front matter option that allows us to override the global stylesheets from the config, so we can restore it back to the default here?

### DIFF
--- a/docs/theme/[theme].md.ts
+++ b/docs/theme/[theme].md.ts
@@ -10,6 +10,8 @@ header: false
 sidebar: false
 ---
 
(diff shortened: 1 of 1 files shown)


Fil (approved):

Fil:
The use case is quite specific, I don't feel the need to generalize

PR #1617, opened 2024-08-28
use CREATE TABLE for small parquet

### LINKED ISSUE #1614: Optimize DuckDBClient’s handling of small Parquet files
We currently use `CREATE VIEW` for Parquet files:

```js
`CREATE VIEW '${name}' AS SELECT * FROM parquet_scan('${file.name}')`
```

Because views are not physically materialized, this can cause many [range requests](https://developer.mozilla.org/en-US/docs/Web/HTTP/Range_requests) as DuckDB reads the file. Sometimes the overhead of these range requests outweighs the benefits of not loading the entire file.

Using the new `file.size` property #1608, we could employ a heuristic that switches to `CREATE TABLE` instead of `CREATE VIEW` for “small” Parquet files. We might also want an option to provide an explicit hint to use a materialized table instead of a view.

mbostock:
Fixes #1614.

### DIFF
--- a/src/client/stdlib/duckdb.js
+++ b/src/client/stdlib/duckdb.js
@@ -258,7 +258,8 @@ async function insertFile(database, name, file, options) {
           });
         }
         if (/\.parquet$/i.test(file.name)) {
-          return await connection.query(`CREATE VIEW '${name}' AS SELECT * FROM parquet_scan('${file.name}')`);
+          const table = file.size < 10e6 ? "TABLE" : "VIEW"; // for small files, materialize the table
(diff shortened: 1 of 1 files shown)


Fil:
Did you have a way to estimate (guesstimate) that threshold value? Do we want to set up some kind of experiment?

mbostock:
No idea what the right threshold is. I’d guess it depends on your server too.

mbostock:
Is there a different value that you would prefer?

Fil:
No idea, I think it's probably a good guess. We could do do some experiments but it doesn't have to be blocking. Let's merge now and revisit later if needed.

Fil (approved):

PR #1618, opened 2024-08-28
use maxItems for unbounded select

mbostock:
Supersedes #1493.


mythmon (approved):
I tested this locally, and it worked great

Fil (commented):
Approved (ex post) 👍

PR #1620, opened 2024-08-29
1.11.0

mbostock:


Fil (approved):

Fil (changes requested):
I think I might have overlooked something in #1523; how do we tell framework to build the page `d3/index.html` when all we have is `d3/index.md.ts`?

PR #1622, opened 2024-08-31
include pages in default paths

mbostock:
If a page is listed in the **pages** config option, we should build it even if it’s not listed in **dynamicPaths**. Also, we shouldn’t build a page more than once.

### DIFF
--- a/observablehq.config.ts
+++ b/observablehq.config.ts
@@ -89,14 +89,12 @@ export default {
     {name: "Contributing", path: "/contributing", pager: false}
   ],
   dynamicPaths: [
-    "/page-loaders",
     "/theme/dark",
     "/theme/dark-alt",
     "/theme/dashboard",
     "/theme/light",
     "/theme/light-alt",
     "/theme/wide",
-    "/themes",
     ...themes.dark.map((theme) => `/theme/${theme}`),
     ...themes.light.map((theme) => `/theme/${theme}`)
   ],
--- a/src/config.ts
+++ b/src/config.ts
@@ -258,6 +258,7 @@ export function normalizeConfig(spec: ConfigSpec = {}, defaultRoot?: string, wat
   // end of the path. Otherwise, remove the .html extension (we use clean
   // paths as the internal canonical representation; see normalizePage).
   function normalizePagePath(pathname: string): string {
+    ({pathname} = parseRelativeUrl(pathname)); // ignore query & anchor
     pathname = normalizePath(pathname);
     if (pathname.endsWith("/")) pathname = join(pathname, "index");
     else pathname = pathname.replace(/\.html$/, "");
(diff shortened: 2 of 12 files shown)


Fil (commented):
Makes sense! I was wondering also if we shouldn't consider any path that has an appropriate double extension, e.g. `path.md.{js,ts,py…}` as a buildable page (unless marked as draft), instead of asking the user to add it explicitly to pages or dynamicPaths.

Fil:
The tests break because we now assume that there is a main /index page, which is not ("not really") the case in all the tests.

To fix:

```diff
diff --git a/src/pager.ts b/src/pager.ts
index 0d01f377..7e4572f2 100644
--- a/src/pager.ts
+++ b/src/pager.ts
@@ -42,7 +42,7 @@ export function findLink(path: string, options: Pick<Config, "pages" | "title">

 /**
  * Walks the unique pages in the site so as to avoid creating cycles. Implicitly
- * adds a link at the beginning to the home page (/index).
+ * adds a link at the beginning to the home page (/index), if it exists.
  */
 function walk(pages: Config["pages"], title = "Home"): Iterable<Iterable<Page>> {
   const pageGroups = new Map<string, Page[]>();
@@ -56,7 +56,8 @@ function walk(pages: Config["pages"], title = "Home"): Iterable<Iterable<Page>>
     pageGroup.push(page);
   }

-  visit({name: title, path: "/index", pager: "main"});
+  const home = pages.find((page) => page.path === "/index");
+  if (home) visit({name: title, path: "/index", pager: "main"});
```

and then we need to fix the test snapshots, because they now (correctly) don't include a pager when there is only one page which is not `/index`.

mbostock:
Thanks for the review and debugging!

I can add the check for the /index page. I thought I would have to check whether that page exists (using loaders.find), but you’re right, I should just check the list of pages already provided.

I debated scanning for static (non-parameterized) page loaders previously in visitMarkdownFiles and decided to punt on it, but maybe it would be easy to add. One challenge is that we can scan for the page loaders synchronously, but loading the page itself is asynchronous, which means we can’t easily read the page title or whether it is marked as draft in the current setup. But I think we don’t need to load the actual page in order to compute the list of paths here, since drafts are skipped later when building.

We’re out at the aquarium (I’m having a coffee) but will look when I get back tonight. Then we can include it in 1.11.0.

mbostock:
> `pages.find((page) => page.path === "/index")`

This doesn’t work because `pages` will never include `/index`. But we can use `loaders.find` instead.

mbostock:
I fixed the issue with `/index`.

I’d like to punt on finding the non-parameterized page loaders (double extensions). That’s harder because it requires knowing the registered list of interpreters which can be extended in the config. I would prefer to do that later.

Fil (approved):

PR #1625, opened 2024-09-01
Default workspace selection to first value

mootari:
In the workspace selection we sort workspaces by role and name, but default the selection to the first unsorted value. Example:
image

 Because this selection may appear arbitrary to users, we should instead preselect the first option in the list.


mbostock:
Maybe it would be better to make the API smarter about what the default workspace should be? Like, it could be the one you most recently deployed to?

mootari:
Sounds good to me! Ideally we also show a hint like "you recently deployed to" to explain the default.

I'll file a PR to have /cli/user return the additional metadata.

PR #1632, opened 2024-09-03
TypeScript

### LINKED ISSUE #79: TypeScript

mbostock:
Reboot of #129. Fixes #79.

TODO

- [x] <code>```ts</code> fenced code blocks
- [x] <code>```tsx</code> fenced code blocks
- [x] <code>.ts</code> modules
- [x] <code>.tsx</code> modules
- [x] ~~TypeScript in inline expressions~~ (not supported)
- [x] [verbatimModuleSyntax](https://www.typescriptlang.org/tsconfig/#verbatimModuleSyntax) to preserve imports 
- [x] documentation
- [x] tests

Unlike #129, I expressly avoided allowing TypeScript in inline expressions, which allows us to avoid running the esbuild transform on them, and in turn avoids the problem of esbuild converting expressions to expression statements and then losing implicit display. Not also that there is no implicit display in <code>```ts</code> blocks. I think this is an acceptable tradeoff, at least for now, as you get the primary benefits of TypeScript modules and TypeScript fenced code blocks and only if you explicitly opt-in to using TypeScript.

Also unlike #129, this only allows TypeScript to be imported with the `.js`, as we do for `.jsx`. This is the recommended practice because it means the import path doesn’t change after transpilation.

### DIFF
--- a/docs/javascript.md
+++ b/docs/javascript.md
@@ -127,6 +127,22 @@ const number = Generators.input(numberInput);
 
 Expressions cannot declare top-level reactive variables. To declare a variable, use a code block instead. You can declare a variable in a code block (without displaying it) and then display it somewhere else using an inline expression.
 
+## TypeScript <a href="https://github.com/observablehq/framework/pull/1632" class="observablehq-version-badge" data-version="prerelease" title="Added in #1632"></a>
+
+TypeScript fenced code blocks (<code>```ts</code>) allow TypeScript to be used in place of JavaScript. You can also import TypeScript modules (`.ts`). Use the `.js` file extension when importing TypeScript modules; TypeScript is transpiled to JavaScript during build.
+
+<div class="warning">
+
+Framework does not perform type checking during preview or build. If you want the additional safety of type checks, considering using [`tsc`](https://www.typescriptlang.org/docs/handbook/compiler-options.html).
+
+</div>
+
+<div class="note">
+
+TypeScript fenced code blocks do not currently support [implicit display](#implicit-display), and TypeScript is not currently allowed in [inline expressions](#inline-expressions).
+
+</div>
+
 ## Explicit display
 
 The built-in [`display` function](#display-value) displays the specified value.
--- a/docs/javascript.md
+++ b/docs/javascript.md
@@ -86,7 +86,7 @@ html`<span style=${{color: `hsl(${(now / 10) % 360} 100% 50%)`}}>Rainbow text!</
 
 ## Inline expressions
 
-Inline expressions <code>$\{…}</code> interpolate values into Markdown. They are typically used to display numbers such as metrics, or to arrange visual elements such as charts into rich HTML layouts.
+JavaScript inline expressions <code>$\{…}</code> interpolate values into Markdown. They are typically used to display numbers such as metrics, or to arrange visual elements such as charts into rich HTML layouts.
 
 For example, this paragraph simulates rolling a 20-sided dice:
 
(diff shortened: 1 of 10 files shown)


docs/javascript.md
```diff
@@ -127,6 +127,22 @@ const number = Generators.input(numberInput);
 
 Expressions cannot declare top-level reactive variables. To declare a variable, use a code block instead. You can declare a variable in a code block (without displaying it) and then display it somewhere else using an inline expression.
 
+## TypeScript <a href="https://github.com/observablehq/framework/pull/1632" class="observablehq-version-badge" data-version="prerelease" title="Added in #1632"></a>
+
+TypeScript fenced code blocks (<code>```ts</code>) allow TypeScript to be used in place of JavaScript. You can also import TypeScript modules (`.ts`). Use the `.js` file extension when importing TypeScript modules; TypeScript is transpiled to JavaScript during build.
+
+<div class="warning">
+
+Framework does not perform type checking during preview or build. If you want the additional safety of type checks, considering using [`tsc`](https://www.typescriptlang.org/docs/handbook/compiler-options.html).
```
Fil, line 136:
```suggestion
Framework strips type annotations from TypeScript code blocks, but does not perform type checking during preview or build. If you want the additional safety of type checks, considering using [`tsc`](https://www.typescriptlang.org/docs/handbook/compiler-options.html).
```

docs/javascript.md
```diff
@@ -127,6 +127,22 @@ const number = Generators.input(numberInput);
...
+
+</div>
+
+<div class="note">
+
+TypeScript fenced code blocks do not currently support [implicit display](#implicit-display), and TypeScript is not currently allowed in [inline expressions](#inline-expressions).
+
+</div>
+
```
Fil, line 145:
```suggestion
```

I don't feel this is strictly necessary, as the documentation for implicit display starts with "JavaScript". We could add "JavaScript" before inline expressions, too.

Otherwise, "currently" seems to hint that these might be future goals. I'm not sure it's necessarily the case; if it is and if we want to keep this note, we could link to an issue to see if there is demand?

Fil (approved):
This feels right and elegant. Just adding two suggestions for the documentation

docs/javascript.md
mbostock: I think it is important to keep because it’s natural for people to otherwise assume that there is parity/consistency between TypeScript and JavaScript when that is not the case. I want to highlight the ways they are different.

We could remove the word “currently”, but I was curious to see if people would ask for it. If we don’t include that word, it may imply that we reject future support for it, which might discourage people from asking for it — and then we wouldn’t get a sense if there’s demand. I think we’re open to adding in the future (at least if it’s easy).

docs/javascript.md
mbostock: I thought about saying this, but converting TypeScript to JavaScript is somewhat more involved than removing the type annotations, at least in the case of enum, right? I thought there were a few parts of the TypeScript language (perhaps regrettably) that don’t meet this criterion. And anyway that felt more like an implementation detail, whereas the point I wanted to make here is that there is no type checking, which is the primary reason you’re using TypeScript, presumably.

Fil: OK!

Fil:
The next release is going to be HUGE!

mpj:
oh my

PR #1636, opened 2024-09-03
detect non-local sql tables

### LINKED ISSUE #1635: framework tries to build a remote file listed in sql front-matter
Here’s `test.md`:
~~~~md
---
sql:
  taxi: https://blobs.duckdb.org/data/taxi_2019_04.parquet
---

```sql echo
SELECT COUNT() FROM taxi;
```
~~~~

The page works fine in preview and in build, and displays the expected result `7,433,139`.

However the logs show an erroneous tentative in build to copy the (non-existing) resource `src/https:/blobs.duckdb.org/data/taxi_2019_04.parquet` (note the single `/` after `https:`).

`copy src/https:/blobs.duckdb.org/data/taxi_2019_04.parquet → error: missing referenced file`

Another symptom is a spurious call to `registerFile` in the page's scripts:

```js
registerFile("https://blobs.duckdb.org/data/taxi_2019_04.parquet", {"name":"https://blobs.duckdb.org/data/taxi_2019_04.parquet","path":"https://blobs.duckdb.org/data/taxi_2019_04.parquet"});
```

mbostock:
Fixes #1635.

### DIFF
--- a/src/render.ts
+++ b/src/render.ts
@@ -59,7 +59,7 @@ if (location.pathname.endsWith("/")) {
 import ${preview || page.code.length ? `{${preview ? "open, " : ""}define} from ` : ""}${JSON.stringify(
     resolveImport("observablehq:client")
   )};${
-    files.size || data?.sql
+    files.size
       ? `\nimport {registerFile${data?.sql ? ", FileAttachment" : ""}} from ${JSON.stringify(
           resolveImport("observablehq:stdlib")
         )};`
--- a/src/resolvers.ts
+++ b/src/resolvers.ts
@@ -123,8 +123,11 @@ export async function getResolvers(
 
   // Add SQL sources.
   if (page.data.sql) {
-    for (const source of Object.values(page.data.sql)) {
-      files.add(String(source));
+    for (const value of Object.values(page.data.sql)) {
+      const source = String(value);
+      if (isAssetPath(source)) {
+        files.add(source);
+      }
     }
   }
 


Fil:
We can probably refine a little with this change:

```
diff --git a/src/render.ts b/src/render.ts
index 926f24e1..4b2f2efe 100644
--- a/src/render.ts
+++ b/src/render.ts
@@ -59,7 +59,7 @@ if (location.pathname.endsWith("/")) {
 import ${preview || page.code.length ? `{${preview ? "open, " : ""}define} from ` : ""}${JSON.stringify(
     resolveImport("observablehq:client")
   )};${
-    files.size || data?.sql
+    files.size
       ? `\nimport {registerFile${data?.sql ? ", FileAttachment" : ""}} from ${JSON.stringify(
           resolveImport("observablehq:stdlib")
         )};`

```

and skip an unnecessary import of stdlib. It passes all the tests but at this point I'm wary of introducing a last-minute bug.

mbostock:
Okay, added. I tested, but I can’t load the Parquet file from blobs.duckdb.org anymore because of a CORS error, which is odd because I thought it worked before. The generated code looks fine though. 🤷

Fil:
It worked before for me too, and now I also got the CORS error. But substituting a different URL confirms this works.

(We still have `<link rel="modulepreload" href="./_observablehq/stdlib.c0fee887.js">`, so maybe it isn't much better.)

Fil (approved):

mbostock:
I don’t think it’s possible to avoid loading the Observable standard library entirely; see `src/client/main.js`. Still fine to avoid the unnecessary import though and save a few bytes. 😁

PR #1637, opened 2024-09-03
embedded api

### LINKED ISSUE #1084: A JavaScript API for embedded use cases
We currently offer only a command-line interface for previewing, building, and deploying projects. We could offer a JavaScript API that provides more flexibility to transpile Markdown and JavaScript.

### LINKED ISSUE #1583: Add CORS headers in preview mode
Hello,

We use Observable with API calls to a few different servers.
The users connect with their own ID to these servers to get the data if they are authorized.
But I can't use Observable Framework preview server while developing because of CORS.
When I create react apps with create-react-app, their preview server (npm start) adds CORS headers automatically and I can test the app:
```
access-control-allow-headers: *
access-control-allow-methods: *
access-control-allow-origin: *
```

So, I'm suggesting to add these CORS headers automatically to the preview server as well or, at least, add a config that enables this feature.

Thank you.

mbostock:
TODO:

- [x] Generate embed script (_e.g._, `/chart.js`) during preview
- [x] Generate embed script (_e.g._, `/chart.js`) during build via ~~**embedPaths**~~ **dynamicPaths**
- [x] Fix build so that module rendering applies content-hashed aliases (via resolvers)
- [x] Fix build so that assets & imports of embedded modules are included
- [x] Fix build so that the tree output is relevant for embedded modules
- [x] Fix messy types in build
- [x] Embed script script takes precedence over page (_e.g._, `/chart.js.md`)
- [x] Ignore pre-empted pages (`/chart.js.md` given `/chart.js`) in the sidebar & pager
- [x] Compute transitive static imports for preloading
- [x] Inline file registration to avoid ordering problem
- [x] Allow CORS during preview
- [x] Restrict CORS preview to specific origins
- [x] Documentation
- [x] Tests

Fixes #1084. 
Fixes #1583.


mbostock:
Would be simpler to just invert the precedence and have `/chart.js.md` take precedence over `/chart.js`? I don’t think it matters which one we choose; we just have to pick one, so we should pick whatever’s easier for us.

mbostock:
Boo, one of our tests has a page `sub/page1..10.md` which already violates the double extension rule, suggesting that we should allow pages to have double extensions but fix the precedence issue… I need to think more.

src/preview.ts
```diff
@@ -116,6 +121,9 @@ export class PreviewServer {
     const {root, loaders} = config;
     if (this._verbose) console.log(faint(req.method!), req.url);
     const url = new URL(req.url!, "http://localhost");
+    const {origin} = req.headers;
+    if (this._origins.includes("*")) res.setHeader("Access-Control-Allow-Origin", "*");
+    else if (origin && this._origins.includes(origin)) res.setHeader("Access-Control-Allow-Origin", origin);
```
Fil, line 126: 👍

docs/embeds.md
```diff
@@ -0,0 +1,120 @@
...
+
+```js run=false
+export default {
+  dynamicPaths: [
+    "/chart.js"
+  ]
+};
+```
+
+Or for [parameterized routes](./params), name the component `product-[id]/chart.js`, then load a list of product identifiers from a database with a SQL query:
```
Fil, line 15: This is a bit too allusive. It might be worth developing an example of how to use `observable.params.id` within `chart.js` so that it corresponds to the requested product and displays data from the corresponding (parametrized) data loader. The example can live as a stand-alone example app, and we'd just add a link here.

Fil (approved):
This is fantastic! and I couldn't find how to break it 👍 
Adding a few suggestions for the documentation.

docs/embeds.md
mbostock: Yeah, I didn’t want to duplicate the documentation of parameterized routes here, but I also wanted to allude to it, since it feels like a natural question (how would I show multiple views?) with perhaps a surprising answer (the views bake-in their data rather than expecting it to be passed in as a prop from the host app).

docs/embeds.md
```diff
@@ -0,0 +1,120 @@
...
+
+```js run=false
+import postgres from "postgres";
+
+const sql = postgres(); // Note: uses psql environment variables
```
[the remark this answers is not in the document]
mbostock: This comment exists in multiple places. (I’ve been copying it every time I use Postgres.js.) I’m not sure we need it. The link from the text is probably enough, but I did think it was useful to mention that it depends on environment variables.

docs/embeds.md
```diff
@@ -0,0 +1,120 @@
...
+
+<div class="tip">
+
+Since both dynamic import and the imported component are async, the code above is careful to clean up the effect and avoid race conditions.
+
+</div>
+
+<div class="tip">
+
+You can alternatively embed Framework pages using [iframe embeds](https://observablehq.observablehq.cloud/framework-example-responsive-iframe/).
```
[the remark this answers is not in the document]
mbostock: Yes. That should just be the `resize` helper, as normal.

docs/embeds.md
```diff
@@ -0,0 +1,120 @@
...
+    for await (const {id} of sql`SELECT id FROM products`.cursor()) {
+      yield `/product-${id}/chart.js`;
+    }
+  }
+};
+```
+
+Embedded modules are vanilla JavaScript, and will behave identically when embedded in an external application as on a Framework page. As always, you can load data from a [data loader](./data-loaders) using [`FileAttachment`](./files), and you can [import](./imports) [self-hosted](./imports#self-hosting-of-npm-imports) local modules and libraries from npm; file and import resolutions are baked into the generated code at build time so that imported modules “just work”.
+
+Embedded modules are often written as functions that return DOM elements. These functions can take options (or “props”), and  typically load their own data via `FileAttachment`. For example, below is a simple `chart.js` module that exports a `Chart` function that renders a scatterplot.
```
Fil:
```suggestion
Embedded modules are often written as functions that return DOM elements. These functions can take options (or “props”), and typically load their own data via `FileAttachment`. For example, below is a simple `chart.js` module that exports a `Chart` function that renders a scatterplot.
```

PR #1639, opened 2024-09-04
embedded

Fil:
This is just to share my (very slow) progress on #1637. I haven't found how to unlock the build process 🤯.

Only a small goal is partially achieved (the side bar not showing the shadowed page `chart.js.md`). I hope the test is useful, I run it like so:

```js
yarn observable build --root test/input/build/embed
```


src/config.ts
```diff
@@ -185,6 +186,8 @@ function readPages(root: string, md: MarkdownIt): Page[] {
   const hash = createHash("sha256");
   for (const file of visitFiles(root, (name) => !isParameterized(name))) {
     if (extname(file) !== ".md" || file === "index.md" || file === "404.md") continue;
+    const path = file.slice(0, -".md".length);
+    if (path.endsWith(".js") && findModule(root, path)) continue;
```
mbostock, line 190: Using `findModule` here is probably too lax; it means that `/chart.js.md` would be preceded by `/[param].ts`, for example, which feels wrong. I guess it makes sense to have `/chart.js.md` preceded by `/chart.ts`, but I’m not sure about parameterized routes.

mbostock: Maybe a simpler approach would be to just disallow `.js.md` entirely; it feels like a very rare case where you’d insist on a page having the `.js` extension as in `/name.js`, and you could always use `/name-js` instead.

mbostock: We could also disallow multiple extensions in pages more generally, like there’s no reason to have `/name.html.md`, either.

mbostock: Using `findModule` here is necessary to match the behavior of the Preview server, so this is correct.

I think we might also want to push this test down into `loaders.findPage` so that it’s not possible to accidentally resolve a page that isn’t servable because it is preceded by a JavaScript module.

PR #1641, opened 2024-09-04
Add version badge to docs referencing file.size

### LINKED ISSUE #1640: Add version badge for reference to file.size
From the docs here: https://observablehq.com/framework/files#static-analysis it would appear users can check for missing files using `file.size`.

From [this commit](https://github.com/observablehq/framework/commit/06ed740f7803df18f46cfd6b23d6601cb9409e1b) it looks to me that file.size is always undefined for versions < `1.11.0`. As such, it may be appropriate to add a version badge to the documentation here so as to prevent developers mistakenly using the `file.size` attribute when it will always be undefined for older versions.

gvanhorne:
Closes #1640

### DIFF
--- a/docs/files.md
+++ b/docs/files.md
@@ -52,7 +52,7 @@ const frames = [
 
 None of the files in `frames` above are loaded until a [content method](#supported-formats) is invoked, for example by saying `frames[0].image()`.
 
-For missing files, `file.size` <a href="https://github.com/observablehq/framework/releases/tag/v1.11.0" class="observablehq-version-badge" data-version="^1.11.0" title="Added in 1.11.0"></a> and `file.lastModified` are undefined. The `file.mimeType` is determined by checking the file extension against the [`mime-db` media type database](https://github.com/jshttp/mime-db); it defaults to `application/octet-stream`.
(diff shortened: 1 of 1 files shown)


docs/files.md
```diff
@@ -52,7 +52,7 @@ const frames = [
 
 None of the files in `frames` above are loaded until a [content method](#supported-formats) is invoked, for example by saying `frames[0].image()`.
 
-For missing files, `file.size` and `file.lastModified` are undefined. The `file.mimeType` is determined by checking the file extension against the [`mime-db` media type database](https://github.com/jshttp/mime-db); it defaults to `application/octet-stream`.
+For missing files, `file.size` <a href="https://github.com/observablehq/framework/releases/tag/v1.11.0" class="observablehq-version-badge" data-version="^1.11.0" title="Added in 1.11.0"></a> and `file.lastModified` are undefined. The `file.mimeType` is determined by checking the file extension against the [`mime-db` media type database](https://github.com/jshttp/mime-db); it defaults to `application/octet-stream`.
```
mbostock, line 55: Do you want to add `<a href="https://github.com/observablehq/framework/releases/tag/v1.4.0" class="observablehq-version-badge" data-version="^1.4.0" title="Added in 1.4.0"></a>` for `file.lastModified` too?

gvanhorne: Sure that makes sense to me! 👍

mbostock (commented):
I think the compromise here is that we should explicitly document the names of these fields the first time they are introduced, along with the version at which they were added. Then we won’t repeat the version badge when they are referenced subsequently. The problem with these fields is that we only introduce them as _e.g._ “size in bytes” and “modification time” rather than `file.size` and `file.lastModified`. I’ll make some edits and approve.

mbostock (approved):
Okay, hopefully this makes it more obvious when `file.size` was added. I’ll merge this but let us know if you’d like further changes. Thank you @gvanhorne!

PR #1646, opened 2024-09-08
Add settings URL to deploy outro

tophtucker:
9FCFBC53-B505-429D-AFBF-C96FE1CDA8A6-676-000A6670FA1700E0

### DIFF
--- a/src/deploy.ts
+++ b/src/deploy.ts
@@ -92,13 +92,7 @@ export async function deploy(deployOptions: DeployOptions, effects = defaultEffe
 
   const deployer = await new Deployer(deployOptions, effects);
   const deployInfo = await deployer.deploy();
-
-  // Get URL for app settings page
-  const deployConfig = await deployer.getUpdatedDeployConfig();
-  const {deployTarget} = await deployer.getDeployTarget(deployConfig);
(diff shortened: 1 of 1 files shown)


Fil:
I find this both useful and disturbing (like every time you have two CTAs).

Maybe we could clarify what these settings are about, and put the link a bit above rather than below the app URL.

```
To review cloud settings: <https://observablehq.com/projects/@observablehq/observable-framework/settings>

[bold]The deployed app is now visible at <https://observablehq-observablecloud.test:5000/observable-framework/>[/bold]
```

mbostock:
Agree, not sure we want a dual CTA here. Should we have a command you can run that automatically opens either the live app, or the app’s setting page, in the browser?

tophtucker:
Separate command sounds good! Hm… some relevant tasks (reasons you’d want web settings from CLI) might be:

1. Opening the page the first time you deploy and the app is created on Observable so you can continue initial configuration
2. Kicking off a cloud build and viewing logs
3. Link the app to the repo you’re in (if possible??)

I think we can definitely improve the connection somehow but those all feel different enough that I’ll close this PR. 🙏

PR #1655, opened 2024-09-12
fix responsive chart dimensions on the EIA example

Fil:
This was a transient warning, due to the fact that the charts run independently. The height of the interchange chart was negative when the map had not yet been inserted in the DOM. By setting a min-height in css, we avoid the warning and more importantly only compute the charts once.

warning

Thanks @mootari for reporting.

In addition, I fixed the display for small screens.

redeployed at https://observablehq.observablehq.cloud/framework-example-eia/


Fil:
We could do more to make sure the page looks good on smaller screens too. For example in this view, we should remove the padding to give a bit more space to the map, make sure the legend doesn't overlap, etc.
Capture d’écran 2024-09-12 à 10 06 12.

mbostock (approved):

PR #1662, opened 2024-09-13
getModuleHash needs to consider data loaders

### LINKED ISSUE #1661: Reactive updates don’t work for modules that load data loaders
Consider `test.md`:

````md
```js
import {test} from "./test.js";

display(await test());
```
````

And `test.js`:

```js
import {FileAttachment} from "npm:@observablehq/stdlib";

export async function test() {
  return FileAttachment("./test.txt").text();
}
```

And `test.txt.js`:

```js
process.stdout.write("hello");
```

If you edit `test.txt.js`, it should change the hash of `test.js`, and cause the page `/test` to update. But it doesn’t because `getModuleHash` doesn’t allow for a `FileAttachment` being backed by a data loader.

mbostock:
Fixes #1661.


mbostock:
TODO Test that if a data loader generates a different output, that the hash of a module that loads that file changes; and likewise if another module imports that module, and its hash changes.

Fil (approved):

PR #1667, opened 2024-09-17
adopt curl -f to make the data loader fail if curl fails to download

Fil:
`curl` is a rare unix utility that does not return an error code > 0 when it fails, unless we use [`-f`](https://curl.se/docs/manpage.html#-f) (`--fail`); seems like a best practice we should promote.


allisonhorst (approved):
Looks good to me. We use curl for the EIA example (in [eia-system-points.json.sh](https://github.com/observablehq/framework/blob/main/examples/eia/src/data/eia-system-points.json.sh)). Even though it doesn't actually update with live data, we may want to update with -f there for consistency.

Fil:
Note to self: I'll have to re-deploy loader-census and loader-duckdb, since they show the loader code in the pages.

PR #1669, opened 2024-09-17
flatten (WIP)

### LINKED ISSUE #1668: Flatten the rendered page structure?
The page structure currently is the following:

* html
  * head
    * observable scripts
    * custom head
  * body
    * sidebar (optional)
    * toc (optional)
    * center
      * header (custom)
      * main (page content)
      * footer (custom)

there are lots of optional and custom content, but the fact that some of it is nested in the center div makes it a bit difficult to reorganize with css (flexbox or grid).

It would feel nice to flatten the body:

  * body
    * sidebar (optional)
    * toc (optional)
    * header (custom)
    * main (page content)
    * footer (custom)

I experimented a bit, and it doesn't seem to require many changes in grid.css — basically we just have to tack an .observablehq-center _class_ on the header, main and footer elements, and change a few of the margins.

An alternative (or complement) would be to reboot #253 and allow "full page templates".

Fil:
- [ ] fix margins

closes #1668

### DIFF
--- a/src/style/layout.css
+++ b/src/style/layout.css
@@ -7,7 +7,7 @@
 #observablehq-main,
 #observablehq-header,
 #observablehq-footer {
-  margin: 1rem auto;
+  _margin: 1rem auto;
   max-width: 1152px;
 }
 
@@ -76,7 +76,7 @@
   content: "Next page";
 }
 
-#observablehq-center {
+.observablehq-center {
   margin: 1rem 2rem;
 }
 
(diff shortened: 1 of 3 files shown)


src/style/layout.css
```diff
@@ -7,7 +7,7 @@
 #observablehq-main,
 #observablehq-header,
 #observablehq-footer {
-  margin: 1rem auto;
+  _margin: 1rem auto;
```
Fil, line 10: all the margins needs fixing

PR #1685, opened 2024-09-26
fix scroll margin for anchor links; fix conditional header layout

### LINKED ISSUE #1686: Space for a header is reserved, even if there isn’t a header
Regression in #1682.

Screenshot 2024-09-26 at 8 51 09 AM

Fil:
(e.g. https://observablehq.com/framework/files#static-analysis)

Now also fixes #1686.

### DIFF
--- a/docs/style.css
+++ b/docs/style.css
@@ -10,15 +10,6 @@
   scroll-padding-top: 4rem;
 }
 
-h1,
-h2,
-h3,
-h4,
-h5,
-h6 {
-  margin-top: revert;
-}
-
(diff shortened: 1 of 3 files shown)


mbostock:
Good find. We can do a simpler fix with scroll-padding-top instead of scroll-margin-top.

mbostock (approved):

mbostock:
I guess this fix should be applied in the default styles rather than our custom stylesheet? But then we need to know whether the header is shown. Hmm. 🤔

PR #1688, opened 2024-09-26
fix hashes for modules importing builtins

mbostock:
`getLocalModuleHash` was computing the wrong result for modules that import built-ins such as `observablehq:stdlib` and `npm:@observablehq/inputs`; it needs to incorporate the Framework version number so that when a new version of Framework is released and the built-in modules may have changed, the hash of the importing local module also changes. (We already do something similar for `npm:` imports.)

We were also forgetting to handle `jsr:` imports here.

Related #1555 #1557.


src/resolvers.ts
```diff
@@ -491,6 +489,12 @@ export function getModuleResolver(
   };
 }
 
+export function resolveBuiltin(specifier: string): string {
+  if (builtins.has(specifier)) return builtins.get(specifier)!;
+  if (!specifier.startsWith("observablehq:")) throw new Error(`not built-in: ${specifier}`);
+  return `/_observablehq/${specifier.slice("observablehq:".length)}${extname(specifier) ? "" : ".js"}`;
```
Fil, line 495: took me a moment to figure out what case that was!

one example: `import {mutable} from "observablehq:stdlib/mutable.js";`

Fil (approved):

src/resolvers.ts
mbostock: Yep! The primary one is `observablehq:stdlib/inputs.css`.

PR #1689, opened 2024-09-27
home config option

mbostock:
This adds a **home** config option which allows overriding how the link to the home page appears in the sidebar. In particular, this allows our docs site to show the Observable logo in the top-left rather than in the fixed header. When the sidebar is shown (either because the page is wide enough to show it naturally, or because it was displayed manually), we hide the Observable logo in the header to avoid showing it twice.

Screenshot 2024-09-26 at 5 52 25 PM

When the sidebar is hidden, the Observable logo reappears in the header as today.

Screenshot 2024-09-26 at 5 52 51 PM

Probably need some tests and polish so marking as draft for now.


CobusT:
Nice. Is the plan to use `logo` when rendering `home`/`title` also? I didn't see where that was being used in this PR yet, instead of putting the logo's svg in the `home` definition?

mbostock:
I originally implemented a **logo** option but I decided to replace it with a **home** option instead (incorporating both **title** and what would have been **logo**). Did I accidentally leave some of that code in this PR? I’ll remove it when I polish and add docs etc.

CobusT (approved):

PR #1692, opened 2024-09-27
embedding docs

mbostock:


docs/embeds.md
```diff
@@ -103,28 +103,39 @@ export function EmbedChart() {
...
 
-```js echo
-import {Chart} from "./chart.js";
+```yaml
+---
+sidebar: false
+header: false
+footer: false
+pager: false
+---
```
Fil, line 132: I wonder if we should add a shorthand for this, so people could say “this page has no chrome” without having to list everything that should not be there?

docs/embeds.md
```diff
@@ -1,37 +1,41 @@
...
 
-Embedded modules are often written as component functions that return DOM elements. These functions can take options (or “props”), and typically load their own data. For example, below is a simple `chart.js` module that exports a `Chart` function that renders a scatterplot of global surface temperature data.
+- [exported modules](#exported-modules) for seamless integration and performance,
+- [exported files](#exported-files) for hotlinking images, data, and other assets, or
+- [iframe embeds](#iframe-embeds) for compatibility.
+
+## Exported modules
+
+Framework allows [JavaScript modules](./imports#local-imports) to be exported for use in another application. Exported modules are vanilla JavaScript and function identically in an external web application as on a Framework page. As with local modules, exported modules can load data from a [static file](./files) or a [data loader](./data-loaders), [import](./imports) other local modules, and import libraries from [npm](./imports#npm-imports) or [JSR](./imports#jsr-imports).
```
Fil, line 15:
```suggestion
Framework allows [JavaScript modules](./imports#local-imports) to be exported for use in another application. Exported modules are vanilla JavaScript and work identically in an external web application as on a Framework page. As with local modules, exported modules can load data from a [static file](./files) or a [data loader](./data-loaders), [import](./imports) other local modules, and import libraries from [npm](./imports#npm-imports) or [JSR](./imports#jsr-imports).
```

Editing to avoid "function" just after "JavaScript". Other synonyms include work, operate, run.

Fil (approved):
Much clearer, thanks!

docs/embeds.md
mbostock: Ya, `chrome: false` sounds nice. 👍

PR #1695, opened 2024-09-29
Update types

mootari:
A typo in an outdated markdown-it type made me notice that several of our types don't match their library versions.

This PR
- updates DT types to the latest available versions (and downgrades in the case of `he`),
- removes `@types/prompts`,
- adds a mocha test to verify that the version constraints match.

Some notes:
- We may want to downgrade @types/node to match the engine version.
- I had to skip checks for most of the packages, but using `this.skip()` allowed me to at least keep them in view.
- I'm not really sure that mocha is the best place to handle this.
- I considered querying the latest versions for skipped packages but it seemed to offer little benefit given the overhead.
- Ideally `yarn.lock` would be the source of truth but that seems more complicated than what I'd want to deal with.


mbostock (changes requested):
I think the test is overkill here, especially given the number of exceptions we already need to allow. How about dropping the test but keeping the rest of the changes?

mootari:
> How about dropping the test but keeping the rest of the changes?

Sure, it didn't feel quite right anyway. 🙂

I also updated the Node types to match the minimum engine version.

mbostock (approved):

mbostock:
Did this PR introduce this warning? We should fix…

```
=============

WARNING: You are currently running a version of TypeScript which is not officially supported by @typescript-eslint/typescript-estree.

You may find that it works just fine, or you may not.

SUPPORTED TYPESCRIPT VERSIONS: >=4.7.4 <5.6.0

YOUR TYPESCRIPT VERSION: 5.6.2

Please only submit bug reports when using the officially supported version.

=============
```

mootari:
@mbostock No, by #957 as far as I can tell. We may need to update eslint I think?

mbostock:
Thanks for tracking that down @mootari!

PR #1696, opened 2024-09-29
import.meta.resolve for files

### LINKED ISSUE #1008: import.meta.resolve for files
You can use `import.meta.resolve` for JavaScript modules, but it’s not currently supported for files (as it assumes that the referenced file is in `_import` rather than `_file`). It would be nice as an alternative to `FileAttachment(name).url()` because it is synchronous, e.g.,

```js
fetch(import.meta.resolve("./data/nyc-taxi.parquet"))
```

Related #954. Perhaps `_import` and `_file` could be combined? But the challenge there is we transpile JavaScript, and hence we want to keep `_import` and `_file` separate. 🤔

mbostock:
Fixes #1008. When `import.meta.resolve` is passed a relative path to a non-JavaScript file (anything other than `.js`, `.mjs`, or `.cjs` — and the latter two we don’t expect in practice), it now uses `resolveFile` instead of `resolveImport` to resolve.


Fil (commented):
in preview, this does not seem to apply to `"volcano.json"` without a leading `./`:

```
FileAttachment("volcano.json").href // "http://127.0.0.1:3000/_file/volcano.json?sha=f5223ec98032634f121adef600c947fab5ab9389d689a0bb1e21d6a877d21178"

import.meta.resolve("volcano.json") // "volcano.json" 🌶

import.meta.resolve("./volcano.json") // "http://127.0.0.1:3000/_file/volcano.json?sha=f5223ec98032634f121adef600c947fab5ab9389d689a0bb1e21d6a877d21178"
```

and in build, it crashes on `import.meta.resolve("./volcano.json")`:

```
copy docs/volcano.json →
Unexpected error: Unexpected token (1:8)
The full error follows

file:///Users/fil/Source/framework/node_modules/acorn/dist/acorn.mjs:3580
  var err = new SyntaxError(message);
            ^
```

Also, can we mention it in `docs/files.md`?

mbostock:
The error is expected. You need to using a leading dot-slash to import a relative path (`./`, `/`, or `../`). Otherwise `volcano.json` is interpreted as a bare module specifier (like `d3` or `npm:d3`). I can work on a better error message.

Fil:
I think I'm seeing a different bug, as I get a crash (`copy docs/volcano.json → \nUnexpected error: Unexpected token (1:8)`) if I add 
~~~~
```js echo
import.meta.resolve("./volcano.json")
```
~~~~
to `docs/files.md`, and run `yarn docs:build`.

However it doesn't seem completely related to this PR, or it's my testing that is wrong, because I see the same issue on main with this minimal repro: https://github.com/observablehq/broken-build/blob/main/src/index.md (I'm a bit puzzled.)

Fil (approved):
I guess it's missing documentation, but LGTM

Fil:
oops auto-merge

PR #1697, opened 2024-09-30
use registry.npmjs.org

mbostock:
This switches from data.jsdelivr.com (which is currently down) to registry.npmjs.org (which is the ground truth) to resolve npm package versions. It also seems to be quite a bit faster.


src/npm.ts
```diff
@@ -221,15 +221,18 @@ async function resolveNpmVersion(root: string, {name, range}: NpmSpecifier): Pro
...
+  if (range === undefined) range = "latest";
+  const disttag = validRange(range) ? null : range;
+  const href = `https://registry.npmjs.org/${name}${disttag ? `/${disttag}` : ""}`;
   let promise = npmVersionRequests.get(href);
   if (promise) return promise; // coalesce concurrent requests
   promise = (async function () {
     const input = formatNpmSpecifier({name, range});
     process.stdout.write(`npm:${input} ${faint("→")} `);
-    const response = await fetch(href);
+    const response = await fetch(href, {...(!disttag && {headers: {Accept: "application/vnd.npm.install-v1+json"}})});
```
mbostock, line 232: This is the “abbreviated” response described here: 

https://github.com/npm/registry/blob/main/docs/responses/package-metadata.md

Fil (approved):

PR #1698, opened 2024-09-30
detect broken links

### LINKED ISSUE #363: Detect broken links during build
It would be nice when we build to have a report on  internal links, including broken links (with an option to make that an error), broken anchors (same), and pages unreachable from the index.

### LINKED ISSUE #1683: case sensitive page URLs
Like https://github.com/observablehq/framework/issues/1367 but for pages. Suppose I have a path/to/page.md, and link to `[link](/PaTh/To/PaGe)`. It works in preview if I'm using a case-insensitive dev machine, but will fail in prod if the server is case-sensitive. (The opposite configuration might create a bug too, but I believe it's less common.)

(Reported by https://github.com/mcmcclur/MarkOnFramework/discussions/5.)

Not sure how to address this one. The solution cannot be the same as #1374: the difference being that, for files, both links and paths are generated from the markdown contents (so we can spot a discrepancy if two references to the same file point to different casings that end up at the same place on disk). For pages, the paths are generated from the src folder, and the links from the markdown contents.

🤔

Fil:
Currently only warning, but we could make it break the build like we do in Plot's documentation. Should this be a build option?

closes #363
closes #1683

(Test by running `yarn docs:build`. Merge #1699 to see the happy state.)

**happy**
happy

**unhappy**
unhappy


src/build.ts
```diff
@@ -125,6 +129,31 @@ export async function build(
   if (pageCount) effects.logger.log(`${faint("built")} ${pageCount} ${faint(`page${pageCount === 1 ? "" : "s"} in`)} ${root}`); // prettier-ignore
   if (assetCount) effects.logger.log(`${faint("built")} ${assetCount} ${faint(`asset${assetCount === 1 ? "" : "s"} in`)} ${root}`); // prettier-ignore
 
+  // Check links.
+  const REF = "proto://host/_observablehq";
```
Fil, line 133: This is needed to detect links that try to go higher up than the root. I'm using `_observablehq/` as the pathname since it's a reserved word — and we should never encounter a link to root/_observablehq/page.md.

Fil:
The page sizes table we show at the end and the link checker both look similar in that they are "reporters" rather than strictly necessary for the build.

I put the link checker at this location because it's fast and if it must break the build it's better to do it sooner.  But I'd prefer the happy report to be shown at the end.

mbostock (approved):
Hooray! Great to have this.

I made [some changes](https://github.com/observablehq/framework/compare/5e58c6df3c190eb6b4170ff2e3f74dcd4c90fc3e...add034334b2b0c4c9712269b1e4ca9daa9a4cd79). To summarize:

- I encapsulated the code a bit more, reducing the number of local variables in the already quite large and stateful `build` function. It would be nice to have separate unit tests for `validateLinks`, perhaps, but it’s pretty simple so not urgent.
- I moved the broken link warning to the end where it’s more likely to be seen. If we want to make this a hard error, then we can move it earlier.
- I renamed `links` to `localLinks` to indicate that we’re only extracting (and validating) local links, not all links.
- I added encoding as appropriate (`${path}#${encodeURIComponent(id)}`).
- I skipped validation of `<a href download>` since these links are rewritten to point to files.
- I disallowed (non-download) links to files, since these links are not rewritten and would be broken.
- I would like to allow links to exported files, but I was lazy and didn’t implement this yet.
- I allowed links to exported modules (which is unlikely but should work).
- I added normalization for trailing `.html` (in addition to the existing `/index`).
- I used `parseRelativeUrl` instead of hacking URL with `proto:…`.
- I fixed a longstanding quirk in `resolvePath` (and added tests).
- I added unit tests for the new `findAssets` logic.

Issue #1469, opened 2024-06-17
Serialize asynchronous evaluation to improve performance


Fil:
`sql` queries —using the function in js code blocks, or within sql code blocks— can depend on variables that change often, such as the value of an interactive range slider.

The performance of these queries can suffer from the sheer quantity of queries sent during interaction, as they need to be executed in turn. Furthermore, it is a huge waste of compute, since all the queries but the last correspond to an invalidated state of the cell, and are discarded after execution.

This problem generalizes to all asynchronous functions.

Registering asynchronous functions per cell—so that each cell only runs one query at a time, and skips the queries launched by cells that have been invalidated—could lead to a dramatic performance boost.

Fil:
I've applied it to https://observablehq.observablehq.cloud/pangea/varia/lap-jv which was otherwise crashing when you moved the slider "too much".

PR #1728, opened 2024-10-08
serialize DuckDBClient.sql

mbostock:
Ref #1469. This uses a WeakMap keyed by the string literal `strings`, effectively creating a queue per SQL cell and per `sql` tagged template literal. I’d prefer to implement a more general solution that also helps out with _e.g._ concurrent `fetch` (maybe in the Observable Runtime?), but this is an easy fix for probably the most common instance of this problem. The performance of the `mag` slider on the [SQL page](https://observablehq.com/framework/sql) is dramatically improved under fast interaction.

### DIFF
--- a/src/client/stdlib/duckdb.js
+++ b/src/client/stdlib/duckdb.js
@@ -58,8 +58,24 @@ export function registerTable(name, source) {
   }
 }
 
-export async function sql(strings, ...args) {
-  return (await getDefaultClient()).query(strings.join("?"), args);
+function queue(f) {
+  let current;
+  let version = 0;
+  return async function () {
+    const v = ++version;
+    try {
+      await current;
+    } catch {
+      // ignore errors
+    }
+    if (version !== v) throw new Error("invalidated");
+    return (current = f.apply(this, arguments));
+  };
+}
+
+export async function sql() {
+  const db = await getDefaultClient();
+  return db.sql.apply(db, arguments);
 }
 
 export async function getDefaultClient() {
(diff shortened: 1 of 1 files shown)


Fil:
The crucial thing here is that even if another cell uses the same query, its *strings* will not be `===` to the first query; so no queuing happens between the two queries. 👏

Fil (approved):

mbostock:
Unfortunately I think this approach has a fundamental flaw as exhibited in this example code:

```js
Promise.all([1, 2, 3].map((i) => sql`SELECT ${i}`))
```

This code will throw `Error: invalidated` because the *i* = 1 and *i* = 2 cases are “invalidated” by the subsequent *i* = 3. In other words, the approach in this PR works only if the `sql` template literal is invoked only once per code block, and any case in which the `sql` template literal exists within a loop will cause problems.

We could still adapt this fix for SQL cells specifically (by changing the generated code to `sql.queue` or something), but perhaps I should just look into making the more general runtime-level fix.

Fil:
Ah… bummer.

If you write `Promise.all([1, 2, 3].map((i) => sql(["SELECT", ""], i)))` instead it works, because it's a new array each time. MDN [explains](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) the difference: “For any particular tagged template literal expression, the tag function will always be called with the exact same literal array, no matter how many times the literal is evaluated.”

Could we salvage this approach if we somehow passed the variable's `v._version` to know if we are running in the same call or in a subsequent one?

PR #1729, opened 2024-10-08
fix links

Fil:
after abef61b

### DIFF
--- a/docs/imports.md
+++ b/docs/imports.md
@@ -230,7 +230,7 @@ Click on any of the imported symbols below to learn more.
 <pre><code class="language-js">import * as <a href="./lib/duckdb">duckdb</a> from "npm:@duckdb/duckdb-wasm";</code></pre>
 <pre><code class="language-js">import {<a href="./lib/duckdb">DuckDBClient</a>} from "npm:@observablehq/duckdb";</code></pre>
 <pre><code class="language-js">import {<a href="./sql">sql</a>} from "npm:@observablehq/duckdb";</code></pre>
-<pre><code class="language-js">import * as <a href="./inputs">Inputs</a> from "npm:@observablehq/inputs";</code></pre>
+<pre><code class="language-js">import * as <a href="./inputs/">Inputs</a> from "npm:@observablehq/inputs";</code></pre>
 <pre><code class="language-js">import <a href="./lib/mapbox-gl">mapboxgl</a> from "npm:mapbox-gl";</code></pre>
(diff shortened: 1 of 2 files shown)


mbostock (commented):
Can you add a trailing slash to /inputs/ please.

docs/imports.md
```diff
@@ -230,7 +230,7 @@ Click on any of the imported symbols below to learn more.
 <pre><code class="language-js">import * as <a href="./lib/duckdb">duckdb</a> from "npm:@duckdb/duckdb-wasm";</code></pre>
 <pre><code class="language-js">import {<a href="./lib/duckdb">DuckDBClient</a>} from "npm:@observablehq/duckdb";</code></pre>
 <pre><code class="language-js">import {<a href="./sql">sql</a>} from "npm:@observablehq/duckdb";</code></pre>
-<pre><code class="language-js">import * as <a href="./lib/inputs">Inputs</a> from "npm:@observablehq/inputs";</code></pre>
+<pre><code class="language-js">import * as <a href="./inputs">Inputs</a> from "npm:@observablehq/inputs";</code></pre>
```
Fil, line 233:
```suggestion
<pre><code class="language-js">import * as <a href="./inputs/">Inputs</a> from "npm:@observablehq/inputs";</code></pre>
```

docs/reactivity.md
```diff
@@ -267,7 +267,7 @@ My favorite baseball team is the ${team}!
 My favorite baseball team is the ${team}!
 ```
 
-The above example uses `Inputs.radio`, which is provided by [Observable Inputs](./lib/inputs). You can also implement custom inputs using arbitrary HTML. For example, here is a [range input](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/range) that lets you choose an integer between 1 and 15 (inclusive):
+The above example uses `Inputs.radio`, which is provided by [Observable Inputs](./inputs). You can also implement custom inputs using arbitrary HTML. For example, here is a [range input](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/range) that lets you choose an integer between 1 and 15 (inclusive):
```
Fil, line 270:
```suggestion
The above example uses `Inputs.radio`, which is provided by [Observable Inputs](./inputs/). You can also implement custom inputs using arbitrary HTML. For example, here is a [range input](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/range) that lets you choose an integer between 1 and 15 (inclusive):
```

Fil (commented):
add a trailing slash

mbostock (approved):
We should have broken links break the build now.

PR #1731, opened 2024-10-08
annotate local paths

Fil:
paths to local files in transpiled JavaScript modules are postfixed with a comment

For example, when building the documentation, we now have in docs/.observablehq/dist/_import/chart.95e875cf.js

```js
path: "../_file/lib/gistemp.08e51068.csv"/* observablehq-file */,
```

in lieu of

```js
path: "../_file/lib/gistemp.08e51068.csv",
```

The annotate.js scripts lists a few cases that can be annotated.


mythmon:
I think we only need an annotation before or after the URL, not both. Both would be fine I suppose, if you prefer.

I think it is ok to not annotate HTML files, since embedded analytics is primarily targeted at JS modules, data files, and images.

src/javascript/transpile.ts
```diff
@@ -259,3 +256,12 @@ export function rewriteParams(output: Sourcemap, body: Node, params: Params, inp
     output.replaceLeft(node.start, node.end, JSON.stringify(params[name]));
   }
 }
+
+/**
+ * Annotate a path to a local import or file so it can be reworked server-side.
+ */
+function annotatePath(uri: string): string {
+  return !uri.startsWith(".") || /(?:[.][/]|(?:[.][.][/])+)(_jsr|_node|_npm|_observablehq)[/]/.test(uri)
```
mbostock, line 264: This should be `isAssetPath` or whatever that’s `['./', '../', '/']`. Don’t want to match against `.foo`.

mythmon:
Sorry for the delay. I plan to review this tomorrow.

src/javascript/transpile.ts
```diff
@@ -111,17 +111,15 @@ export async function transpileModule(
...
-              lastModified: info.mtimeMs,
-              size: info.size
-            }
-          : p
-      )}, import.meta.url`
+          ? `{"name":${JSON.stringify(p)},"mimeType":${JSON.stringify(
+              mime.getType(name) ?? undefined
+            )},"path":${annotatePath(relativePath(servePath, resolveFile(name)))},"lastModified":${JSON.stringify(
+              info.mtimeMs
+            )},"size":${JSON.stringify(info.size)}`
```
mythmon, line 120: This string is missing a trailing `}`, which makes the output file invalid.

image

Would it be possible to generate everything here as a real JS object and the `JSON.stringify` it so that we don't have to worry about JSON syntax?

test/javascript/transpile-test.ts
```diff
@@ -210,11 +210,11 @@ describe("transpileModule(input, root, path)", () => {
   it("rewrites npm imports", async () => {
     const input = 'import "npm:d3-array";';
     const output = (await transpileModule(input, options)).split("\n").pop()!;
-    assert.strictEqual(output, 'import "../_npm/d3-array@3.2.4/_esm.js";');
+    assert.strictEqual(output, 'import "../_npm/d3-array@3.2.4/_esm.js"/* observablehq-file */;');
```
mythmon, line 213: If this link is tagged as `observable-file`, then we should also rewrite it to tag Observable files in it. If we don't rewrite it, I don't think we should tag it.

test/output/build/data-loaders/_import/test.86a60bc6.js
```diff
@@ -1,3 +1,3 @@
-import {FileAttachment} from "../_observablehq/stdlib.00000003.js";
+import {FileAttachment} from "../_observablehq/stdlib.00000003.js"/* observablehq-file */;
 
-export const test = FileAttachment({"name":"../test.txt","mimeType":"text/plain","path":"../_file/test.f2ca1bb6.txt","lastModified":/* ts */1706742000000,"size":5}, import.meta.url).text();
+export const test = FileAttachment({"name":"../test.txt","mimeType":"text/plain","path":"../_file/test.f2ca1bb6.txt"/* observablehq-file */,"lastModified":/* ts */1706742000000,"size":5, import.meta.url).text();
```
mythmon, line 3: You can see the missing `}` here too. This should be the below, I think
```suggestion
export const test = FileAttachment({"name":"../test.txt","mimeType":"text/plain","path":"../_file/test.f2ca1bb6.txt"/* observablehq-file */,"lastModified":/* ts */1706742000000,"size":5}, import.meta.url).text();
```

mythmon (changes requested):

src/javascript/transpile.ts
Fil: I don't believe JSON.stringify can generate comments. Thanks for catching the bug!

test/javascript/transpile-test.ts
Fil: I was initially doing that (see line 307 in https://github.com/observablehq/framework/pull/1731/commits/6cfb74de17c13d54608a14dd396bf56a97b30e2f#diff-a479e6903ca67c840829983998e33849155911842cf08fa118b42b2d88fe3c9b), then I think we decided to tag everything and let the server decide what it should do with it.

mythmon: My concern is that this is a reference to an `_npm` file that is tagged, but references to `_npm` files from other `_npm` files are not tagged. As a concrete example, in the docs build there is this chain:

1. `dist/chart.js` tags its import to the `_import` file:
2. `docs/.observablehq/dist/_import/chart.95e875cf.js` tags its import of an `_npm` file:
3. `docs/.observablehq/dist/_npm/@observablehq/plot@0.6.16/e828d8c8.js` does not tag its import of the `_npm` file:
4. `../../isoformat@0.2.1/c68fbd73.js`

It would be helpful if the tagging was done based on the taget of the URL, not the place the URL occurs. As it stands, the server can't really make the decision fully, because links to `_npm` files aren't consistent tagged/not tagged, so it _has_ to special case them as not needing extra authentication.

mythmon (approved):
I think this will work. Thanks!

mythmon:
@Fil is there anything more you want to do on this PR before merging it? It would be nice to have this in Framework's main branch for further development.

Fil:
Not on my side; I just wanted to give Mike a chance to review.

mbostock (commented):
Do we want this to be always on, or do we only want to turn it on when building for Cloud? I’d slightly prefer the latter and believe that was what we proposed (via environment variable)?

Fil:
Annotations are now conditioned on the `ANNOTATE_FILES=true` environment variable. (Any other value throws an error.)

mootari:
We should probably make that `OBSERVABLE_ANNOTATE_FILES`?

Fil:
Sure, or `OBSERVABLE_ANNOTATE` for short. I don't have an opinion on this :)

mythmon:
I'd prefer if all environment variables that need to be set in Cloud Builds begin with `OBSERVABLE_`. I don't have a preference for `OBSERVABLE_ANNOTATE` vs `OBSERVABLE_ANNOTATE_FILES`.

package.json
```diff
@@ -26,8 +26,9 @@
...
-    "test:mocha": "yarn test:build && rimraf --glob test/.observablehq/cache test/input/build/*/.observablehq/cache && cross-env OBSERVABLE_TELEMETRY_DISABLE=1 TZ=America/Los_Angeles mocha --timeout 30000 -p \"test/build/test/**/*-test.js\"",
-    "test:mocha:serial": "yarn test:build && rimraf --glob test/.observablehq/cache test/input/build/*/.observablehq/cache && cross-env OBSERVABLE_TELEMETRY_DISABLE=1 TZ=America/Los_Angeles mocha --timeout 30000 \"test/build/test/**/*-test.js\"",
+    "test:mocha": "yarn test:build && rimraf --glob test/.observablehq/cache test/input/build/*/.observablehq/cache && cross-env OBSERVABLE_TELEMETRY_DISABLE=1 TZ=America/Los_Angeles mocha --timeout 30000 -p \"test/build/test/**/*-test.js\" && yarn test:annotate",
+    "test:mocha:serial": "yarn test:build && rimraf --glob test/.observablehq/cache test/input/build/*/.observablehq/cache && cross-env OBSERVABLE_TELEMETRY_DISABLE=1 TZ=America/Los_Angeles mocha --timeout 30000 \"test/build/test/**/*-test.js\" && yarn test:annotate",
+    "test:annotate": "yarn test:build && cross-env OBSERVABLE_ANNOTATE_FILES=true TZ=America/Los_Angeles mocha --timeout 30000 \"test/build/test/**/annotate.js\"",
```
mbostock, line 31: This pattern breaks `yarn test:mocha:serial` with `it.only` or `describe.only` because it always runs the `test:annotate` in addition to your only’d tests.

src/javascript/annotate.ts
```diff
@@ -0,0 +1,14 @@
+import {isPathImport} from "../path.js";
+
+/**
+ * Annotate a path to a local import or file so it can be reworked server-side.
+ */
+
+const annotate = process.env["OBSERVABLE_ANNOTATE_FILES"];
+if (typeof annotate === "string" && annotate !== "true")
+  throw new Error(`unsupported OBSERVABLE_ANNOTATE_FILES value: ${annotate}`);
+export default annotate
```
mbostock, line 10: We don’t use default exports (because there’s no real reason to, and it encourages more inconsistency in how people name an imported symbol). This should be called `annotatePath` or something.

mbostock (commented):
Noting a few things I plan on fixing.

Issue #1598, opened 2024-08-22
Allow vgplot to use DuckDBClient


angrytongan:
From <https://talk.observablehq.com/t/observable-framework-using-duckdbclient-sql-with-vgplot/9676/3>:

I would like to load data using DuckDBClient.sql() and display with vgplot. I am able to do this with Observable Plot:

````
```js
const sql = DuckDBClient.sql({ rand: FileAttachment("./rand-xy.csv") });
```

```sql id=data
SELECT * FROM rand
```

```js
display(
  Plot.plot({
    height: 200,
    marks: [
      Plot.dot(data, {
        x: "x",
        y: "y",
      }),
    ],
  })
);
```
````

but not with vgplot:

````
```js
const sql = DuckDBClient.sql({ rand: FileAttachment("./rand-xy.csv") });
```

```js
display(
  vg.plot(
    vg.height(200),
    vg.dot(vg.from("rand"), {
      x: "x",
      y: "y",
    })
  )
);
```
````

@mbostock responded:

> Overriding the definition of sql doesn’t work (currently) with our implementation of vgplot. You’ll need to use the [SQL front matter](https://observablehq.com/framework/sql) instead.

**Comment by mbostock** [comment]
For now, the easiest way to do this would be to provide your own definition of `vg`, replacing the use of the default DuckDB client in [`vgplot.js`](https://github.com/observablehq/framework/blob/bbea2c504854c9e80672795207d12a781d0eed8b/src/client/stdlib/vgplot.js#L1-L10). That would look like this:

```js
import * as vgplot from "npm:@uwdata/vgplot";

const db = await DuckDBClient.of({trips: FileAttachment("lib/nyc-taxi.parquet")});
const sql = db.sql.bind(db);

const coordinator = new vgplot.Coordinator();
const vg = vgplot.createAPIContext({coordinator});
coordinator.databaseConnector(vgplot.wasmConnector({duckdb: db._db}));
```

angrytongan:
Thanks @mbostock , works a treat 👍

mbostock:
Not sure what else there is to do here. Maybe the `vg` built-in should default to consuming `sql._db` as its DuckDB instance rather than the internal `getDefaultClient` method, so that when `sql` is redefined on the page, `vg` automatically inherits the altered definition?

Or is it sufficient to just have people redefine `vg` if they also want to redefine `sql` as shown in https://github.com/observablehq/framework/issues/1598#issuecomment-2319374786?

[...]

Issue #1508, opened 2024-07-09
Timezone Trouble with DuckDB and Observable plot


c-koster:
Hi all, 

I've been having some trouble working with some timezone-aware DuckDB SQL statements. I am trying to use some timestamptz operations which exist in the [DuckDB documentation](https://duckdb.org/docs/sql/functions/timestamptz.html#date_diffpart-startdate-enddate) but don't appear to be installed in framework.

For example, this query
```sql
SELECT date_diff('day', TIMESTAMPTZ '1992-09-30 23:59:59', TIMESTAMPTZ '1992-10-01 01:58:00')
```
fails but the following one works fine:
```sql
SELECT date_diff('day', TIMESTAMP '1992-09-30 23:59:59', TIMESTAMP '1992-10-01 01:58:00')
```
I think this means that the ICU library isn't installed and also doesn't get auto-loaded when one of these functions is called. I also tried preceding the first query with `INSTALL icu;` and `LOAD icu;` but got the following error from the install step:
```
Error: IO Error: Can't find the home directory at '/home/web_user'
Specify a home directory using the SET home_directory='/path/to/dir' option.
```

Is there any way to import non-standard modules to the Duck DB instance? I think this is related to #1070.

Thank you!

Issue #1380, opened 2024-05-22
More SQL configuration options


mbostock:
It’d be nice if there were a way to change these defaults when using SQL code blocks:

https://github.com/observablehq/framework/blob/97cc4b522eef7d9a49d477ea45bf541857501f6a/src/client/stdlib/duckdb.js#L165-L170

Issue #1070, opened 2024-03-15
Better support for DuckDB extensions


Fil:
It would be great to be able to install/load [DuckDB extensions](https://duckdb.org/docs/extensions/overview.html) as simply as:

```yaml
---
sql:
  install: [httpfs, spatial]
  tables:
    taxis: nyc-taxis.parquet
---
```

https://github.com/observablehq/framework/issues/1007

(This means we'll need to extend the shape of the sql front matter option, see #1054)

Fil:
We should also enable autoloading, I guess. And we should copy the extension files to the local cache — currently even the parquet extension is loaded directly from the  extensions.duckdb.org domain (i.e. `https://extensions.duckdb.org/v1.0.0/wasm_eh/parquet.duckdb_extension.wasm`.

jaanli:
Yes @Fil this would be amazing - we mainly use this extension: https://github.com/duckdb/duckdb_vss

do you have a sense of what it would take to support this? 

I’m free the next month to help with our team @onefact because I am giving a talk at the @duckdb conference:

 https://duckdb.org/2024/08/15/duckcon5.html

I would love to help on this and demo it in August!

Fil:
I recently added an example here
https://observablehq.observablehq.cloud/pangea/party/duckdb-spatial

it's not very rich feature-wise, but it shows that it is technically possible

PR #1734, opened 2024-10-09
duckdb 1.29.0; self-host extensions

Fil:
🎉 1.29.0! new version of DuckDB-wasm 🎉

https://github.com/duckdb/duckdb-wasm/releases/tag/v1.29.0

The repo had [296 commits](https://github.com/duckdb/duckdb-wasm/compare/v1.28.0...v1.29.0) since the last stable release a year ago. This is not including the commits on the linked DuckDB itself, which is now in version 1.1.1.

See https://duckdb.org/2024/09/09/announcing-duckdb-110 for the new features and changes in DuckDB. For example, the nice HISTOGRAM() function:

duckdb-histogram

The most notable new feature in duckdb-wasm is the support for **extensions**, in particular the "spatial" extension which includes the whole of GDAL, enabling geographic compute (projections, areas, etc), and introducing compatibility with dozens of new formats (shapefiles, excel sheets, etc.).

* https://duckdb.org/2023/12/18/duckdb-extensions-in-wasm.html
* https://duckdb.org/docs/extensions/spatial/overview.html
* https://gdal.org/
* https://duckdb.org/2024/09/09/announcing-duckdb-110
* https://github.com/duckdb/duckdb/releases/tag/v1.1.1

Other extensions: autocomplete, fts, icu, inet, json, parquet, spatial, sqlite_scanner, substrait, tpcds, tpch, vss.

related:
- #1070
- #1508
- #1598 (maybe)
- #1009
- #1380
- https://github.com/observablehq/framework/discussions/834


Fil:
*About the package size*

Correlatively to the new features, this new release weighs a ton: the base files have doubled in size, and with the addition of extensions the binaries now take **153M of disk space** on the server.

Fortunately this is not what the user has to download. First, depending on the browser used, they will only download the "mvp" version (older browsers) or the "eh" version of the wasm files. Second, they will not load all extensions (only "spatial" is quite big). Third, the wasm files are gzip'ed when transmitted to the browser.

But it's still doubling the (compressed) size of the base files from 4MB to ~8MB (depending on the extensions needed… here I compare 1.28.0 with 1.29.0+parquet). Is there a case to be made for users who would prefer to stay with 1.28 because of that? I prefer not to, since it would add much complexity.


*About self-hosting extensions*

A key feature of Framework is self-hosting. I didn't want to support 1.29 without self-hosting at least the extensions that used to be part of the monolithic 1.28 ("parquet", "json"). The status of extensions is however still a bit unclear to me. Some of the core extensions are built-in (such as httpfs). It's unclear how that list will change in the future (httpfs changed status during development, I think). So instead of linking to `duckdb-wasm@latest`, I thought it better to continue pinning the version, so as to prevent unexpected changes.

Moreover, only the core extensions are self-hosted for now, and we might want a path also for people who want to self-host community extensions (such as "h3") or custom extensions.

Maybe self-hosting "all the core extensions" is too much, and we could have a smaller list of extensions we self-host. However judging from the sizes of extensions, "spatial" dwarfs all the others—so it might not make sense to try and optimize disk space if we keep "spatial". Another option would be to make a configurable list of self-hosted extensions (including community and custom extensions). We would then have to pass that list to `client/duckdb.js` to install. More configuration means more complexity, though.

mbostock (commented):
Rather than overloading the meaning of `https://extensions.duckdb.org/` and `npm:extensions.duckdb.org`, we’d probably want a `duckdb:` protocol for specifying extensions, and to put them in `_duckdb` parallel to `_npm`. But that’s quite a bit of machinery to support DuckDB extensions…

src/client/stdlib/duckdb.js
```diff
@@ -182,6 +182,40 @@ Object.defineProperty(DuckDBClient.prototype, "dialect", {
...
+      "parquet",
+      // "postgres_scanner",
+      "spatial",
+      "sqlite_scanner",
+      "substrait",
+      "tpcds",
+      "tpch",
+      "vss"
+    ]
+      .map((ext) => `INSTALL ${ext} FROM '${repo}';`)
```
mbostock, line 213: How do these paths get content-hashed and/or versioned (for immutable caching)?

Fil:
An alternative approach could be to publish a package on jsr or npm with the extensions we want to self-host.

src/client/stdlib/duckdb.js
Fil: I believe they are versioned by the `1.1.1` in their path, but I'm not sure of what happens server-side in duckdb-land. I'm talking with @carlopi to understand this better.

mbostock:
I guess my inclination is to have users explicitly list which DuckDB extensions they want, and where they come from. And then Framework can download them for self-hosting. So maybe in the config you would say something like:

```js
export default {
  duckdb: {
    extensions: {
      json: "https://extensions.duckdb.org/v1.1.1/wasm_eh/json.duckdb_extension.wasm",
      parquet: "https://extensions.duckdb.org/v1.1.1/wasm_eh/parquet.duckdb_extension.wasm"
    }
  }
};
```

If we wanted to have shorthand, we could also allow something like:

```js
export default {
  duckdb: {
    extensions: {
      json: true,
      parquet: true
    }
  }
};
```

Or even shorter:

```js
export default {
  duckdb: {
    extensions: ["json", "parquet"]
  }
};
```

So Framework would self-host the specified files. And internally we’d have some resolution magic so that `DuckDBClient` knows where to find the self-hosted extensions. And if we’re allowing arbitrary URLs for extensions we’d need to use content hashing so that if the content of the extension changes it’s still immutably cached.

Fil:
*About the `LOAD` statements*

Currently if several sql blocks use spatial functions (for example), you have to remember to type `LOAD spatial` in all of them, otherwise it's hard to predict which queries will run after it's loaded or run (and fail) before it's loaded. Only the first to run is actually loading it to the DuckDB instance, so it's a bit of a waste.

To avoid this issue we could maybe hoist any `LOAD` statement, so that you can have `LOAD spatial` in just one of your sql code blocks instead of having to repeat it in every block that needs this extension. This means static analysis of the sql code, but it's probably not too bad(?).

Or, maybe simpler, we could add a top-level config in front-matter. Something like:
```
sql:
  - load: [spatial, h3]
  - 
```
or keep the `sql` key for tables, and add a new key for duckdb options
```
duckdb:
  - load: [spatial, h3]
  - 
```

(we could also make it possible to reference an Excel or Shapefile dataset in front-matter, since `spatial`’s ST_Read function supports so many formats?)

mbostock:
@Fil In my previous comment I meant that could be specified in the project config. But we could also let it be specified in the page front matter, overriding the project config if different pages want different extensions.

Fil:
The config option would indicate which extensions are self-hosted and where they're sourced from. Thus, they would be `INSTALL`ed from the self-hosted version. But `INSTALL` only tells duckdb where to find to the wasm binary, it doesn't actually load it to the browser.

For many core extensions this is happening implicitly, when duckdb recognizes that one the functions or file formats used belongs to a given extension (the extension is then said to be “auto-loaded”). The documentation in `lib/duckdb.md` shows this with the "inet" example. For other extensions, such as "spatial", you have to give an explicit `LOAD` statement before you can use any of the features.

For performance reasons I don't think we should _load_ all the extensions (not even all the self-hosted extensions) on all the pages, but it should depend on what the page needs. However currently, when an extension needs to be loaded explicitly, it has to be mentioned in every sql code block, because their order of execution is not guaranteed. That's a bit too much, and I think the correct level to define these `LOAD` statements is the DuckDBClient instance—or more simply, the page.

mbostock:
Right, so the config could say whether to load the extension explicitly or to let it autoload if desired. But in either case the installing and loading of any desired extensions would happen prior to the `sql` literal resolving so that downstream code can rely on the extensions being available.

Having equivalent extension registration for the front matter as for the project config makes sense.

docs/lib/duckdb.md
```diff
@@ -124,42 +124,41 @@ LOAD h3;
...
-### Self-hosted core extensions
-
-Framework downloads a copy of the [core extensions](https://duckdb.org/2023/12/18/duckdb-extensions-in-wasm.html), and the DuckDBClient installs them by default. This ensures that all the common extensions ("json", "inet", "spatial", etc.), are self-hosted.
-
-You can however override this (for example, if you need to test something against a new version of an extension), and install explicitly:
-
-```sql echo run=false
-INSTALL json FROM core;
--- use JSON features
+INSTALL custom FROM 'https://example.com/';
```
mbostock, line 130: I think we should discourage people from installing extensions from within SQL blocks: doing so globally changes the behavior of the DuckDBClient instance and can lead to race conditions/nondeterministic behavior across blocks, and also because we want to favor self-hosting of extensions rather than hotlinking to an external website.

The recommended way to install extensions should be via the front matter or the project config (or to do it in JavaScript by redefining the `sql` literal and awaiting the loading of the extensions).

Fil:
Getting closer.

TODO:
- [ ] configuration to allow the "core" and "community" keywords
- [ ] decide which extensions are loaded and which aren't (typically, "json" and "parquet" don't need to be loaded since they're autoloaded)
- [ ] allow front-matter configuration (or maybe punt on that for now… it's already quite complex)

src/client/stdlib/duckdb.js
```diff
@@ -182,6 +183,22 @@ Object.defineProperty(DuckDBClient.prototype, "dialect", {
   value: "duckdb"
 });
 
+async function registerExtensions(db) {
+  const connection = await db.connect();
+  try {
+    const extensions = await fetch(import.meta.resolve("observablehq:duckdb_manifest.json")).then((r) => r.json());
```
mbostock: We’ll want to generate code for this on the page, like we do for `registerTable`. When we render the page we’ll know what the desired extensions are, and we can register them so that the `sql` variable awaits their loading.

mbostock: Though now with exported modules, we might also want to consider how a local module could use a default DuckDB client — I guess we need an equivalent JavaScript API for installing and loading self-hosted extensions for when you can’t use the front matter/page initialization.

Fil: I've implemented a method to allow all the flexibility we want — you have to go through DuckDBClient if you don't want the default LOADs from the configuration, but I think that's fine.

mbostock: The other reason I want to generate the code is that it’ll be a lot faster since you won’t have to wait for a manifest.json to load.

Fil: yes the manifest is now appropriately baked-in

docs/lib/duckdb.md
```diff
@@ -105,3 +105,13 @@ const sql = DuckDBClient.sql({quakes: `https://earthquake.usgs.gov/earthquakes/f
...
 ```
+
+## Extensions
+
+DuckDB’s [extensions](../sql#extensions)<a href="https://github.com/observablehq/framework/pull/1734" class="observablehq-version-badge" data-version="prerelease" title="Added in #1734"></a> are supported.
+
+By default, `DuckDBClient.of` and `DuckDBClient.sql` load the (self-hosted) extensions referenced in the [configuration](../config#duckdb). You can pass an options object with a **load** key set to a different list if you want a different environment:
+
+```js echo run=false
+const geodb = await DuckDBClient.of({}, {load: ["spatial", "h3"]});
```
mbostock, line 116: Are you only able to load extensions that were previously registered in the config? Or would it be possible to register a “one-off” extension here? And presumably you’d need to use `import.meta.resolve` to self-host the URL to the extension.

docs/sql.md
```diff
@@ -206,3 +206,51 @@ Inputs.table(await sql([`SELECT * FROM gaia WHERE source_id IN (${[source_ids]})
...
+
+However, for performance and ergonomy, we strongly recommend adding all the extensions you actually use to your site’s configuration.
+
+<div class="tip">
+
+Since some extensions are autoloading, it can be hard to tell which ones are
+effectively in use. You can inspect the network tab in your browser, or (in
+preview only), open your browser console and type `DuckDBClientReport()` — this
+utility offers additional information, such as the path used for self-hosted
+extensions.
```
mbostock, line 252: Don’t hard-wrap Markdown.

src/build.ts
```diff
@@ -140,6 +141,20 @@ export async function build(
...
+      const sourcePath = join(cacheRoot, path);
+      effects.output.write(`${faint("build")} ${path} ${faint("→")} `);
+      const contents = await readFile(sourcePath);
+      const hash = createHash("sha256").update(contents).digest("hex").slice(0, 8);
+      const alias = applyHash(path, hash);
+      aliases.set(path, alias);
+      await effects.writeFile(alias, contents);
+    }
+  }
+  const duckdb_manifest = await duckDBManifest(duckdb, {root, log: true, aliases});
```
mbostock, line 156:
```suggestion
  const duckDbManifest = await getDuckDbManifest(duckdb, {root, log: true, aliases});
```

Style:

* Use lowerCamelCase instead of lower_snake_case
* Only capitalize the first letter or acronyms
* Use a verb to distinguish between a function that generates something and the noun result

src/build.ts
```diff
@@ -149,6 +164,9 @@ export async function build(
       effects.output.write(`${faint("bundle")} ${path} ${faint("→")} `);
       const clientPath = getClientPath(path === "/_observablehq/client.js" ? "index.js" : path.slice("/_observablehq/".length)); // prettier-ignore
       const define: {[key: string]: string} = {};
+      if (path === "/_observablehq/stdlib/duckdb.js") {
+        define["process.DUCKDB_MANIFEST"] = JSON.stringify(duckdb_manifest);
```
mbostock, line 168: Seems a little strange to use `process.DUCKDB_MANIFEST` as the magic here. Like either we’d pretend it’s an environment variable as `process.env.DUCKDB_MANIFEST` or just make it a global `DUCKDB_MANIFEST`.

docs/sql.md
mbostock: I don’t think we should add a new global `DuckDBClientReport` just for debugging in preview.

src/build.ts
```diff
@@ -398,6 +416,7 @@ function validateLinks(outputs: Map<string, {resolvers: Resolvers}>): [valid: Li
 }
 
 function applyHash(path: string, hash: string): string {
+  if (path.startsWith("/_duckdb/")) return join("/_duckdb/", `${hash}-${path.slice("/_duckdb/".length)}`);
```
mbostock, line 419: First, why do we need special hashing behavior for DuckDB files? And second, if we do, we should have a separate function instead of putting this into `applyHash`. The only place we currently call `applyHash` with a path that starts with `/_duckdb/` is on L151, so we can just do the special behavior there. But I’d like to avoid the special behavior unless it’s necessary.

src/client/stdlib/duckdb.js
```diff
@@ -29,16 +29,26 @@ import * as duckdb from "npm:@duckdb/duckdb-wasm";
...
-    mainModule: import.meta.resolve("npm:@duckdb/duckdb-wasm/dist/duckdb-mvp.wasm"),
-    mainWorker: import.meta.resolve("npm:@duckdb/duckdb-wasm/dist/duckdb-browser-mvp.worker.js")
-  },
-  eh: {
-    mainModule: import.meta.resolve("npm:@duckdb/duckdb-wasm/dist/duckdb-eh.wasm"),
-    mainWorker: import.meta.resolve("npm:@duckdb/duckdb-wasm/dist/duckdb-browser-eh.worker.js")
-  }
-});
+// Baked-in manifest.
+const manifest = process.DUCKDB_MANIFEST;
```
mbostock, line 33: Thinking out loud… baking the manifest into /_observablehq/stdlib/duckdb.js means that the content hash will change when the DuckDB manifest changes… and that by extension the module hash of /_observablehq/stdlib.js will change because it imports /_observablehq/stdlib/duckdb.js. We should double-check that this works but I believe the hashing is fairly robust now. 😅

src/client/stdlib/duckdb.js
```diff
@@ -29,16 +29,26 @@ import * as duckdb from "npm:@duckdb/duckdb-wasm";
...
+      mainModule: import.meta.resolve("npm:@duckdb/duckdb-wasm/dist/duckdb-mvp.wasm"),
+      mainWorker: import.meta.resolve("npm:@duckdb/duckdb-wasm/dist/duckdb-browser-mvp.worker.js")
+    }
+  }),
+  ...(manifest.bundles.includes("eh") && {
+    eh: {
+      mainModule: import.meta.resolve("npm:@duckdb/duckdb-wasm/dist/duckdb-eh.wasm"),
+      mainWorker: import.meta.resolve("npm:@duckdb/duckdb-wasm/dist/duckdb-browser-eh.worker.js")
+    }
+  })
```
mbostock, line 47: Is now the right time to simplify and drop support for the mvp bundle and only use eh? Or better to decouple?

src/client/stdlib/duckdb.js
```diff
@@ -178,9 +189,38 @@ export class DuckDBClient {
   }
 }
 
-Object.defineProperty(DuckDBClient.prototype, "dialect", {
-  value: "duckdb"
-});
+Object.defineProperty(DuckDBClient.prototype, "dialect", {value: "duckdb"});
```
mbostock, line 192: The DatabaseClient.dialect isn’t used for anything in Framework. I think it’s used in notebooks for ejecting from a SQL cell, but there’s no analogous concept in Framework so maybe we should consider dropping it. Not directly related to this PR though!

src/client/stdlib/duckdb.js
```diff
@@ -178,9 +189,38 @@ export class DuckDBClient {
...
+          const refs = new Map(extensions);
+          const ext = await connection.query(
+            "SELECT extension_name, description FROM duckdb_extensions() WHERE loaded;"
+          );
+          console.table(Array.from(ext, (e) => ({...e, ...refs.get(e.extension_name)})));
+        } finally {
+          await connection.close();
+        }
+      };
+    }
```
mbostock, line 212: I think we should remove this as I noted above. I think we can just document “run this SQL if you want to know what DuckDB extensions are loaded” and not make a helper method for it.

Principles:

- Don’t pollute global scope
- Functions should be bound to appropriate scope (here the DuckDBClient/`db` instance)
- Minimize differences in logic between production and development
- Favor minimal API surface area so it’s easier for us to keep things light
- Favor methods that return data, not that have a side-effect of console.log/console.table

If we also want to expose the manifest/registered extensions and not just what DuckDB sees, then it should be a method on DuckDBClient. Or if want to expose the manifest, it should be a static property on DuckDBClient. But better to minimize what we expose so we have more flexibility to change it in the future, and there’s less for us to document, and less for readers to read.

src/client/stdlib/duckdb.js
```diff
@@ -178,9 +189,38 @@ export class DuckDBClient {
...
+    await Promise.all(
+      extensions.map(([name, {ref, load: l}]) =>
+        connection
+          .query(`INSTALL ${name} FROM '${import.meta.resolve(`../..${ref}`)}'`)
```
mbostock, line 216: We might want to do some static validation that extension names are \w+ to avoid arbitrary SQL injection when registering an extension.

src/client/stdlib/duckdb.js
Fil: Though it would be simpler to support only one bundle type, I think it's better to decouple and keep the possibility to support several. (Although I don't know what extensions are loaded if you use "coi". duckdb.org doesn't seem to host coi extensions.)

src/config.ts
```diff
@@ -488,3 +499,24 @@ export function mergeStyle(
...
+  const source = new Map(Object.entries(spec?.["source"] ?? {}));
+  return {
+    bundles: ["eh", "mvp"],
+    install,
+    load: load.filter((name: string) => install.includes(name)),
+    source: Object.fromEntries(
+      install.map((name: string) => {
+        let href = source.get(name) ?? "core";
+        if (href === "core") href = "https://extensions.duckdb.org";
+        else if (href === "community") href = "https://community-extensions.duckdb.org";
```
mbostock, line 515: If we want to also allow registration of (potentially non-self-hosted) extensions at runtime, this will need to move to src/client/stdlib/duckdb.js rather than here… it might also be more appropriate to move this logic to a function exported by src/duckdb.ts so that it’s more centralized.

Like what if you could say:

```js
DuckDBClient.of(…, {
  extensions: {
    spatial: {
      load: true,
      source: import.meta.resolve("duckdb:core:spatial")
    }
  }
})
```

Potentially with shorthand (we could know that `spatial` isn’t one of the automatically loaded extensions and load it by default):

```js
DuckDBClient.of(…, {extensions: {spatial: import.meta.resolve("duckdb:core:spatial")}})
```

Because this uses `import.meta.resolve` this would mean that the spatial extension would be self-hosted the same way as if you had used the **duckdb** config option (or front matter), i.e., it would come from /_duckdb/.

src/client/stdlib/duckdb.js
Fil: I think this works, yes. But always double check.

src/config.ts
```diff
@@ -488,3 +499,24 @@ export function mergeStyle(
...
+        if (!href?.["startsWith"]?.("https://"))
+          throw new Error(`unknown source for duckdb extension ${name}: ${href}`);
```
mbostock, line 517: What if you want to load an extension that you’ve manually copied into your source root?

src/build.ts
Fil: We need this because when you `INSTALL foo FROM host` DuckDB will load it from `${host}/v1.1.1/wasm_{p}/parquet.duckdb_extension.wasm` and not accept anything else.

(If we could give the full path I would have gone with `${host}/v1.1.1/wasm_{p}/parquet.duckdb_extension.${hash}.wasm` but that doesn't seem to be supported. Maybe I missed something, though.)

src/duckdb.ts
```diff
@@ -0,0 +1,97 @@
...
+ * for every supported platform p ("eh" and "mvp"), and returns a content-hashed
+ * reference (_duckdb/{hash}) to use in the corresponding DuckDB INSTALL
+ * statement. The repo is structured as required by DuckDB with:
+ * ${repo}/v1.1.1/wasm_{p}/${name}.duckdb_extension.wasm
+ */
+export async function resolveDuckDBExtension(root: string, p: string, repo: string, name: string): Promise<string> {
+  if (!repo.startsWith("https://")) throw new Error(`invalid repo: ${repo}`);
+  const cache = join(root, ".observablehq", "cache");
+  const file = `${name}.duckdb_extension.wasm`;
+  const ref = `${repo}/v1.1.1/wasm_${p}/${file}`.slice("https://".length);
```
mbostock, line 46: Where does the `v1.1.1` come from? Will this need to change when new releases of DuckDB come out? I would like to minimize the number of things we need to maintain and ideally allow Framework to pick up new releases of DuckDB automatically. If we bake-in versions here then we become the bottleneck for upgrading. The main reason we pinned the version of DuckDB-Wasm before was they made the mistake of publishing the prerelease of 1.29.0 under the `latest` disttag (and didn’t correct it for months). It’s great that we’re adding self-hosting of DuckDB extensions to Framework, but otherwise I’d still like to minimize how much we’re “in the loop” for new DuckDB-Wasm releases.

src/client/stdlib/duckdb.js
Fil: It was very useful to me while debugging, but there's no point in keeping this. The network tab is enough to know which extensions are loaded.

src/duckdb.ts
```diff
@@ -0,0 +1,97 @@
...
+}
+
+async function duckDBHash(name: string, files: string[]): Promise<string> {
+  const hash = createHash("sha256");
+  hash.update(name);
+  for (const file of files) hash.update(await readFile(file, "utf-8"));
+  return join("_duckdb", `${name}-${hash.digest("hex").slice(0, 8)}`);
+}
+
+*/
```
mbostock, line 97:
```suggestion
```

src/npm.ts
```diff
@@ -162,7 +162,7 @@ export async function getDependencyResolver(
         (name === "arquero" || name === "@uwdata/mosaic-core" || name === "@duckdb/duckdb-wasm") && depName === "apache-arrow" // prettier-ignore
           ? "latest" // force Arquero, Mosaic & DuckDB-Wasm to use the (same) latest version of Arrow
           : name === "@uwdata/mosaic-core" && depName === "@duckdb/duckdb-wasm"
-          ? "1.28.0" // force Mosaic to use the latest (stable) version of DuckDB-Wasm
+          ? "1.29.0" // force Mosaic to use the latest (stable) version of DuckDB-Wasm
```
mbostock, line 165: Can we use `latest` here instead?

src/npm.ts
```diff
@@ -248,9 +248,7 @@ async function resolveNpmVersion(root: string, {name, range}: NpmSpecifier): Pro
 export async function resolveNpmImport(root: string, specifier: string): Promise<string> {
   const {
     name,
-    range = name === "@duckdb/duckdb-wasm"
-      ? "1.28.0" // https://github.com/duckdb/duckdb-wasm/issues/1561
-      : undefined,
+    range = name === "@duckdb/duckdb-wasm" ? "1.29.0" : undefined,
```
mbostock, line 251: Can we remove the pinning of DuckDB-Wasm now?

src/duckdb.ts
Fil: Given the history here, I think we're better off manually locking the versions. (Agree that, ideally, we wouldn't have to.) I think `v1.1.1` is the underlying DuckDB version, but I can't find the explanation in the docs.

src/duckdb.ts
```diff
@@ -0,0 +1,97 @@
...
+
+/**
+ * Given a duckdb configuration and an extension name such as "parquet", saves
+ * the binary to _duckdb/{hash}/v1.1.1/wasm_{p}/parquet.duckdb_extension.wasm
+ * for every supported platform p ("eh" and "mvp"), and returns a content-hashed
+ * reference (_duckdb/{hash}) to use in the corresponding DuckDB INSTALL
+ * statement. The repo is structured as required by DuckDB with:
+ * ${repo}/v1.1.1/wasm_{p}/${name}.duckdb_extension.wasm
+ */
+export async function resolveDuckDBExtension(root: string, p: string, repo: string, name: string): Promise<string> {
```
mbostock, line 42: This is downloading the extension greedily during resolution, but it looks to me like we could defer that to later, and that we know synchronously what the URL of the extension will be. If we can make DuckDB extension resolution synchronous, that might be nice for simplicity and performance (only having the Preview server fetch the file lazily when requested). But this might be related to my [other question](https://github.com/observablehq/framework/pull/1734/files#r1807904150) on the `v1.1.1`; I’m guessing we need to resolve the latest version of the extension, in which case this still needs to be asynchronous.

mbostock, line 42: The `p` name here isn’t great. I try to only use cryptic single-letter variables for short-lived local variables; function arguments should ideally have more descriptive names, especially for functions that are exported from the module. Should be `platform` instead, right?

src/duckdb.ts
mbostock: Not sure if there is any way other than `SELECT VERSION()` to get the version of DuckDB that DuckDB-Wasm uses under the hood… and we’d need to make sure that we’re using the right version of DuckDB-Wasm when doing the resolution here. So maybe pinning both is the only viable option for now. Still, we could at least pull the `v1.1.1` up to a global variable, and keep the DuckDB and DuckDB-Wasm versions defined in the same place so that we’re more likely to update both of them at the same time.

src/config.ts
mbostock: I’m not sure if `import.meta.resolve("duckdb:core:spatial")` should resolve to the full path to the extension…

```
/_duckdb/extensions.duckdb.org/v1.1.1/wasm_eh/spatial.duckdb_extension
```

Or if it should resolve to the self-hosted [“custom repository”](https://duckdb.org/docs/extensions/working_with_extensions#creating-a-custom-repository) folder:

```
/_duckdb/extensions.duckdb.org/
```

Since the latter is the path that you’d need to pass to `INSTALL spatial FROM '<repository>'`; The former is only possible if we only support the `eh` platform. So, I think I’m inclined to do the latter. So `import.meta.resolve("duckdb:core:spatial")` would download the needed files for that extension from extensions.duckdb.org, but the resulting value would be just the custom repository location, meaning that e.g. `import.meta.resolve("duckdb:core:parquet")` would resolve to the same value (even though it would download different files).

src/config.ts
mbostock: Maybe possible if you said `import.meta.resolve("duckdb:local:foo")`? But this is probably so rare that we don’t need to support this from the get-go.

src/build.ts
mbostock: Yep makes sense once I grokked DuckDB’s requirements on the custom repositories for extensions. But I still think we should break out this into a separate function rather than overloading `applyHash`, and I’m not sure we need content hashes on anything under `_duckdb/` assuming that they don’t change the contents of extensions independently of the DuckDB version. If they do, then we’ll need a separate custom repository per extension since they’ll all have different hashes…

mbostock:
Lots here! Excited about this. I’ll try to help this week.

src/duckdb.ts
Fil: I think so, though they use the word "bundle", and adjust it to the "platform" (which are characteristics of the browser) in the `selectBundle` function:
https://github.com/duckdb/duckdb-wasm/blob/a9a38bb480c1b44bb90e6fa35e731c572e8868ef/packages/duckdb-wasm/src/platform.ts#L27

src/duckdb.ts
Fil: > it’s the DuckDB version as described [in the DuckDB docs](https://duckdb.org/docs/extensions/overview.html#installation-location)

I think it might be slightly different for duckdb-wasm. For stable versions it's certainly going to follow the same pattern, but for nightly builds I'm not sure how the URL is built.

For now, when I upgrade to a nightly by changing `1.29.0` to `1.29.1-dev3.0` (which is built against DuckDB `1.1.2`, I can see that the client is trying to load
`http://127.0.0.1:3000/_duckdb/extensions.duckdb.org/v1.1.2/wasm_eh/parquet.duckdb_extension.wasm` (which is not available in framework since `1.1.1` is hard-coded).

I've now centralized the version numbers to make this easier to update in the future.

src/client/stdlib/duckdb.js
Fil: done

src/build.ts
Fil: Yes—currently I don't rely on any constancy from the extensions repository. Even if duckdb's hosted extensions are indeed changed only when the duckdb version changes, it won't necessarily be the case for custom extensions. Which is why each extension gets its own INSTALL FROM command with a specific hashed path.

docs/lib/duckdb.md
Fil: This would be great. Maybe we shouldn't expose this "load" option until we manage to implement it as you're suggesting?

docs/config.md
```diff
@@ -297,6 +297,23 @@ export default {
...
+
+```js run=false
+duckdb: {
+  install: ["json", "spatial", "h3", "custom"],
+  load: ["spatial"],
+  source: {
+    h3: "community",
+    custom: "https://my-custom-repo.tld"
+  }
+}
```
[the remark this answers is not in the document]
mbostock: Have you had a chance to consider this feedback yet?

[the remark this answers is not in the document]
Fil: I've now added tests and updated the documentation.

In addition, I have decoupled install and load. You can now *load* an extension from a remote source, for example if you want to create a "spatial"+"h3" sql template literal without self-hosting h3. (This might be more useful for users who develop their own extensions and want to serve them from a custom extensions repo, rather than from a local self-hosted copy.)

src/duckdb.ts
Fil: I've tried to make the duckdb-wasm and DuckDB versions part of the configuration (to allow users to experiment with upgrades), but this seems to require deep changes, that I'm not ready to do. I've left this out for now (and note that it might be solved by an environment variable instead).

docs/sql.md
```diff
@@ -206,3 +206,47 @@ Inputs.table(await sql([`SELECT * FROM gaia WHERE source_id IN (${[source_ids]})
...
+```sql echo
+SELECT '127.0.0.1'::INET AS ipv4, '2001:db8:3c4d::/48'::INET AS ipv6;
+```
+
+During development, you can experiment freely with extensions that are not self-hosted. For example to try out the “h3” `community` extension:
+
+```sql echo run=false
+INSTALL h3 FROM community;
+LOAD h3;
+SELECT format('{:x}', h3_latlng_to_cell(37.77, -122.43, 9)) AS cell_id;
```
mbostock, line 239: As I believe I commented previously, I feel this is an antipattern and not something we should recommend even during development. It has the side-effect of mutating the set of extensions loaded into DuckDB, changing the behavior of other SQL code blocks. And it does not self-host the extension, as would happen if you went through the recommend route. We should only recommend one way of registering extensions that works in both development and production. (And if we want to have some pages have different sets of extensions, we should use a duckdb front matter or equivalent.) I’m restructuring the docs now and will remove this section.

src/duckdb.ts
```diff
@@ -0,0 +1,90 @@
...
+export async function resolveDuckDBExtension(
+  root: string,
+  platform: string,
+  repo: string,
+  name: string
+): Promise<string> {
+  if (!repo.startsWith("https://")) throw new Error(`invalid repo: ${repo}`);
+  const cache = join(root, ".observablehq", "cache");
+  const file = `${name}.duckdb_extension.wasm`;
+  const ref = `${repo}/v${DUCKDBVERSION}/wasm_${platform}/${file}`.slice("https://".length);
```
mbostock, line 72: Use `new URL(path, base)` to compute URLs rather than string concatentation.

src/build.ts
```diff
@@ -204,7 +222,7 @@ export async function build(
   // Anything in _observablehq also needs a content hash, but anything in _npm
   // or _node does not (because they are already necessarily immutable).
   for (const path of globalImports) {
-    if (path.endsWith(".js")) continue;
+    if (path.endsWith(".js") || path.startsWith("/_duckdb/")) continue;
```
mbostock, line 225: The comment above this line is now a bit confusing:

```js
  // Copy over global assets (e.g., minisearch.json, DuckDB’s WebAssembly).
  // Anything in _observablehq also needs a content hash, but anything in _npm
  // or _node does not (because they are already necessarily immutable).
```

Here we’re expressly skipping `/_duckdb/`, but the confusing part is that this only covers DuckDB’s extensions; DuckDB’s Wasm bundles still live in `/_npm/@duckdb/duckdb-wasm/`… Maybe it’s okay though.

src/libraries.ts
```diff
@@ -72,14 +74,20 @@ export function getImplicitStylesheets(imports: Iterable<string>): Set<string> {
...
   const implicits = new Set<string>();
   if (set.has("npm:@observablehq/duckdb")) {
     implicits.add("npm:@duckdb/duckdb-wasm/dist/duckdb-mvp.wasm");
     implicits.add("npm:@duckdb/duckdb-wasm/dist/duckdb-browser-mvp.worker.js");
     implicits.add("npm:@duckdb/duckdb-wasm/dist/duckdb-eh.wasm");
     implicits.add("npm:@duckdb/duckdb-wasm/dist/duckdb-browser-eh.worker.js");
+    if (!duckdb) throw new Error("Implementation error: missing duckdb configuration");
+    for (const [name, {source}] of Object.entries(duckdb.extensions)) {
+      for (const platform of duckdb.bundles) {
+        implicits.add(`duckdb:${platform},${name},${source}`);
```
mbostock, line 88: Any reason not to follow the same convention that DuckDB does here for custom repositories?

src/config.ts
```diff
@@ -499,3 +521,41 @@ export function mergeStyle(
 export function stringOrNull(spec: unknown): string | null {
   return spec == null || spec === false ? null : String(spec);
 }
+
+// TODO convert array of names
```
mbostock, line 525: TODO Remove this to-do.

src/libraries.ts
mbostock: I guess one answer is that `name` and `platform` should never contain a comma, making it easier to split — but there’s no guarantee that `source` doesn’t contain a comma, so it’s not safe to `path.split(",")` and expect to get all the parts back out again. I’m going to tweak this a bit to match the DuckDB convention and make it more robust.

mbostock (approved):

fabito:
How do we use/enable the extensions support ? Do we need to wait for an official release or can we install the prerelease vesion ?

Fil:
It's possible but difficult; my recommendation is to wait (a few days max) for the next release of Framework.

PR #1736, opened 2024-10-10
add made by observable component

CobusT:
Adds the Made by Observable button to the header of the site.
- created a `logo()` function in the config file since the code was duplicated
- desktop vs mobile classes to show different versions of the GitHub information 
- removed the toc z-index since it conflicts with the component's popover. It doesn't seem to be needed for the toc.


mbostock (approved):

src/style/layout.css
```diff
@@ -296,7 +296,6 @@ body {
   position: fixed;
   color: var(--theme-foreground-muted);
   font: 400 14px var(--sans-serif);
-  z-index: 1;
```
mbostock, line 299: The TOC links are currently not clickable without the z-index: 1 because they are obscured (invisible) by the main column. We’ll need to fix this ASAP, perhaps by putting the observable-made-by within its own z-index container.

CobusT: Thanks @mbostock

PR #1738, opened 2024-10-10
bug in sql.md: fix the query to ignore nulls

Fil:
I see this bug in the last chart of https://observablehq.com/framework/sql

spurious-rect

this spurious rect corresponds to datum 34 which has mag1 = mag2 = null. Plot does not handle correctly the `null`s in arrow vectors.

<del>It's still unclear to me why it does not happen all the time, and why it happens at all.</del>

### DIFF
--- a/docs/sql.md
+++ b/docs/sql.md
@@ -174,11 +174,12 @@ Plot.plot({
   marks: [
     Plot.axisY({tickFormat: (d) => d / 1000, label: "count (thousands)"}),
     Plot.rectY(await sql`
-      SELECT
-        FLOOR(phot_g_mean_mag / 0.2) * 0.2 AS mag1
-      , mag1 + 0.2 AS mag2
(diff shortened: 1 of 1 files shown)


Fil:
And this is how it looks like under the duckdb-1.29 branch
broken-1 29

Fil:
And sometimes when I'm just fiddling with the code block, it looks like this:
bug-0-100

even if I'm not changing the query!

Fil:
This seems to be an issue with the native handling of arrow tables in `@observablehq/plot@0.6.16` in fact, because the bug disappears when I use `@observablehq/plot@0.6.15`.

Fil:
maybe it's better to fix it also independently of Plot’s https://github.com/observablehq/plot/issues/2194

mbostock (approved):

PR #1745, opened 2024-10-11
Deploy option to enable continuous deployment

tophtucker:
Notes
- Whether you build locally or in cloud, the deploy method ends with pollForProcessingCompletion. In the local case, that just means uploading files; in the cloud case, it means doing a whole cloud build. Should we have the same timeout in both cases?
- Currently this prompts you to enable continuous deployment even if your project directory is ineligible. 
- Once your continuous deployment setting is saved, do we ever prompt you about it again? Is there a command to turn it on, or do we instruct you to edit deploy.json?

### DIFF
--- a/src/deploy.ts
+++ b/src/deploy.ts
@@ -204,15 +204,13 @@ class Deployer {
   }
 
   private async cloudBuild(deployTarget: DeployTargetInfo) {
-    if (deployTarget.create) return false;
-
-    // kick off new cloud deploy w/ link to deploy details
+    if (deployTarget.create) return false; // TODO
     const {deployPollInterval: pollInterval = DEPLOY_POLL_INTERVAL_MS} = this.deployOptions;
     await this.apiClient.postProjectBuild(deployTarget.project.id);
     const spinner = this.effects.clack.spinner();
     spinner.start("Requesting deploy");
     const pollExpiration = Date.now() + DEPLOY_POLL_MAX_MS;
-    pollLoop: while (true) {
+    while (true) {
       if (Date.now() > pollExpiration) {
         spinner.stop("Requesting deploy timed out");
         throw new CliError("Requesting deploy failed");
@@ -227,12 +225,10 @@ class Deployer {
             deployTarget.workspace.login
           }/${deployTarget.project.slug}/deploys/${latestCreatedDeployId}`
         );
-        // break pollLoop;
         return latestCreatedDeployId;
       }
       await new Promise((resolve) => setTimeout(resolve, pollInterval));
     }
-    return true;
   }
 
   private async maybeLinkGitHub(deployTarget: DeployTargetInfo): Promise<boolean> {
@@ -241,7 +237,7 @@ class Deployer {
       // can do cloud build
       return true;
     } else {
-      // TODO Where should it look for .git?
+      // TODO Where should it look for .git? only supports projects at root rn…
       // const isGit = existsSync(this.deployOptions.config.root + "/.git");
       const isGit = existsSync(".git");
       if (isGit) {
@@ -252,68 +248,54 @@ class Deployer {
         const gitHub = remotes.find(([, url]) => url.startsWith("https://github.com/"));
         if (gitHub) {
           const repoName = formatGitUrl(gitHub[1]);
-          const confirmLinkGitHub = await this.effects.clack.confirm({
-            message: `Do you want to link to GitHub repository ${repoName}?`,
-            active: "Yes",
-            inactive: "No"
-          });
-          if (this.effects.clack.isCancel(confirmLinkGitHub) || !confirmLinkGitHub) {
-            throw new CliError(
-              "Continuous deployment is enabled in deploy.json but you cannot deploy in the cloud without a GitHub repository",
-              {print: true, exitCode: 0}
-            );
-          }
-          if (confirmLinkGitHub) {
-            const {repositories} = await this.apiClient.getGitHubRepositories();
-            const authedRepo = repositories.find(({url}) => formatGitUrl(url) === repoName);
-            if (authedRepo) {
-              // authed repo found
-              await this.apiClient.postProjectEnvironment(deployTarget.project.id, {
-                source: {
-                  provider: authedRepo.provider,
-                  provider_id: authedRepo.provider_id,
-                  url: authedRepo.url,
-                  branch: null // TODO detect branch
-                }
-              });
-              return true;
-            } else {
-              // repo not auth’ed; kick off web auth flow
-              this.effects.clack.log.info(
-                "Authorize Observable to access this repository: https://github.com/apps/observable-data-apps-dev/installations/select_target"
-              );
-              const spinner = this.effects.clack.spinner();
-              spinner.start("Waiting for repository to be authorized");
-              const pollExpiration = Date.now() + 2 * 60_000; //DEPLOY_POLL_MAX_MS; // TODO
-              pollLoop: while (true) {
-                if (Date.now() > pollExpiration) {
-                  spinner.stop("Waiting for repository to be authorized timed out");
-                  throw new CliError("Deploy failed");
-                }
-                const {repositories} = await this.apiClient.getGitHubRepositories();
-                const authedRepo = repositories.find(({url}) => formatGitUrl(url) === repoName);
-                if (authedRepo) {
-                  spinner.stop("Repository authorized");
-                  await this.apiClient.postProjectEnvironment(deployTarget.project.id, {
-                    source: {
-                      provider: authedRepo.provider,
-                      provider_id: authedRepo.provider_id,
-                      url: authedRepo.url,
-                      branch: null // TODO detect branch
-                    }
-                  });
-                  // break pollLoop; // TODO
-                  return true;
-                }
-                await new Promise((resolve) => setTimeout(resolve, 2000));
+          const {repositories} = await this.apiClient.getGitHubRepositories();
+          const authedRepo = repositories.find(({url}) => formatGitUrl(url) === repoName);
+          if (authedRepo) {
+            // authed repo found
+            await this.apiClient.postProjectEnvironment(deployTarget.project.id, {
+              source: {
+                provider: authedRepo.provider,
+                provider_id: authedRepo.provider_id,
+                url: authedRepo.url,
+                branch: null // TODO detect branch
               }
+            });
+            return true;
+          } else {
+            // repo not auth’ed; link to auth page and poll for auth
+            this.effects.clack.log.info(
+              `Authorize Observable to access the ${bold(repoName)} repository: ${link("https://github.com/apps/observable-data-apps-dev/installations/select_target")}`
+            );
+            const spinner = this.effects.clack.spinner();
+            spinner.start("Waiting for repository to be authorized");
+            const pollExpiration = Date.now() + DEPLOY_POLL_MAX_MS;
+            while (true) {
+              if (Date.now() > pollExpiration) {
+                spinner.stop("Waiting for repository to be authorized timed out");
+                throw new CliError("Deploy failed");
+              }
+              const {repositories} = await this.apiClient.getGitHubRepositories();
+              const authedRepo = repositories.find(({url}) => formatGitUrl(url) === repoName);
+              if (authedRepo) {
+                spinner.stop("Repository authorized");
+                await this.apiClient.postProjectEnvironment(deployTarget.project.id, {
+                  source: {
+                    provider: authedRepo.provider,
+                    provider_id: authedRepo.provider_id,
+                    url: authedRepo.url,
+                    branch: null // TODO detect branch
+                  }
+                });
+                return true;
+              }
+              await new Promise((resolve) => setTimeout(resolve, 2000));
             }
           }
         } else {
-          // no github remote
+          throw new CliError("No GitHub remote"); // TODO better error
         }
       } else {
-        // not a repo
+        throw new CliError("Not at root of a git repository"); // TODO better error
       }
     }
     return false;
@@ -325,6 +307,7 @@ class Deployer {
     const deployConfig2 = await this.getUpdatedDeployConfig(); // TODO inelegant… move cd prompt to getUpdatedDeployConfig?
     let deployId;
     if (deployConfig2.continuousDeployment) {
+      // TODO move maybeLinkGitHub so that continuous deployment is only enabled if it succeeds
       await this.maybeLinkGitHub(deployTarget);
       deployId = await this.cloudBuild(deployTarget);
     } else {
--- a/src/deploy.ts
+++ b/src/deploy.ts
@@ -1,7 +1,10 @@
+import {exec} from "node:child_process";
 import {createHash} from "node:crypto";
 import type {Stats} from "node:fs";
+import {existsSync} from "node:fs";
 import {readFile, stat} from "node:fs/promises";
 import {join} from "node:path/posix";
+import {promisify} from "node:util";
 import slugify from "@sindresorhus/slugify";
 import wrapAnsi from "wrap-ansi";
 import type {BuildEffects, BuildManifest, BuildOptions} from "./build.js";
@@ -20,6 +23,7 @@ import type {
   DeployManifestFile,
   GetCurrentUserResponse,
   GetDeployResponse,
+  GetProjectEnvironmentResponse,
   GetProjectResponse,
   WorkspaceResponse
 } from "./observableApiClient.js";
@@ -32,6 +36,10 @@ import {bold, defaultEffects as defaultTtyEffects, faint, inverse, link, underli
 const DEPLOY_POLL_MAX_MS = 1000 * 60 * 5;
 const DEPLOY_POLL_INTERVAL_MS = 1000 * 5;
 const BUILD_AGE_WARNING_MS = 1000 * 60 * 5;
+
+export function formatGitUrl(url: string) {
+  return new URL(url).pathname.slice(1).replace(/\.git$/, "");
+}
 
 export interface DeployOptions {
   config: Config;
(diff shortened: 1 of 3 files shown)


src/deploy.ts
```diff
@@ -204,7 +204,9 @@ class Deployer {
   }
 
   private async cloudBuild(deployTarget: DeployTargetInfo) {
-    if (deployTarget.create) return false; // TODO
+    if (deployTarget.create) {
+      throw Error("Incorrect deployTarget state");
```
mbostock, line 208:
```suggestion
      throw new Error("Incorrect deployTarget state");
```

src/deploy.ts
```diff
@@ -225,20 +227,25 @@ class Deployer {
...
+        return latestCreatedDeployId!;
       }
       await new Promise((resolve) => setTimeout(resolve, pollInterval));
     }
   }
 
   private async maybeLinkGitHub(deployTarget: DeployTargetInfo): Promise<boolean> {
-    if (!this.effects.isTty || deployTarget.create) return false;
+    if (deployTarget.create) {
+      throw Error("Incorrect deployTarget state");
```
mbostock, line 241:
```suggestion
      throw new Error("Incorrect deployTarget state");
```

src/deploy.ts
```diff
@@ -190,14 +203,128 @@ class Deployer {
...
+    while (true) {
+      if (Date.now() > pollExpiration) {
+        spinner.stop("Requesting deploy timed out.");
+        throw new CliError("Requesting deploy failed");
+      }
+      const {latestCreatedDeployId} = await this.apiClient.getProject({
+        workspaceLogin: deployTarget.workspace.login,
+        projectSlug: deployTarget.project.slug
+      });
+      if (latestCreatedDeployId !== deployTarget.project.latestCreatedDeployId) {
```
tophtucker, line 224: chatting with fil: there's no guarantee that the new deploy ID is the one you just kicked off; maybe your colleague click the deploy button around the same time. currently, the postProjectBuild method can't return the new deploy ID because it just dispatches a message to the job-manager, which at some point will get around to making a new deploy. two options:

- create deploy id earlier, in api, and pass to job manager (which feels like it might mess with our messaging protocol? i don't know all the reasons it was designed like it is today)
- pass some unique string to the api, which will pass it to the job manager, which will eventually respond with it, which would let us identify that the deploy was our own

but fil thinks this is probably not a blocking issue with this PR

src/deploy.ts
```diff
@@ -190,14 +203,128 @@ class Deployer {
     return deployInfo;
   }
 
-  private async startNewDeploy(): Promise<GetDeployResponse> {
-    const deployConfig = await this.getUpdatedDeployConfig();
-    const deployTarget = await this.getDeployTarget(deployConfig);
-    const buildFilePaths = await this.getBuildFilePaths();
```
tophtucker, line 196: fil notes that this could be parallelized: getBuildFilePaths doesn't depend on the previous two; it's filesystem i/o, as opposed to talking to the api. might speed up local deploys a bit

src/deploy.ts
```diff
@@ -190,14 +203,128 @@ class Deployer {
...
+    let deployId: string | null;
+    if (deployConfig.continuousDeployment) {
+      deployId = await this.cloudBuild(deployTarget);
+    } else {
+      const buildFilePaths = await this.getBuildFilePaths();
+      deployId = await this.createNewDeploy(deployTarget);
+      await this.uploadFiles(deployId, buildFilePaths);
+      await this.markDeployUploaded(deployId);
+    }
     return await this.pollForProcessingCompletion(deployId);
```
tophtucker, line 328: this pollForProcessingCompletion pattern, in the continuous deployment case, keeps the user's terminal busy even though at that point all the work is being done on our servers. i mean, they can always ctrl-c, and we could tell them it's safe to do that. we could return early and say "when your deploy is done, if it succeeds, it will be visible at this url". but maybe it's better to wait so that we can return an error code if it fails?

tophtucker: fil thinks it's correct as-is; if someone has scripted this, they can run `yarn deploy` to kick off a cloud build, and they will receive a CliError code if the cloud deploy fails.

src/deploy.ts
```diff
@@ -274,15 +401,18 @@ class Deployer {
...
+  ): Promise<{deployTarget: DeployTargetInfo; deployConfig: DeployConfig}> {
     let deployTarget: DeployTargetInfo;
     if (deployConfig.workspaceLogin && deployConfig.projectSlug) {
       try {
         const project = await this.apiClient.getProject({
           workspaceLogin: deployConfig.workspaceLogin,
           projectSlug: deployConfig.projectSlug
         });
-        deployTarget = {create: false, workspace: project.owner, project};
+        const environment = await this.apiClient.getProjectEnvironment({id: project.id});
```
tophtucker, line 414: i want to build this into getProject. it could at least be part of that api response; ideally we'd also fold the project_config table into new columns on the projects table. 

(currently, getProjectEnvironment has slightly different permissions: it's only accessible by members of the project's workspace who can view the project. for a public project, we should not return those fields. but that would still be doable if it were one table and one endpoint; we'd just serve a redacted version in the public response.)

src/deploy.ts
```diff
@@ -190,14 +203,128 @@ class Deployer {
...
+      }
+      await new Promise((resolve) => setTimeout(resolve, pollInterval));
+    }
+  }
 
-    await this.uploadFiles(deployId, buildFilePaths);
-    await this.markDeployUploaded(deployId);
+  private async maybeLinkGitHub(deployTarget: DeployTargetInfo): Promise<boolean> {
+    if (deployTarget.create) {
+      throw new Error("Incorrect deployTarget state");
```
tophtucker, line 241: fil says: update to something more user-readable

```suggestion
      throw new Error("Inconsistent deploy target state");
```

they still wouldn't know what to do with that but at least they could send it to us…

src/deploy.ts
```diff
@@ -190,14 +203,128 @@ class Deployer {
...
+    }
+    if (!this.effects.isTty) return false;
```
tophtucker, line 243: this is probably wrong…

src/deploy.ts
```diff
@@ -190,14 +203,128 @@ class Deployer {
...
+    if (deployTarget.environment.build_environment_id && deployTarget.environment.source) {
+      // can do cloud build
+      return true;
```
tophtucker, line 246: fil notes: if we broke our local git configuration, shouldn't we check it? do we want to check if everything is correct? there's a good chance everything will work, but…

if you have erased your .git folder, or you have changed your remotes, or if you're not on the right branch…

should we check that local and remote git refs match?? the user is expecting a cloud deploy of the same application state they are looking at locally!! if they have local unstaged changes, or are on a different branch, they could get surprising results from doing a cloud build.

if observable cloud is configured to pull from the main branch, and you're on toph/onramp locally, what are you expecting to happen when you run `yarn deploy`?

and it'll get more confusing with branch previews. how does it work on vercel? on vercel, there's no way to trigger a deploy except to push, so there's no question here. maybe that should be our norm? except you still wanna be able to re-run data loaders, which depend on changing external state not captured by commits.

tophtucker: maybe at a minimum we should just check that the local git repo still exists, still has that remote, and is on the same branch.

src/deploy.ts
```diff
@@ -190,14 +203,128 @@ class Deployer {
...
+    } else {
+      // We only support cloud builds from the root directory so this ignores this.deployOptions.config.root
+      const isGit = existsSync(".git");
```
tophtucker, line 249: todo: log cwd when running yarn deploy. you can run yarn deploy from a child directory like src, but i think it still runs in the context of the root directory, in which case this is correct.

src/deploy.ts
mythmon: Given a magic wand, would you want to stream the build logs here?

src/deploy.ts
```diff
@@ -190,14 +203,128 @@ class Deployer {
...
+      if (isGit) {
+        const remotes = (await promisify(exec)("git remote -v", {cwd: this.deployOptions.config.root})).stdout
```
tophtucker, line 251: note that in loader.ts we make our promises "by hand" (and with spawn instead of exec), but in create.ts we already use promisify, so… seems fine?

src/deploy.ts
```diff
@@ -190,14 +203,128 @@ class Deployer {
...
+          .split("\n")
+          .filter((d) => d)
+          .map((d) => d.split(/\s/g));
+        const gitHub = remotes.find(([, url]) => url.startsWith("https://github.com/"));
```
tophtucker, line 255: fil instead has, i think, the SSH case, which we should also check for:

```
❯ git remote -v
observable	git@github.com:observablehq/pangea.git (fetch)
observable	git@github.com:observablehq/pangea.git (push)
origin	git@github.com:Fil/pangea.git (fetch)
origin	git@github.com:Fil/pangea.git (push)
```

src/deploy.ts
```diff
@@ -190,14 +203,128 @@ class Deployer {
...
+        if (gitHub) {
+          const repoName = formatGitUrl(gitHub[1]);
```
tophtucker, line 257: formatGitUrl will also be different in the SSH case

src/deploy.ts
```diff
@@ -190,14 +203,128 @@ class Deployer {
...
+          const repositories = (await this.apiClient.getGitHubRepositories())?.repositories;
+          const authedRepo = repositories?.find(({url}) => formatGitUrl(url) === repoName);
```
tophtucker, line 259: feels strange to ask for the list of all repositories instead of asking if you're authorized for a given repository!

src/deploy.ts
```diff
@@ -190,14 +203,128 @@ class Deployer {
...
+          if (authedRepo) {
+            // Set branch to current branch
+            const branch = (
+              await promisify(exec)("git rev-parse --abbrev-ref HEAD", {cwd: this.deployOptions.config.root})
```
tophtucker, line 263: i think cwd may be wrong here, it should be running in the normal cwd of the command, at the root, since we don't support non-root repos/projects

src/deploy.ts
```diff
@@ -190,14 +203,128 @@ class Deployer {
...
+                    branch: null // TODO detect branch
+                  }
+                });
+                return true;
+              }
+              await new Promise((resolve) => setTimeout(resolve, 2000));
+            }
+          }
+        } else {
+          this.effects.clack.log.error("No GitHub remote found; cannot enable continuous deployment.");
```
tophtucker, line 308: these should exit with a CliError rather than just logging a message and continuing on with a local deploy

src/deploy.ts
tophtucker: should use this instead https://docs.github.com/en/rest/repos/repos?apiVersion=2022-11-28#get-a-repository

src/deploy.ts
```diff
@@ -384,18 +524,40 @@ class Deployer {
       }
     }
 
+    let {continuousDeployment} = deployConfig;
+    if (continuousDeployment === null) {
+      const enable = await this.effects.clack.confirm({
+        message: wrapAnsi(
+          `Do you want to enable continuous deployment? ${faint(
+            "This builds in the cloud and redeploys whenever you push to this repository."
```
tophtucker, line 532: Dunstan says this could note upfront that continuous deployment requires a GitHub repository

src/deploy.ts
tophtucker: @mythmon Eventually, yeah!

src/deploy.ts
tophtucker: done in https://github.com/observablehq/observablehq/pull/18390

src/deploy.ts
tophtucker: done in https://github.com/observablehq/framework/pull/1745/commits/9e234a2d9135173b2cb907fd24612200bbb30c37 (in this PR) and https://github.com/observablehq/observablehq/pull/18314/commits/3c1b098fce580cf567f4f351b61fdea00b37f9c9 (in https://github.com/observablehq/observablehq/pull/18314)

src/deploy.ts
tophtucker: should also verify that the repo hasnt been unlinked on the web side

src/deploy.ts
tophtucker: done

src/deploy.ts
tophtucker: done

src/deploy.ts
tophtucker: sticking with this pattern for now

src/deploy.ts
tophtucker: done. (now throws instead of returning false, because whole method has changed from a philosophy of "return true or false" to "throw an error or return nothing".)

src/deploy.ts
tophtucker: done — now validates repo every time and checks that local and remote repos match. i don't check refs yet

src/deploy.ts
tophtucker: not gonna worry about this for now; also, buildFilePaths is only relevant for the local deploy path, so it's not as clean to parallelize anymore

src/deploy.ts
tophtucker: done

src/deploy.ts
tophtucker: done

src/deploy.ts
tophtucker: done

src/deploy.ts
tophtucker: i don't think it matters for this command but it was unnecessary and conceptually wrong so i removed the cwd option

src/deploy.ts
tophtucker: done

src/deploy.ts
tophtucker: tried logging cwd out of curiosity and indeed, as one would hope/expect, yarn deploy runs in the root directory no matter where you call it from.

src/deploy.ts
```diff
@@ -384,18 +524,40 @@ class Deployer {
...
+        ),
+        active: "Yes, enable and build in cloud",
+        inactive: "No, build locally"
+      });
+      if (this.effects.clack.isCancel(enable)) throw new CliError("User canceled deploy", {print: false, exitCode: 0});
+      continuousDeployment = enable;
+    }
+
+    // Disables continuous deployment if there’s no env/source & we can’t link GitHub
+    if (continuousDeployment) continuousDeployment = await this.maybeLinkGitHub(deployTarget);
```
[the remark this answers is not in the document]
tophtucker: done, and it feels like a much better way to think about it

src/deploy.ts
tophtucker: not planning to do anything about this for now

tophtucker:
I think this is ready for review again!

## Purpose

The idea is that, when a user runs `yarn deploy`, they’re given the option to enable continuous deployment (persisted in deploy.json). If they do, we link the repo and immediately kick off a cloud build; the command exits when the build completes (w/ a link to the deployed app) or fails, same as with a local build. Thereafter, users shouldn’t really have to run `yarn deploy` anymore, since they can just push to GitHub — but if they do run `yarn deploy` again (e.g. to pick up new data), it kicks off another cloud build.

Note that the deploy command ensures you’re currently in the same repo and branch that’s configured in our database. (We could go further and add some info to the API response so we can check that the refs match.) The idea is that you’re trying to deploy the state of the repo you’re in, not just controlling the data app management console from a command line in a repo in some different state.

Note that currently there’s no CLI command to change the continuousDeployment setting once set; Fil suggested we could have a separate command to review all settings in deploy.json, but I’m inclined to save that for another PR. 

## Implementation

When run manually from the command line (as opposed to in our cloud build environment where we pass in an existing deploy ID), running deploy calls startNewDeploy. It calls getDeployTarget to find out where it’s deploying to; in this PR, I’ve updated getDeployTarget to check the continuousDeployment option; if it’s not present, we prompt the user whether to enable it. Back in startNewDeploy, if continuousDeployment is on, we call validateGitHubLink and then cloudBuild; else we build locally and deploy as usual. (validateGitHubLink checks that you have an authorized and matching repo linked in our database; if not, and you’re in a repo with a GitHub remote, we link it, or prompt you to go to the web (to a new screen added in https://github.com/observablehq/observablehq/pull/18314) to authorize it if necessary.)

This adds dependencies on a couple API routes:
- /cli/github/repository?owner=${props.ownerName}&repo=${props.repoName}, to check the authorization of a single repository
- /cli/project/${id}/environment, to set a GitHub link; this endpoint could potentially be rolled into the post project edit endpoint
- /cli/project/${id}/build, to initiate a cloud build & deploy of a project

docs/deploying.md
```diff
@@ -36,9 +36,11 @@ npm run deploy -- --help
 
 </div>
 
-## Automated deploys
+## Continuous deployment
 
-After deploying an app manually at least once, Observable can handle subsequent deploys for you automatically. You can automate deploys both [on commit](https://observablehq.com/documentation/data-apps/github) (whenever you push a new commit to your project’s default branch) and [on schedule](https://observablehq.com/documentation/data-apps/schedules) (such as daily or weekly).
+<!-- TODO: decide whether to primarily use “continuous deployment” or “automatic deploys” -->
```
Fil, line 41: I believe CD is the accepted industry term

Fil (dismissed):
review and suggestions #1805

Fil:
I wonder if we could/should provide a post-commit or post-push hook that would remind the user that their push is going to trigger a build (and where to see the logs etc). (I was just editing a project and could not remember if I had already enabled CD on it.)

Fil (commented):
Can I undo my "Request changes" status without giving approval?

mbostock:
@Fil You can dismiss your review.

PR #1746, opened 2024-10-11
some opinionated layout changes to match other docs

CobusT:


mbostock (commented):
Appreciate the contribution but I’m going to wait on this a bit to decide how I feel. I don’t think I want to tie us pixel-perfect to the VitePress layout, or that I like the VitePress layout better than ours, and I’m not sure about the effort to make all the sites match exactly.

CobusT:
Understood. Should I change this to a draft for now?

PR #1748, opened 2024-10-12
@observablehq/runtime 6.0.0

### LINKED ISSUE #573: @observablehq/runtime should declare dist/runtime.js in exports.
Currently we make an assumption about this file relative to the src/index.js, but if that package declared it as an export, it would be more stable.

### LINKED ISSUE #937: Remove the stdlib and inspector dependency from the runtime
We’ve implemented our own self-hosted standard library within Framework, but we’re still using the same Runtime that notebooks use which depends on the notebook standard library. Related to #573, we should restructure the Runtime so that it doesn’t depend on `@observablehq/stdlib` and `@observablehq/inspector`, giving users (and us) more flexibility and saving a few bytes.

### LINKED ISSUE #1469: Serialize asynchronous evaluation to improve performance

mbostock:
Fixes #1469. Fixes #937. Fixes #573. Depends on https://github.com/observablehq/runtime/pull/374.

~~To test locally, you can `yarn link` in @observablehq/runtime, and then `yarn link @observablehq/runtime` in Framework.~~ I published 6.0.0-rc.1 so you can simply run `yarn` to test.


mbostock:
Going to merge this now so we can test the rc in production.

PR #1751, opened 2024-10-13
Expose SQLite.export().

angrytongan:


Fil:
This is interesting. Can we clarify what the use case is, though? I can see how re-importing the same data is useful to show that it works, but besides the demo would you want to reuse that export on the same page? Or is it more—say—that you would use the page to edit a db, then click on a  button to download an export or save it to an asset server?

mbostock:
Maybe the better solution here would be to document how to access the underlying SQLite instance. I don’t think we want to create wrapper functions for everything that SQLite supports, especially if it’s not a use case we expect is common.

df-rw:
> Can we clarify what the use case is, though?

🤔 I want to persist the SQLite database at the client by puling it out with `export()` and storing it somewhere, then pushing it back in with `load()`.

> would you want to reuse that export on the same page?

You're right, probably not the best way to show the example by using the same data on the same page.

> Maybe the better solution here would be to document how to access the underlying SQLite instance. I don’t think we want to create wrapper functions for everything that SQLite supports, especially if it’s not a use case we expect is common.

My thinking was "if there is a `load()`, there should be a matching `save()`" 🤷 Happy to go with accessing SQLite directly for things like this 👍

PR #1758, opened 2024-10-14
handle absolute global imports

### LINKED ISSUE #1755: absolute URL imports aren’t correctly resolved during build
I'm importing from npm, ESM code that imports other ESM by URL reference.

Framework is rewriting those URLs so that they are invalid and I get 404s in my build.

image

In dev mode URLs aren't rewritten and my code loads correctly.

Not sure if Framework wants to traverse remote URLs or probably just avoid rewriting them.

mbostock:
Fixes #1755. ~~I’d like a test here, but I think it would require us to mock out the contents of jsDelivr responses. I tested manually using the repro in the linked issue.~~ Added an automated test, easy enough.


Fil (approved):
👍

PR #1761, opened 2024-10-16
handle repeated 429s from Observable API with exponential back off

mythmon:
Framework already respects HTTP 429 rate limit responses from the Observable API. However, it will only retry once. This can sometimes not be enough, especially in cases where there are more than one Framework instances on the same IP, or Framework is not the only thing making request to the Observable API.

This PR changes the retry logic to retry up to 3 times, each time multiplying the Retry-After time the server instructed us to use by successive powers of two. After 3 times, Framework gives up and just lets the request fail.

In the worst case scenario, this would mean Framework waits in total 7 times longer than the server's "Retry-After" header. I've seen this Retry-After header as high as 60 seconds on our API, so this could mean a 7 minute delay.

I'm happy to bike shed on the number of retries or the way in which the time grows. We're also implementing some server side changes to help the situation.

### DIFF
--- a/src/observableApiClient.ts
+++ b/src/observableApiClient.ts
@@ -6,6 +6,7 @@ import type {ApiKey} from "./observableApiConfig.js";
 import {faint, red} from "./tty.js";
 
 const MIN_RATE_LIMIT_RETRY_AFTER = 1;
+const MAX_RATE_LIMIT_ATTEMPTS = 3;
 
 export function getObservableUiOrigin(env = process.env): URL {
   const urlText = env["OBSERVABLE_ORIGIN"] ?? "https://observablehq.com";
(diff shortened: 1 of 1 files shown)


jessedearing (approved):

src/observableApiClient.ts
```diff
@@ -69,11 +70,18 @@ export class ObservableApiClient {
       throw error;
     }
 
-    if (response.status === 429) {
+    let rateLimitCount = 0;
+    while (response.status === 429) {
+      if (rateLimitCount >= MAX_RATE_LIMIT_ATTEMPTS) {
```
mbostock, line 75: It looks like `rateLimitCount` is only incremented once when `this._rateLimit === null`; am I reading it wrong? Normally I like to combine the increment and check into the same statement so that it’s more obvious that it can never infinite loop.

```suggestion
      if (rateLimitCount++ >= MAX_RATE_LIMIT_ATTEMPTS) {
```

mbostock (changes requested):

src/observableApiClient.ts
mythmon: You're right that the rateLimitCount is only incremented when this._rateLimit is null. However, the bug is something different. `this._rateLimit` is intended to be used for multiple concurrent callers of this function to wait collaborative. That's not working with this new approach either though. I'll fix it up.

mythmon:
I'm not going to get to this for a while, I'm going to close it for now.

PR #1762, opened 2024-10-17
tailwind

Fil:


Fil:
Another thing to figure out is the hash. I don't think we want to update the css each time any md or js file changes… so maybe we have to watch all these files for changes in the list of statically analyzable "potential tailwind class names" that they contain.

CobusT:
I was just playing with this PR, and it looks like, once the `.observablehq/cache/tailwind.config.js` file is created without having a config file in place, the cached file has `const config = {};` and doesn't get regenerated with a reference to the config file when you then add a config file. 

- When I delete the cached file, it generates a new cache file with the import for the config.
- When I then delete the `docs/tailwind.config.js` file, the cache file gets updated correctly, and then has `const config = {};` again.
- When I then add a new `docs/tailwind.config.js` file, then cache file does not refresh and doesn't import the config file.

mbostock:
In preview we could generate a tailwind.css that is specific to just the current page (+ any imported local modules) rather than continuously trying to recompute it for the entire app.

Fil:
I've updated the logic a bit, now we have continuous updating of a separate tailwind.css

(It's still not limited to the current page, because I don't know how to pass an object configuration—it seems it only wants a file-based configuration).

I've had to blocklist the grid classes, since they conflict with ours. I guess it's going to be a fork in the road: continue with our own grid, or switch to tailwind's. The difference is that in tailwind you have to be explicit about your breakpoints.

Speaking of breakpoints we should make sure we use the same. Adding this to the TODO.

Fil:
I've now tested tailwind plugins, it just works. For example, 

```sh
yarn add --dev @tailwindcss/typography
```

then in `docs/tailwind.config.js`:

```js
import typography from "@tailwindcss/typography";

export default {
  plugins: [ typography ]
};
```

src/rollup.ts
```diff
@@ -59,6 +59,10 @@ export async function bundleStyles({
   });
   let text = result.outputFiles[0].text;
   if (path === getClientPath("stdlib/inputs.css")) text = rewriteInputsNamespace(text);
+  // dirty patch for tailwind: remove margin:0 and styles resets for headers
+  // etc. It should probably be a tailwind plugin instead.
+  if (path === getClientPath("tailwind.css"))
+    text = text.replaceAll(/}[^{]*h1,\n*h2,\n*h3,\n*h4,\n*h5,\n*h6[^}]+}\s*/g, "}");
```
mbostock: You want the [typography plugin](https://github.com/tailwindlabs/tailwindcss-typography) and `prose` class.

Fil: Maybe, but there are other resets that wreck our page, and we might want to find a way to remove the resets independently of adding /prose/ (this [works](https://github.com/observablehq/framework/pull/1762#issuecomment-2437441570) already, anyway). I was thinking maybe a custom plugin we'd inline.

CobusT:
Do we need to change the default breakpoints for tailwind? Users may find that surprising.

src/rollup.ts
Fil: Yes that's a good possibility. For now I'm merely taking stock of the things that change.

The other big change I've noticed are the `grid-cols-*` classes, which I have to block for now because they are quite different. Ours is responsive to width, tailwind's responsiveness is based on a breakpoint modifier (_e.g._ `lg:grid-cols-3`).

`blocklist: ["grid", "grid-cols-2", "grid-cols-3", "grid-cols-4"],`

Fil:
These are the breakpoints of framework, my assumption here is that users are working inside “main” so the reference for sizes is the width of main.

Here's a way to test:

```md
## breakpoints

<ul>
  <li><span class="text-red-500">&lt;sm</span> (always on)</li>
  <li><span class="sm:text-red-500">sm</span></li>
  <li><span class="md:text-red-500">md</span></li>
  <li><span class="lg:text-red-500">lg</span></li>
  <li><span class="xl:text-red-500">xl</span></li>
</ul>
```

if we leave tailwind's breakpoints unchanged, these element light up at very "random" places.

(It's the same issue that the [tailwindcss-container-queries](https://github.com/tailwindlabs/tailwindcss-container-queries) plugin addresses.)

CobusT:
> These are the breakpoints of framework, my assumption here is that users are working inside “main” so the reference for sizes is the width of main.

That probably makes sense, then. If they want to reset it they can do that in their own config file, but this change is probably more good than unexpected :-)

Fil:
The reference is here, with an example showing the contorsions you have to do when you have a sidebar.
https://v1.tailwindcss.com/docs/breakpoints

In our case it's even more complicated, because the space available to draw stuff (which I believe is what should drive the definitions of sm, md, lg), depends on whether the TOC is shown. And that's not only a function of css and the page width…

(I don't like the way the TOC works, and the fact that it is not visible at all on smaller screens is also problematic, so it might be the final draw telling that we have to rewrite it with tailwind.)

PR #1766, opened 2024-10-18
previews exported files, adds preview tests

### LINKED ISSUE #1694: preview static assets
(excerpted from #1681)

Fil:
closes #1694


mbostock (commented):
This needs to enforce that the required file is listed in **dynamicPaths**, no?

Fil:
It's not something that we're doing for parameterized pages and resources, _i.e._ you can call `page-foo` if there is a `page-[var].md`, even if that parameter is not listed in dynamicPaths.

Maybe we should (at least) log a line of warning? Another approach might be to return a 404 with an explicit message that says “this resource exists and we could serve it if you declared it in dynamicPaths”… or even maybe a button to bypass that problem, when you're developing something and don't want to get thrown out of your flow.

PR #1767, opened 2024-10-18
a parameterized route can map [foo] onto itself

### LINKED ISSUE #1760: bracketed parameter name as a parameter value
Suppose you have `src/[param].md`

* opening http://127.0.0.1:3000/[foo] shows the page parameterized with param=[foo]. ⁽ᶠᶦˣᵉᵈ· ʷᵃˢ ᵖᵃʳᵃᵐ⁼ᶠᵒᵒ⁾
* opening http://127.0.0.1:3000/[param] shows the non-parameterized page

If dynamicPaths contains `"/[foo]", "/[param]"`, the first page is generated as a parameterized page, the second as a normal page. 

Definitely a corner case, but this feels inconsistent. We should probably prioritize routing to the parameterized page?

Fil:
closes #1760

### DIFF
--- a/src/route.ts
+++ b/src/route.ts
@@ -66,6 +66,7 @@ function routeParams(root: string, cwd: string, parts: string[], exts: string[])
         const path = join(root, cwd, first + ext);
         if (existsSync(path)) {
           if (!statSync(path).isFile()) return; // ignore non-files
+          if (first.match(/\[.*\]/)) continue; // skip [param]
           return {path: join(cwd, first + ext), ext};
         }
       }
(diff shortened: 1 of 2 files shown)


src/route.ts
```diff
@@ -66,6 +66,7 @@ function routeParams(root: string, cwd: string, parts: string[], exts: string[])
         const path = join(root, cwd, first + ext);
         if (existsSync(path)) {
           if (!statSync(path).isFile()) return; // ignore non-files
+          if (first.match(/\[.*\]/)) continue; // skip [param]
```
mbostock, line 69: This should use the stricter regex we use elsewhere for params. Let me see if I can fix…

mbostock: Also, can we pull this test out of the inner loop?

mbostock (approved):

PR #1769, opened 2024-10-19
adopt picocolors

mbostock:
Replaces our hand-rolled version with the popular [picocolors](https://github.com/alexeyraspopov/picocolors), offering support for `NO_COLOR` and relying on picocolor’s own color support detection. As an alternative, we could also use [yoctocolors](https://github.com/sindresorhus/yoctocolors) which already uses [`hasColors`](https://nodejs.org/api/tty.html#writestreamhascolorscount-env)… but I figured picocolors is nice because it’s already a dependency of clack.

Related #1764.

### DIFF
--- a/package.json
+++ b/package.json
@@ -83,6 +83,7 @@
     "mime": "^4.0.0",
     "minisearch": "^6.3.0",
     "open": "^10.1.0",
+    "picocolors": "^1.1.1",
     "pkg-dir": "^8.0.0",
     "resolve.exports": "^2.0.2",
     "rollup": "^4.6.0",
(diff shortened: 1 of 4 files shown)


Fil:
It works in local tests (with say, `yarn docs:build | tee build.log`), but doesn't work at all on cloud: my logs are full of color codes.
https://observablehq.com/projects/@observablehq/pangea/deploys/e9b695e6f5001670

logs

When I look at the source code, the logic is defaulting to using colors when `!!env.CI`. In other words, it is expected that CI always supports ANSI colors.
https://github.com/alexeyraspopov/picocolors/blob/7249f8c5d4825550f70bc1ea98652639933d3bbd/picocolors.js#L4

I think it's a fair assumption, and that we should merge this in nonetheless, when we're ready.

mbostock:
That’s a pre-existing problem with Cloud logs, right? E.g., this [oss-analytics deploy](https://observablehq.com/projects/@observablehq/oss-analytics/deploys/f09a23a547d024cc). This makes it worse, but seems like something that shouldn’t block this PR.

Screenshot 2024-10-25 at 7 42 08 PM

Fil (approved):

PR #1775, opened 2024-10-22
Save source information to the build manifest

Fil:
- [x] Each time we run a loader, save the loader information to the cache.
  - [ ] Using a big JSON file for now, but we might want to use a sqlite db instead to solve the race conditions.
- [ ] Add loader information to manifest.pages
- [ ] Add manifest.modules
- [ ] Add manifest.files (?)

### DIFF
--- a/src/route.ts
+++ b/src/route.ts
@@ -62,11 +62,13 @@ function routeParams(root: string, cwd: string, parts: string[], exts: string[])
       return;
     case 1: {
       const [first] = parts;
-      for (const ext of exts) {
-        const path = join(root, cwd, first + ext);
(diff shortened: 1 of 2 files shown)


mbostock (commented):
Don’t think we need (or should) write this out to a cache. We know this information at build time, in memory, and should write this out to the manifest then. If we write it out to a file and read it back in again it feels more complex and also more likely to get out of sync. Is there another reason to write this out to the cache that I’m missing?

Fil:
Thanks for pointing this out. I think I took a wrong turn when I thought the information was lost once the data loaders have run. 😓

PR #1776, opened 2024-10-22
note loaders

Fil:
- [x] get source & params for pages
- [x] get source & params for modules
- [x] get source & params for files (copied and generated)
- [x] write the build manifest
- [x] fix types in deploy
- [x] test deploy
- [x] add unit test for all 4 cases
- [x] add `config:{root}` to the build manifest
- [x] fix deploys (currently error 400)
- [x] add unit test for deploys


src/loader.ts
```diff
@@ -73,25 +73,15 @@ export class LoaderResolver {
...
    * Loads the page at the specified path, returning a promise to the parsed
    * page object.
    */
   async loadPage(path: string, options: LoadOptions & ParseOptions, effects?: LoadEffects): Promise<MarkdownPage> {
     const loader = this.findPage(path);
     if (!loader) throw enoent(path);
-    const source = await readFile(join(this.root, await loader.load(options, effects)), "utf8");
-    return parseMarkdown(source, {params: loader.params, ...options});
+    const input = await readFile(join(this.root, await loader.load(options, effects)), "utf8");
+    return parseMarkdown(input, {source: loader.path, params: loader.params, ...options});
```
Fil, line 84: TODO: this change of variables names is wrong, like it is below in parseMarkdown. The issue is that there are too many things that we want to call "path".

src/build.ts
```diff
@@ -74,6 +75,23 @@ export async function build(
...
+    type: string,
+    file: string,
+    {title, path, params}: {title?: string | null; path?: string; params?: Params}
+  ) => {
+    const source = path == null || path === file.slice(1) ? null : join("/", path);
+    buildManifest[type].push({
+      path: config.normalizePath(file),
+      ...(title != null && {title}),
+      ...(source && {source}),
+      ...(params && {params})
```
Fil, line 91: we don't strictly need to write the params since they can be derived from the file + source by pattern matching

should I remove them?

mythmon: It'd be nice to include them here, since you have them. It saves us work later, and I'm not worried about the extra file size.

mythmon:
I'm not sure I'm following what this is doing exactly. I used this to build the Framework docs, and the generated `_build.json` had lots of entries that didn't seem to exist in `dist/`. For example, `/observable.png` was listed in the `files` key, but `dist/observable.png` doesn't exist. It's actually at `dist/_file/observable.1af93621.png`.

I expected this manifest would only list items when they were a part of the `dynamicPaths` option (or a similar one if we rename this). I also expected that the `path` field would be the path to the file in the dist directory, and thus the path on the esrver. Did I misunderstand? Or is this just not quite done yet because it is in draft?

Fil:
Thanks for the review. I have now:
* fixed the file name for normal (non-exported) files and modules, and
* commented them out of the manifest, leaving only the files and modules found in `dynamicPaths`

I've left the code because, even if it's out of scope for now, I think there is value potentially in knowing which (non-exported) files and modules depend on which source files. So maybe it would be right to uncomment these two calls, even if we don't yet have a reporter script to make use of them. (Or at least we know that we can reinstate them when we want.)

For instance, for an internal (non exported) parameterized module, it's nice to know that `_import/module-exported.281d2524.js` is coming from `/module-[type].js`:

```json
{
  "params": {"type": "exported"},
  "path": "/_import/module-exported.281d2524.js",
  "source": "/module-[type].js"
}
```

src/build.ts
Fil: My concern here is that we expose a bit more API surface, so we now have to wonder _e.g._ what happens if someone uses a modified version of framework that creates a discrepancy. (And in any case there is work to do to validate the inputs.)

mythmon:
I agree that information would be interesting. I think that I'd want to be able to distinguish the unexported items from the exported ones though. Maybe with a boolean flag?

src/build.ts
mythmon: That's a good point. Either way, we don't need this now so we could go with the more minimal version at first.

Fil:
Added the source information for non-exported modules and files — with new keys (`_file` and `_import`) rather than a boolean.

src/loader.ts
Fil: Well there is no good solution; for example the path is "/" and the source is "/index.md.js", and we need the source for tracking it, and the path for the fragment functions.

Fil:
tested live… 🎉

mythmon (approved):
Apologies, I didn't realize this was waiting on me.

src/build.ts
```diff
@@ -75,6 +75,23 @@ export async function build(
...
+
+  const buildManifest: BuildManifest = {
+    ...(title && {title}),
+    config: {root},
+    pages: [],
+    modules: [],
+    files: []
+  };
+  const addToManifest = (type: string, file: string, {title, path}: {title?: string | null; path?: string}) => {
+    const source = path == null || path === file.slice(1) ? null : join("/", path);
```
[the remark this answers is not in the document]
Fil: `path === file.slice(1)` is for files that point to themselves (e.g. `path=chart.js` `file=/chart.js`) 

the `path` is `null` for pages that have no `source` 
this is only used in the tests for `parseMarkdown`… I can find a better way to express this.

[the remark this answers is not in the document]
Fil: A bit better now; I'm still a bit unhappy with the naming (too many things we call "path")

mbostock (approved):

PR #1780, opened 2024-10-25
dark/light body class

Fil:
This adds a `dark` or `light` class on the body to reflect the `color-scheme` property set by css.

For the rationale: we can't access the `color-scheme` property through css; only the browser can do that. I don't know why it's like that, and find it irritating. This "fixes" it somehow, and allows to create styles such as:

`.dark .warning { border: 1px solid red }`

which are otherwise impossible to do.

This is also necessary to support dark mode properly with tailwind (#1762).

### DIFF
--- a/src/client/stdlib/generators/dark.ts
+++ b/src/client/stdlib/generators/dark.ts
@@ -1,20 +1,18 @@
 import {observe} from "./observe.js";
 
-// Watches dark mode based on the color-scheme set by the current theme.
+// Watches dark mode based on theme and user preference.
+// TODO: in preview, also watch for changes in the theme meta.
 export function dark() {
   return observe((notify: (dark: boolean) => void) => {
     let dark: boolean | undefined;
-    const probe = document.createElement("div");
-    probe.style.transitionProperty = "color";
-    probe.style.transitionDuration = "0.001s";
+    const media = matchMedia("(prefers-color-scheme: dark)");
     const changed = () => {
-      const d = getComputedStyle(probe).getPropertyValue("color-scheme") === "dark"; // TODO "light dark"?
+      const d = getComputedStyle(document.body).getPropertyValue("color-scheme") === "dark";
       if (dark === d) return; // only notify if changed
       notify((dark = d));
     };
-    document.body.append(probe);
-    probe.addEventListener("transitionstart", changed);
     changed();
-    return () => probe.remove();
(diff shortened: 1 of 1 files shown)


mbostock (changes requested):
This is inconsistent with how we’ve (currently) implemented Generators.dark; the two should definitely be consistent — and ideally using the same code.

It also feels a little overkill to me to use the probe transition technique in production when we only expect to need it (versus a media listener) because the stylesheets can change during development. Perhaps we could use a conditional NODE_ENV test inside of Generators.dark. I know I’ve also advised to minimize the differences between production and development, but if it’s entirely hidden within the Generators.dark implementation, that seems pretty safe. Or we could just always do the probe transition technique…

Fil:
I've come to believe that the media listener is only one of the top-level mechanisms that allow the page to decide what mode it is in. Other mechanisms are ignoring the preferred media (ok), enforcing the color mode via a toggle (not a best practice, but can be useful for developers), changing the mode based on an arbitrary computation (good for artists and fun).

In all these cases the styles should set a correct `color-scheme` property, so that the browser knows what to do when it displays native inputs… basically because otherwise, the site is broken.

This I think implies that the current node’s (computed)  is the correct value that must drive the decision when a component has both light and dark colors. (In other words, our components should work like the native browser inputs.)

So, I need to modify Generators.dark, and see how to use it so that it also sets the class.

src/render.ts
```diff
@@ -93,9 +93,9 @@ ${preview ? `\nopen({hash: ${JSON.stringify(resolvers.hash)}, eval: (body) => ev
 ${html.unsafe(rewriteHtml(page.body, resolvers))}</main>${renderFooter(page.footer, resolvers, options)}
 </div>
 </body>
-<script>{${html.unsafe(
-    (await rollupClient(getClientPath("dark-init.js"), options.root, path, {resolveImport, minify: true})).trim()
-  )}}</script>
+<script type="module">import{Generators as g}from'${resolveImport(
```
mbostock: I think the old approach of having this in a separate script was cleaner, and meant we could use minification.

Also, why not just add 

```js
import "./dark.js";
```

to `src/client/index.js` rather than having it as another script element?

Fil: yes this is much better, thanks! (now fixing tests…)

Fil:
to wrap into #1762 if that's the best course of action

PR #1784, opened 2024-10-26
preserveIndex, preserveExtension

### LINKED ISSUE #1781: `cleanUrls: "preserve-index"` to preserve `/index.html`
Currently even with `cleanUrls: false` we still convert `/index.html` to `/`, but not all hosts support this. We should have a setting that preserves `/index.html`.

mbostock:
Fixes #1781. Replaces the **cleanUrls** option with separate **preserveIndex** and **preserveExtension** options.


Fil (approved):
added unit tests

PR #1794, opened 2024-11-02
show 1 extension

Fil:
cheat a little to make the output of FROM duckdb_extensions() more interesting

| before | after |
|--|--|
| Capture d’écran 2024-11-02 à 13 07 28 | Capture d’écran 2024-11-02 à 13 07 19 |

### DIFF
--- a/docs/lib/duckdb.md
+++ b/docs/lib/duckdb.md
@@ -140,9 +140,8 @@ SELECT ST_Area('POLYGON((0 0, 0 1, 1 1, 1 0, 0 0))'::GEOMETRY) as area;
 
 To tell which extensions have been loaded, you can run the following query:
 
-```js
-await new Promise((r) => setTimeout(r, 2000));
-display(Inputs.table(await sql`FROM duckdb_extensions() WHERE loaded;`));
+```sql
+FROM duckdb_extensions() WHERE loaded AND JSON '1';
 ```
 
 ```sql run=false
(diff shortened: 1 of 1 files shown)


docs/lib/duckdb.md
```diff
@@ -140,7 +140,12 @@ SELECT ST_Area('POLYGON((0 0, 0 1, 1 1, 1 0, 0 0))'::GEOMETRY) as area;
 
 To tell which extensions have been loaded, you can run the following query:
 
-```sql echo
+```js
+await new Promise((r) => setTimeout(r, 2000));
```
mbostock, line 144: Instead of waiting an arbitrary timeout, how about waiting a query that is known to autoload the JSON extension?

Fil: fixed with `WHERE JSON '1'`

docs/lib/duckdb.md
```diff
@@ -140,9 +140,8 @@ SELECT ST_Area('POLYGON((0 0, 0 1, 1 1, 1 0, 0 0))'::GEOMETRY) as area;
 
 To tell which extensions have been loaded, you can run the following query:
 
-```js
-await new Promise((r) => setTimeout(r, 2000));
-display(Inputs.table(await sql`FROM duckdb_extensions() WHERE loaded;`));
+```sql
+FROM duckdb_extensions() WHERE loaded AND JSON '1';
```
mbostock, line 144: Better, but now we’re confusing how you query for which extensions are loaded. 😅 I suggested awaiting an earlier query.

Fil: this query is not shown to the user

mbostock: Ah, right you are! Thank you. I missed that from the diff.

mbostock (approved):

PR #1804, opened 2024-11-07
fix within on Windows

mbostock:
Ref. https://github.com/observablehq/framework/discussions/1792

### DIFF
--- a/src/files.ts
+++ b/src/files.ts
@@ -7,13 +7,8 @@ import {cwd} from "node:process";
 import {fileURLToPath} from "node:url";
 import {isEnoent} from "./error.js";
 
-export function toOsPath(path: string): string {
-  return path.split(sep).join(op.sep);
-}
-
-export function fromOsPath(path: string): string {
-  return path.split(op.sep).join(sep);
-}
+export const toOsPath = sep === op.sep ? (path: string) => path : (path: string) => path.split(sep).join(op.sep);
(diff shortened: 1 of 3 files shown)


src/files.ts
```diff
@@ -7,13 +7,8 @@ import {cwd} from "node:process";
...
 
-export function toOsPath(path: string): string {
-  return path.split(sep).join(op.sep);
-}
-
-export function fromOsPath(path: string): string {
-  return path.split(op.sep).join(sep);
-}
+export const toOsPath = sep === op.sep ? (path: string) => path : (path: string) => path.split(sep).join(op.sep);
+export const fromOsPath = sep === op.sep ? (path: string) => path : (path: string) => path.split(op.sep).join(sep);
```
mbostock, line 11: This doesn’t change functionality; just trivial optimization.

src/path.ts
```diff
@@ -86,6 +87,7 @@ export function parseRelativeUrl(url: string): {pathname: string; search: string
 }
 
 export function within(root: string, path: string): boolean {
+  const {relative, normalize, resolve, isAbsolute} = op;
   path = relative(normalize(resolve(root)), normalize(resolve(path)));
```
mbostock, line 91: The main issue was that `process.cwd()` returns a “namespaced” absolute path starting with _e.g._ `D:\`, whereas `resolve(path)` returns a non-namespaced absolute path. See these logs for reference:

https://github.com/observablehq/framework/actions/runs/11717337869/job/32636834499

Fil (approved):

PR #1808, opened 2024-11-11
use event.target.value

Fil:
Currently when you wrap an interactive element in a `div`, maybe for aesthetic reasons, or because you want to add a title etc, you can't use `Generators.input` to read the value — or you need to copy the value somehow to the top-level element, as we do in Plot with the `figure` element.

(This, even though the `input` event bubbles up, which makes `Generators.input` slightly inconsistent.)

This change makes Generators.input read the value on the `input` event’s **target**, making it easier to read the value of, say, a chart within a `resize` call.

~~~~md
```js echo
function scatterPlot(width) {
  return Plot.plot({
    width,
  marks: [Plot.dot(cars, { x: "power (hp)", y: "economy (mpg)", tip: true })]
})};
```

<div id=chart class="grid grid-cols-1">
  <div class="card">${resize((width) => scatterPlot(width))}</div>
</div>

```js echo
const v = Generators.input(document.getElementById("chart"));
```

<pre>${JSON.stringify(v, null, 2)}</pre>
~~~~ 

Note: if there are several inputs inside the "chart", so that whichever input (chart) you're touching returns its value. To combine values, you still need to use Inputs.form (but maybe Inputs.form could benefit from the same treatment?)

(OP: https://github.com/observablehq/framework/discussions/1806#discussioncomment-11211770)

Leaving as a draft as there may be unwanted consequences that I'm not seeing.

### DIFF
--- a/src/client/stdlib/generators/input.js
+++ b/src/client/stdlib/generators/input.js
@@ -4,7 +4,7 @@ export function input(element) {
   return observe((change) => {
     const event = eventof(element);
     let value = valueof(element);
-    const inputted = () => change(valueof(element));
+    const inputted = (event) => change(valueof(event.target));
     element.addEventListener(event, inputted);
     if (value !== undefined) change(value);
     return () => element.removeEventListener(event, inputted);


Fil:
Three tests fail, but unrelated to what the change does; it's due to changes in the hash instead. I've filed https://github.com/observablehq/framework/pull/1809 to fix that, independently.

src/client/stdlib/generators/input.js
```diff
@@ -4,7 +4,7 @@ export function input(element) {
   return observe((change) => {
     const event = eventof(element);
     let value = valueof(element);
-    const inputted = () => change(valueof(element));
+    const inputted = (event) => change(valueof(event.target));
```
mbostock, line 7: The fact that the initial value is `valueof(element)` while subsequent values are `valueof(event.target)` suggests that this may not be desirable. I see the benefit of being able to listen to a parent element, but in that case, how to determine the initial value of the generator? The generator would need to search within the container for an element that exposes a value. (And pick the first of multiple, arbitrarily?)

We could have the `resize` helper proxy the `value` of the returned element. Then you could say:

````md
```js
const chart = resize((width) => scatterPlot(width));
const value = Generators.input(chart);
```

<div class="grid grid-cols-1">
  <div class="card">${chart}</div>
</div>
````

mbostock (changes requested):

src/client/stdlib/generators/input.js
Fil: Good remark, I hadn't thought about the initial value. I think I'd prefer the “search” approach because it would be easier to use — and also because it fits my mental model of an event that bubbles up (wants to be seen and useful).

PR #1809, opened 2024-11-11
in build tests, sort by extension then path

Fil:
this fixes the wobbling tests (duckdb, sql, npm) that change when a commit modifies the content hash of inputs.js vs inputs.css


test/build-test.ts
```diff
@@ -75,7 +76,7 @@ describe("build", () => {
       // renumber the hashes so they are sequential. This way we don’t have to
       // update the test snapshots whenever Framework’s client code changes. We
       // make an exception for minisearch.json because to test the content.
-      for (const path of findFiles(join(outputDir, "_observablehq"))) {
+      for (const path of sort(findFiles(join(outputDir, "_observablehq")), extname, (d) => d)) {
```
mbostock, line 79: I think the third argument `(d) => d` is being ignored here. Presumably we don’t need to pass any arguments to `sort` and should sort by the path?

Fil: `findFiles` already sorts by path, by construction. However the annoying bit is the ordering of `inputs.xxxxx.css` vs `inputs.yyyyy.js`, which depends on both hashes. Sorting by `extname` ensures all the `css` are numbered before the `js`. But if we only sort by `extname`, then the css and the js are arbitrarily sorted within their group.

mbostock (commented):
This’ll definitely help, but I don’t think it’s quite enough. There’s also a `findFiles(outputDir)` that calls `normalizeHash` on L94 that would need to be sorted. 

Also note that even with sorting, it doesn’t guarantee that the hashing is completely deterministic _across_ projects because the set of hashed files can be different. For example, the deploy test expects `theme-air,near-midnight.dcdbf18e.css` to get the normalized hash of `00000004`. That works in this case:

```
/_observablehq/client.addacdeb.js 00000001
/_observablehq/runtime.9393ab6d 00000002
/_observablehq/stdlib.0891ee25.js 00000003
/_observablehq/theme-air,near-midnight.dcdbf18e.css 00000004
```

But it would be a different hash if the project (simple-public) used search:

```
/_observablehq/client.addacdeb.js 00000001
/_observablehq/minisearch.55535b23.js 00000002
/_observablehq/runtime.9393ab6d.js 00000003
/_observablehq/search.2515eb68.js 00000004
/_observablehq/stdlib.0891ee25.js 00000005
/_observablehq/theme-air,near-midnight.dcdbf18e.css 00000006
```

I think it’s okay that the hashes aren‘t consistent across projects, but for some reason when I fixed added the `sort` on L94, it’s looking for `theme-air,near-midnight.00000001.css`… maybe that’s a holdover from you sorting by `extname`?

Fil:
I think the call on L94 is safe because it's only reading the Map of aliases that were computed earlier, not setting any alias.

test/build-test.ts
mbostock: Oh… I completely forgot that d3.sort accepts multiple accessors/comparators!

mbostock (approved):
I tweaked the sort order to reduce churn. It doesn’t perfectly match what was there before but it’s less.

PR #1815, opened 2024-11-13
Automatically annotate file paths if dynamicPaths include JS modules or is a function

mythmon:
This is a bit messy, especially the leaky abstraction of the config file, but it would fix the problem.

Some alternate approaches I thought about while writing this:

* Unconditionally enable file annotations
  * This is simpler, but creates a larger diff and is unneeded on many data apps.
* Enable annotation on dynamicPath modules and all of their transitive dependencies.
  * This would be a lot harder for me, and I'm not sure if it would be robust enough.
* Somehow force that file annotations are enabled, possibly by rebuilding JS files, when deploying.
  * Haven't quite figured this one out yet
* Reject un-annotated builds on the server side if they seem required.


Fil:
_Unconditionally enable file annotations_ sounds good to me.

mbostock:
> Reject un-annotated builds on the server side if they seem required.

I feel like this is the way to go. Even with this PR, old releases of Framework will still not work, so we’d need to enforce that a minimum version of Framework is used. So why not more directly enforce that Framework was built with path annotation enabled? (I don’t recall if we stash some metadata when building, but we certainly could if we aren’t already.)

PR #1837, opened 2024-11-22
add lastModified to a module’s hash when it references a file attachment

### LINKED ISSUE #1836: immutable module hash should reflect a referenced file attachment’s lastModified
When a module references a data loader, and that loader creates a new file with the same contents as before, the hash of the module doesn't change.

But its contents do change, because the module references the file _and its lastModified_.

Fil:
when a module references a file attachment, include the file's `lastModified` timestamp in the module's hash.

closes #1836

Note that this allows us to remove from the tests the ts rewrite that we used to do on modules, because their payload’s lastModified is now always set to the fake currentDate injected by the tests.

Some tests don't want to use the fake date though! (I hope they never run in parallel?)


test/loader-test.ts
```diff
@@ -118,6 +119,8 @@ describe("LoaderResolver.find(path)", () => {
 
 describe("LoaderResolver.getSourceFileHash(path)", () => {
   const time = new Date(Date.UTC(2023, 11, 1));
+  before(() => setCurrentDate(null));
+  after(() => setCurrentDate(new Date("2024-01-10T16:00:00")));
```
mbostock, line 123: This seems highly undesirable since this test now has the side-effect of setting the current date _for all tests that run subsequently_. I’m investigating why this is needed.

PR #1870, opened 2024-12-07
fix react-dom@19

### LINKED ISSUE #1866: Framework v1.13 building a broken version of React 19
I'm opening an issue because I think I've found a bug while debugging [an issue with my own work](https://github.com/observablehq/framework/discussions/1865) on the discussion forum.

Here are my steps to reproduce:

I installed a fresh version of Framework 1.13 using `npx @observablehq/framework@latest create`.

I created a super simple JSX component in `components/Button.jsx`:

```jsx
import * as React from "npm:react";

export default function Button() {
  return (
    <button>Test</button>
  )
}
```

I import and add this button to the existing `index.md` file:

```js
import Button from './components/Button.js';

display(<Button />)
```

The above `display()` call is in a fenced `jsx` block, sorry I don't know how to display that properly in here.  

Anyways, then I run `npm run dev` and load the homepage, and the button does not appear.  

I see this error in the browser console:

```
Uncaught (in promise) TypeError: createRoot is not a function at client.js:211:19
```

And sure enough in the client.js bundle it's importing the `_npm/react-dom@19.0.0` library but that file does not export a method called `createRoot`..  

Now, if I look in the `.observablehq/cache/_npm` folder I see that the `react-dom@19.0.0` folder contains one file:  `_esm.js` and that file is only 4,492 bytes.   

The same file in a previous Framework v11.1 project, for `react-dom@18.3.1` is around 192k ...  Plus there are other files in that folder including a `package.json` etc..  Those other files are not present here in v1.13

Similarly the `_esm.js` file in the `cache/_npm/react@19.0.0` folder seems suspiciously small as well (11k)..   

This is all from a fresh vanilla install with no other changes except creating a basic JSX component to test.

I'm on a Mac (Sequoia 15.1.1) and I'm running Node 20.11.1

Fil:
The root cause is https://github.com/facebook/react/pull/28271 that was merged into version 19, which means we now have to pull in ReactDOM from "npm:react-dom/client" instead of "npm:react-dom".

This is (barely) documented in the release notes https://github.com/facebook/react/releases/tag/v19.0.0

From some manual testing, this PR is compatible with react-dom@18, but not @\17.

- I'm trying to see if we can reroute "npm:react-dom" to "npm:react-dom/client" for asc. compat. (NO)
- how to make this work with node (local install) (I DON'T KNOW)

closes #1866

### DIFF
--- a/src/node.ts
+++ b/src/node.ts
@@ -83,7 +83,13 @@ export function extractNodeSpecifier(path: string): string {
  */
 function isBadCommonJs(specifier: string): boolean {
   const {name} = parseNpmSpecifier(specifier);
-  return name === "react" || name === "react-dom" || name === "react-is" || name === "scheduler";
+  return (
+    name === "react" ||
+    name === "react-dom" ||
+    name === "react-dom/client" ||
+    name === "react-is" ||
+    name === "scheduler"
+  );
 }
 
 function shimCommonJs(specifier: string, require: NodeRequire): string {
--- a/docs/jsx.md
+++ b/docs/jsx.md
@@ -72,7 +72,7 @@ import {useState} from "npm:react";
 React DOM is also available as `ReactDOM` in Markdown, or can be imported as:
 
 ```js run=false
-import * as ReactDOM from "npm:react-dom";
+import * as ReactDOM from "npm:react-dom/client";
 ```
 
 You can define components in JSX modules. For example, if this were `components/Card.jsx`:
--- a/src/client/main.js
+++ b/src/client/main.js
@@ -94,7 +94,7 @@ function reject(root, error) {
 }
 
 function displayJsx(root, value) {
-  return (root._root ??= import("npm:react-dom").then(({createRoot}) => {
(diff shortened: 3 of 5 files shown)


Fil:
I'm not sure what to do about 
https://github.com/observablehq/framework/blob/8efdcf45c00f7cb471deab01f65fe93e1b539d44/package.json#L132-L133

ref. https://github.com/observablehq/framework/pull/1229

src/node.ts
```diff
@@ -83,7 +83,13 @@ export function extractNodeSpecifier(path: string): string {
  */
 function isBadCommonJs(specifier: string): boolean {
   const {name} = parseNpmSpecifier(specifier);
-  return name === "react" || name === "react-dom" || name === "react-is" || name === "scheduler";
+  return (
+    name === "react" ||
+    name === "react-dom" ||
+    name === "react-dom/client" ||
```
mbostock, line 89: The `/client` would never be part of the `name`; it’s the `path`.

Fil: true—I shouldn't have commited that part, it is for node and … I haven't found how to make it work (at all)

PR #2052, opened 2026-03-03
Local npm resolution

yucelkaratas-acn:
# Add `localNpmResolve` config option for offline / air-gapped environments

**Vibecoded** — this branch was created with AI assistance.

## Summary

Adds a new `localNpmResolve` boolean config property that switches all npm package resolution away from the jsDelivr CDN and onto the project's local `node_modules`. Useful for air-gapped environments, corporate networks that block CDN access, or workflows that need reproducible builds from pinned local installs.

When enabled, Observable Framework logs a one-time notice at startup:
```
[observable] npm resolution: local (node_modules)
```

## Usage

```js
// observablehq.config.js
export default {
  localNpmResolve: true
};
```

### New module — `src/packageResolution.ts`
A routing layer that sits between the rest of the codebase and the two resolution backends:

| Mode | Backend | Output path prefix |
|---|---|---|
| CDN (default) | `resolveNpmImport` / `resolveNpmImports` / `populateNpmCache` | `/_npm/…` |
| Local | `resolveNodeImportFrom` / `resolveNodeImports` / local cache path | `/_node/…` |

Three public functions replace the direct calls to the npm/node backends throughout the codebase:
- `resolvePackageImport(root, specifier)` — resolves a single import specifier
- `resolvePackageImports(root, path)` — resolves all transitive imports of a cached module
- `ensurePackageCache(root, path)` — returns (and populates) the cache path for a module

`/_node/` paths are always handled locally regardless of the mode flag (they are already local bundles).

### Config (`src/config.ts`)
- Added `localNpmResolve: boolean` to the `Config` and `ConfigSpec` interfaces (defaults to `false`)
- `normalizeConfig` reads the value and calls `setLocalNpmResolve()` to push it into the routing layer

### Specifier translation (`toLocalSpecifier` in `packageResolution.ts`)
CDN specifiers often carry `/+esm` suffixes (e.g. `d3-array/+esm`, `mime/lite/+esm`). These are stripped before handing off to Node's resolver:
- `d3-array/+esm` → `d3-array`
- `mime/lite/+esm` → `mime/lite`
- `d3-array/src/index.js/+esm` → `d3-array/src/index.js`

### Package root discovery (`findPackageRoot`)
Walks up from the project root using `pkg-dir` (already a dependency), with a manual fallback that looks for a `node_modules` directory, so it works even without a `package.json`.

### Other files
- `src/build.ts`, `src/preview.ts`, `src/resolvers.ts`, `src/rollup.ts`, `src/javascript/module.ts`, `src/node.ts`, `src/client/stdlib/fileAttachment.js` — switched from direct `resolveNpmImport`/`resolveNodeImport` calls to the new routing functions
- `src/node.ts` — exports new `resolveNodeImportFrom(root, packageRoot, spec)` so the routing layer can resolve against an arbitrary package root rather than `root` itself
- `package.json` — bumps a minor dep version

### Tests — `test/packageResolution-test.ts`
16 new tests covering:
- `isLocalNpmMode` / `setLocalNpmResolve` state toggling
- `resolvePackageImport` in CDN mode (uses `mockJsDelivr`, verifies `/_npm/` paths)
- `resolvePackageImport` in local mode (verifies `/_node/` paths, `/+esm` stripping)
- `resolvePackageImports` routing for `/_node/` paths in both modes
- `ensurePackageCache` local-path return values in all relevant cases

Tests follow existing repo conventions: Mocha + Node `assert`, `mockJsDelivr()` for CDN tests, unique `testRoot` per `describe` block, `before`/`after` hooks to reset shared module state.

## Testing

```sh
yarn test:mocha
```

The local-mode tests (`resolvePackageImport in local mode`, `resolvePackageImports`) require `node_modules` to be installed (`yarn install`) since they resolve against the project's actual dependencies (`d3-array@3.2.4`, `mime@4.0.6`).

## Known limitations / follow-ups

- `localNpmResolve: true` skips version negotiation with the npm registry entirely — the exact version bundled is whatever is installed in `node_modules`
- No validation that a required package is actually installed; Node's resolver will throw with a clear `Cannot find module` error
- The parquet DuckDB extension default was changed from `false` to `{}` as a side-effect of unrelated config normalization touched during this work — worth a separate review

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -5,21 +5,20 @@ on:
   release:
     types: [published]
 
+permissions:
+  id-token: write
+  contents: read
+
 jobs:
   publish:
     runs-on: ubuntu-latest
-    permissions:
-      contents: read
-      packages: write
     steps:
-      - uses: actions/checkout@v4
-      - uses: actions/setup-node@v4
+      - uses: actions/checkout@v6
+      - uses: actions/setup-node@v6
         with:
-          node-version: 20
-          registry-url: 'https://registry.npmjs.org'
-          cache: 'yarn'
+          node-version: 24
+          registry-url: "https://registry.npmjs.org"
+          cache: "yarn"
       - run: yarn --frozen-lockfile
       - run: yarn test
-      - run: npm publish
-        env:
-          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+      - run: npm publish --access public
--- a/docs/config.md
+++ b/docs/config.md
@@ -340,6 +340,26 @@ The `json` and `parquet` are configured (and therefore self-hosted) by default. 
 
 For more, see [DuckDB extensions](./lib/duckdb#extensions).
 
+## localNpmResolve
+
+If true, resolves npm imports from the local `node_modules` directory instead of fetching them from the [jsDelivr](https://www.jsdelivr.com/) CDN. Defaults to false.
+
+This is useful for air-gapped environments, corporate networks that block CDN access, or workflows that require reproducible builds from pinned local installs.
+
+```js run=false
+export default {
+  localNpmResolve: true
+};
+```
+
+When enabled, Framework logs a one-time notice at startup:
+
+```
+[observable] npm resolution: local (node_modules)
+```
+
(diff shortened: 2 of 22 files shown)


tmaier:
HI, this has been tried and tested for 2 months now. We would like to upstream this. Feedback is highly appreciated

PR #2054, opened 2026-03-25
Add locale-aware config and localized framework UI defaults

khashashin:
This PR adds project- and page-level locale support to Framework and localizes Framework-owned UI defaults.

Changes:
  - add `locale`, `lang`, and `dir` to Framework config
  - add front matter overrides for `locale`, `lang`, and `dir`
  - set root HTML `lang` / `dir` during render
  - localize Framework-owned default UI strings, including:
    - home
    - contents / table of contents label
    - untitled page fallback
    - search placeholder
    - sidebar toggle title
    - pager labels
    - default footer text and date formatting
  - update docs and tests

This intentionally keeps translation catalogs and fallback policy out of Framework core; those remain user-land concerns.

This is the Framework-side companion to the [Plot locale work](https://github.com/observablehq/plot/pull/2396), keeping translation catalogs and fallback logic out of Framework core.

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -106,17 +106,35 @@ In this case, the path to the stylesheet is resolved relative to the page’s Ma
 
 The app’s title. If specified, this text is appended to page titles with a separating pipe symbol (“|”). For instance, a page titled “Sales” in an app titled “ACME, Inc.” will display “Sales | ACME, Inc.” in the browser’s title bar. See also the [**home** option](#home).
 
+## locale
+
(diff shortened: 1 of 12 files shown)


Fil:
Is it global, or is it possible to have a multilingual project? Where some articles would be in English and other in Arabic?

khashashin:
@Fil yes, page-level overrides are supported, so you can have a multilingual project where different pages use different language/locale settings.

This is handled via front matter support here: https://github.com/observablehq/framework/pull/2054/changes#diff-77b3519370ff48f6caf7c316f0ece966f2f0871c40e9381d451d2c5b1c76b40fR113

For example:

```yaml
---
lang: ar
dir: rtl
locale: ar-EG
---
```

So you can have one page in English and another in Arabic within the same project.

One current limitation is that some config-derived defaults remain project-level unless explicitly overridden, such as the default `home` label and default `TOC` label. If that should also become page-localized, that could be a follow-up.

